PR #362, opened 2014-07-12
Closes #360: non-ascii heading id generation

### LINKED ISSUE #360: Problems in heading id generation
The custom heading rendering in [the renderer](https://github.com/GitbookIO/gitbook/blob/master/lib/parse/renderer.js#L150) adds id attributes to headings. That's what it adds (after it adds the prefix):
```
raw.toLowerCase().replace(/[^\w -]+/g, '').replace(/ /g, '-');
```

The problems:
1. `\w` matches only ASCII characters in JS. What about other languages?
2. Not only hyphens might be in headings, but also dashes (en dashes).

My suggestion is to change the first `replace` from `[^\w -]` to something like `[,;]`.

Any comments?

OrelBeY:


AaronO:
Fixed by GitbookIO/kramed#8

PR #517, opened 2014-12-08
fix weibo sharing link's class

codepiano:


winniehell:
@SamyPesse moved the file to `theme/templates/website/includes/header.html` in #555. I think you need to rebase your branch.

AaronO:
@winniehell, @SamyPesse is preparing the `2.0.0` release, we'll push to `master` in a few days.

It's best to let PRs rest for a few days till that comes out. So we don't break peoples PRs etc ...

Happy you guys are contributing :+1: We just need to wait a few days to get this release done correctly that will already deal with a lot of the pending PRs and issues.

So it's something to look forward to :)

winniehell:
@AaronO Alright, you guys deal with releases quite differently than in my job. :alien:

AaronO:
@winniehell Haha, to be honest. This is the first release we're doing this way.

It will most likely be pushed to `master` before it gets shipped as `2.0.0`. @SamyPesse is just wanting to cover a few things by himself so as to be cohesive before showing that work. It's really just a few days.

(I'm pretty busy with our internal/private tech, so he's taking ownership of the core features/refactoring for the `2.0.0`).

Besides that, we really try to be as open as possible in the release process ;)

I think @SamyPesse just wants some "peace" to make sure that he does things right.

It's not the typical way we do things either, but I think it makes sense in this particular scenario. He's kind of just working on his own fork before getting back into master :)

Issue #474, opened 2014-10-13
Need a way to change options that are passed to ebook-convert


alexandreaquiles:
The default options that are passed to `ebook-convert` are reasonable, but it is important to have some way to customize those parameters.

In PR #473, hooks before and after `ebook-convert` call are proposed.

The `ebook-before` hook can be used to change options beyond the default.

PR #525, opened 2014-12-15
Hooks before and after ebook generation 

alexandreaquiles:
Solves issue #474.

Hooks `ebook:before` and `ebook:after` are called, respectively, before and after the call to Calibre's `ebook-convert`. The object with `ebook-convert`  options is passed to `ebook:before`.  

Than, an user can create a plugin to change `ebook-convert` default options, among other things.

Example of use:
```
module.exports = {
	hooks: {
		"ebook:before": function(options) {
			options["--disable-font-rescaling"] = true;
			options["--chapter-mark"] = "none";
			options["--level2-toc"] = "//h:h2",
			options["--level3-toc"] = "";
			options["-d debug"] = true;

			console.log("ebook:before");

			return options;
		},
		"ebook:after": function() {
			console.log("ebook:after");
		}
	}
};
```

### DIFF
--- a/lib/generate/ebook/index.js
+++ b/lib/generate/ebook/index.js
@@ -60,27 +60,37 @@ Generator.prototype.finish = function() {
                 "--pdf-footer-template": String(pdfOptions.footerTemplate)
             });
         }
+        
+        return Q()
+            .then(function() {
+                return that.callHook("ebook:before", _options);
+            })
+            .then(function(_options) {
+                var command = [
+                    "ebook-convert",
+                    path.join(that.options.output, "SUMMARY.html"),
+                    path.join(that.options.output, "index."+format),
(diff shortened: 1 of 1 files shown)


xuv:
I'm in need also of some way to change or add ebook-convert options.
I've forked the main repo just for that. In my case especially for embedding fonts in a pdf.
Here are my changes. 
https://github.com/xuv/gitbook/commit/beedc8a090399bc27795c6208ed967d699e47ebb
Did not do a pull request to not conflict with this one. But I can do it if needed.

alexandreaquiles:
@xuv I think this PR should not be merged anymore:

1. This PR was done targeting an older version of Gitbook (1.5.0). Much has chaged since then.
2. The intent was to make a generic way to change Caliber's options. Changes that are more specific probably are better on a focused PR. Exemple: #456

PR #541, opened 2015-01-01
Add Dockerfile

humboldtux:
You can test it my image created with a similar Dockerfile there:

https://registry.hub.docker.com/u/humboldtux/gitbook/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,16 @@
+FROM node
+
+RUN apt-get -y update
+
+RUN npm install gitbook -g
+
+# Support gitbook pdf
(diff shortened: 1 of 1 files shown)


znmeb:
Thank you for this! I'm just getting started with Gitbook - what kind of user testing do you need?

humboldtux:
No partcular user testing needed, but if you want to test the Dockerfile you can build an image from it yourself, or use my existing image https://registry.hub.docker.com/u/humboldtux/gitbook/.

SamyPesse:
Sorry but I think this should not be in this repo, it can easily be done in another repo like `gitbook-image`.

Issue #504, opened 2014-11-22
localize table of content


piranna:
the table of content has its name hardcoded on the top in english, it should localize to the book language, or at least allow to override it.

Ebizproduction:
Is there a way to create a plugin to override this ?

AaronO:
@piranna @Ebizproduction In which output formats ?

Could you attach screenshots to clarify exactly what you're referring to ?

Ebizproduction:
I think we are both referring to : gitbook/theme/templates/ebook/summary.html (line 3)
when compiling to epub or pdf.

`{% block title %}Table of Contents | {{ title }}{% endblock %}``

Table of Contents is hard-coded, and i'd like to render it as "Table des matières" or "Sommaire" in French.

winniehell:
Yes, please give other languages a chance to use GitBook! :+1:
Actually, the problem seems to be located at [line 40](/GitbookIO/gitbook/blob/3a0f004/theme/templates/ebook/summary.html#L40). I could not find any documentation on how to overwrite templates. The [help](http://help.gitbook.io/book/themes.html) is not very helpful in that context. :disappointed:

piranna:
Yes, I was talking about that line

CoDEmanX:
I believe this issue can be closed, because it seems to be implemented:
https://github.com/GitbookIO/gitbook/pull/556/

Issue #64, opened 2014-04-09
i18n in templates


SamyPesse:
Add an option for building the book in a specific language and add default value for this option for multi-languages book.

winniehell:
Related to #504?

PR #556, opened 2015-01-15
i18n support

winniehell:
This introduces a translation mechanism for built-in themes as asked for in #64 and #504.

I am new to JavaScript, so I would be glad if anybody knows how to simplify `i18n/index.js`.

As an example for how translations are done, I added a file with German texts.


winniehell:
Closing the pull request because it is part of 2.0 now.

Issue #554, opened 2015-01-14
Install as project dependency?


jehoshua02:
I know the gitbook docs instruct to install gitbook globally, but is it possible to install it as a project dependency? You don't want to write the book against one version of gitbook and build with a non-backward compatible version later. Installing it globally, other developers have no idea what version of gitbook was originally used. Am I missing something?

AaronO:
@jehoshua02 We try to be as backwards compatible as possible.

To be entirely honestly, I don't think there's anything major that we've broke since the initial release, some minor features may have been dropped (in favor of plugins), but mostly we've added features, or improved options without breaking compatibility.

Backwards compatibility is extremely important for something that intends to last, such as a book format (like ours). It's highly unlikely that we break features before a `2.0.0` release (and even then we'll only do so where necessary).

As a testimony of backwards compatibility I tried building https://github.com/GitbookIO/javascript and https://github.com/GitbookIO/documentation with our `0.0.1` release (latest is `1.5.0`, 54 releases in total) 
 and they both work like a charm, maybe not as pretty and with less features but still work :) (no changes whatsoever ;) ).

That said if you want to build against a specific version of `gitbook`, you can use https://github.com/GitbookIO/grunt-gitbook.

We recommend to build against the latest release, and ideally using [GitBook.com](https://www.gitbook.com) where you get automated builds etc ... (it has great GitHub integration).

I don't think backwards compatibility is something you should be worried about, we've done a good job so far and I hope we'll continue to so in the future.

sindresorhus:
To be honest, having to install `gitbook` globally is a no go for me. I know you think you won't break anything in the future, but the smallest differences in versions will be a problem for users, which might depend on some behaviour of a bug you fixed. Especially a problem in this case as books last for a long time. I was involved in grunt from the start, and we did that mistake and had to pay for it later in breaking all peoples projects.

What most build-systems like grunt/gulp/etc do is having the globally installed binary as just a dumb runner for the local core dependency. I would recommend you do the same.

You might be able to use https://github.com/tkellen/node-liftoff

> I don't think backwards compatibility is something you should be worried about, we've done a good job so far and I hope we'll continue to do so in the future.

Unfortunately you can't always control this.

-

These people can attest to the above:

// @tkellen @shama @contra @phated

SamyPesse:
[...]

Issue #543, opened 2015-01-08
PDF images are not centered


petermichaux:
I don't know why #289 is closed. It is not fixed and I cannot reopen it.

When I create a PDF my images are not centered. I'm using the latest version of gitbook. I installed it yesterday.

Issue #523, opened 2014-12-15
PDF Table of Contents Output


eljefe6a:
The README.md for GitBook no longer talks about the PDF `"toc": false` property. It appears to be ignored now. Is there a new setting to turn it off?

hackhowtofaq:
Any news about this?

pparavicini:
I am using gitbook to create a template for writing proposals, which are rather short, and deliver them in pdf.  The Table Of Contents is a bit overkill. 

+1 for the ability to suppress it via configuration.

Issue #515, opened 2014-12-03
Support ReStructuredText (.rst) in addition to Markdown


webmaven:
http://docutils.sourceforge.net/rst.html

PR #562, opened 2015-01-22
Version 2.0.0

SamyPesse:
# Version 2.0.0

### How to test it?

Uninstall GitBook 1.x.x:

```
$ npm uninstall gitbook -g
```

Install **GitBook CLI**:

```
$ npm install gitbook-cli -g
```

You can now build a book using the version 2.0.0 (this will download the latest alpha version of GitBook):

```
$ gitbook build ./mybook
```

### Changes:

- [x] Externalize the CLI into another [module](https://github.com/GitbookIO/gitbook-cli) (#554)
	- [x] Build command
	- [x] Serve command
        - [x] eBook commands
        - [x] init command
- [x] Use `gitbook` from book.json to define the version of gitbook to use
- [x] Glossary terms are handled correctly in ebook
- [x] Custom styles to style ebook and website (#543, #523)
- [x] Content templating using nunjucks
- [x] Content reference
	- [x] Internal content reference using `{% include "./test.md" %}` (#226)
	- [x] External content reference using `{% include "git+https://github.com/GitBookIO/markdown.git/test.md#sha" %}` or using variables `{% include book.markdownBook+"/test.md#sha" %}`
- [x] i18n support (#556,  #64 and #504)
	- [x] i18n in website and ebook
	- [x] Use `language` parameter or detect for multilingual books
- [x] Other input markup than Markdown
	- [x] AsciiDoc (#239)
	- [x] Support other markdown extensions (`.md`, `.mdown`, `.markdown`) (#307 and #436)
	- [x] ReStructuredText (#515)
        - [x] Maybe LaTex (#151)
- [x] Nice URLs: `README.html` => `/` (really `/index.html`)
- [x] Allow structure to be configured, to change the readme, summary ans glossary files used (#392, #476 and #416)
- [x] Better tests
- [x] Add back `page:before` hook
- [x] Fix `segmentation fault: 11`, maybe due to asciidoc
- [x] Complete README documentation
- [x] Translate it (https://github.com/GitbookIO/gitbook/tree/version/2.0/theme/i18n)
- [x] Don't remove `.git` and `.svn` when cleaning output folder (#273)
- [x] Transform svg as png before converting to ebook (#190)
- [x] Download remote images in ebook format (#570)
- [x] Fail hard if book.json is invalid or plugins is loaded with error (#567)
- [x] Rewrite filename in ebook format (#585)

### Plugins Compatibility

- [x] MathJax
- [x] KaTeX
- [x] [Richquotes](https://github.com/erixtekila/gitbook-plugin-richquotes)

### Related:

- [gitbook-parsers](https://github.com/GitbookIO/gitbook-parsers)
- [gitbook-markdown](https://github.com/GitbookIO/gitbook-markdown)
- [gitbook-asciidoc](https://github.com/GitbookIO/gitbook-asciidoc)
- [gitbook-cli](https://github.com/GitbookIO/gitbook-cli)

### DIFF
--- a/lib/book.js
+++ b/lib/book.js
@@ -507,13 +507,20 @@ Book.prototype.parsePage = function(filename, options) {
         return interpolate(options.interpolateTemplate);
     })
 
+    // Prepare page markup
+    .then(function() {
+        return filetype.page.prepare(page.content)
+        .then(function(content) {
+            page.content = content;
+        });
+    })
+
     // Generate template
     .then(function() {
         return that.template.renderPage(page);
     })
 
     // Prepare and Parse markup
-    .then(filetype.page.prepare)
     .then(function(content) {
         page.content = content;
 
--- a/lib/generators/ebook.js
+++ b/lib/generators/ebook.js
@@ -64,6 +64,8 @@ Generator.prototype.finish = function() {
             "--comments": that.options.description,
             "--isbn": that.options.isbn,
             "--authors": that.options.author,
+            "--language": that.options.language,
+            "--book-producer": "GitBook",
             "--publisher": "GitBook",
             "--chapter": "descendant-or-self::*[contains(concat(' ', normalize-space(@class), ' '), ' book-chapter ')]",
             "--chapter-mark": "pagebreak",
@@ -90,6 +92,10 @@ Generator.prototype.finish = function() {
                 "--pdf-page-numbers": Boolean(pdfOptions.pageNumbers),
                 "--pdf-header-template": String(pdfOptions.headerTemplate) || "<p class='header'><span>"+that.options.title+"</span></p>",
                 "--pdf-footer-template": String(pdfOptions.footerTemplate) || "<p class='footer'><span>_SECTION_</span> <span style='float:right;'>_PAGENUM_</span></p>"
+            });
+        } else if (that.ebookFormat == "epub") {
+            _.extend(_options, {
+                "--dont-split-on-page-breaks": true
             });
         }
 
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
         "resolve": "0.6.3",
         "fs-extra": "0.16.3",
         "fstream-ignore": "1.0.2",
-        "gitbook-parsers": "0.5.0",
+        "gitbook-parsers": "0.5.1",
         "nunjucks": "git+https://github.com/mozilla/nunjucks.git#42d2b9f5c0ad5db8ca3a1fdbba5bde9fc4cc2c45",
         "nunjucks-autoescape": "0.1.1",
         "nunjucks-filter": "0.1.0",
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
         "resolve": "0.6.3",
         "fs-extra": "0.16.3",
         "fstream-ignore": "1.0.2",
-        "gitbook-parsers": "0.5.1",
+        "gitbook-parsers": "0.5.2",
         "nunjucks": "git+https://github.com/mozilla/nunjucks.git#42d2b9f5c0ad5db8ca3a1fdbba5bde9fc4cc2c45",
         "nunjucks-autoescape": "0.1.1",
         "nunjucks-filter": "0.1.0",


znmeb:
I like LaTeX but I definitely think it's a "nice-to-have". On the GitBook website it'll be OK, but the dependencies it pulls in on Windows or Linux workstations are huge - on the order of a gigabyte *minimum*.

That said, every machine I use has those already (plus Pandoc and *its* dependencies!) so it won't impact me if you do support LaTeX.

SamyPesse:
@znmeb LaTeX will not need to be a dependency of GitBook, it will be needed only if the author is using latex files.

Supporting LaTeX is not the priority of this release, but it'll now be possible.

winniehell:
I made some inline corrections for the German texts (f121635).

winniehell:
@SamyPesse no :beers: today anymore :wink: :zzz:

winniehell:
@SamyPesse :+1: (a little bit impressed that you do a complete rewrite within one week)

SamyPesse:
Thanks, I know perfectly the gitbook structure, so it's easy to rebuild it and make it better.

Is "Einführung" right for "Introduction" (https://github.com/GitbookIO/gitbook/commit/99d9d4961076424dc9681ee7f2ab2fed28d5d7c0#diff-d20357f2347472b13b032ff4c59f77e7R8) ?

erixtekila:
I may be interested to retrieve the current git version or an internal version  of the book in the javascript book.

One workflow I'm thinking of is to permit inline notes.
In order to keep them, I consider using local storage.
By the way, as the book may evolve, the notes may be wrongly associated later to some content.

One way to resolve this would be to associate a book with one version number. (may be semver also ?!)
and produce some kind of migration tool to export the notes if needed.

erixtekila:
ebook-convert produce a pdf that is not selectable.
See for yourself : each character is separated from its phrase context.
That produce a non exploitable pdf IMHO.

AaronO:
@SamyPesse It would be great, to avoid having to wrap chapters with ```{% raw %}...{% endraw %}``` for templates containing `{{` and `}}` which is a very common pattern especially for technical books (django, mustache, `nunjucks`, even our own doc ...)

The ideal behaviour would be that if the expression to display (in the ```{{ ... }}``) resolves to "nothing" then print it as it was `{{ ... }}`. I have to see if `nunjucks` allows that.

The other option would be to skip them inside `markdown` code blocks ...

What do you think ?

SamyPesse:
@AaronO I think this is a **bad** idea, because:
- You are focus on a very small usecase (probably less than `0.01%`)
- To solve this very small usecase, you are bringin more conditional complexity in the parsing pipeline, this will make it less easy to understand/reproduce; this will transform gitbook as a conditional mess.
- Your solution can probably handle `{{ .. }}`, but you'll have to support all the other templating syntax, and it'll not be as easy, for example for a condition `{% if book.newVersionAlpha %}`, how do you decide to process this condition or to display it? what if `newVersionAlpha` is defined only in the config of the book?

`{% raw %}` is easy to use, use the same syntax as blocks.

I think who are focus on this, because you just commit a small change on the documentation and you are focusing on your tiny issue, without understanding the bad consequence on bringing that kind of complexity into the simple generation pipeline.

Another solution can be a configuration in the `book.json` to disable tempalting, but I think we should just stick with `{% raw %}` because the problem doesn't exist for `99,99%` of the authors and it's not even a problem because the workaround is simple.

AaronO:
@SamyPesse I disagree,

It's used in "formatting" url examples for API docs etc ... : http://help.gitbook.com/build/push.html#git-url.

`{% raw %}` blocks aren't very obvious to be honest, I think people will just be really puzzled as to why GitBook isn't displaying what they expect and won't know about or think of `{% raw %}`.

Books on Go programming, Django programming (web programming in general), api docs etc ... will be effected by this suggesting that they patch their pages with `{% raw %}` just doesn't seem very friend to me, it's something we broke for them for no good reason (to those people in that case, it adds no benefit while adding the inconvenience of having to patch). 

And over that, that's the current behaviour right now : https://github.com/GitbookIO/gitbook/blob/master/lib/parse/include.js#L8.

This really shouldn't be complicated to fix and I think it makes it more obvious to see typos if the person is using variables. It doesn't at all have to be a big change in the pipeline, it should be two or three lines with no conditional, that's not going to turn GitBook into a conditional "mess" ...

The way I see it, there's only two possible reasons a person's `{{ x }}` expression would evaluate to "nothing" :
* They made a typo, for example : `{{ exmple }}` vs `{{ example }}`.
* They really want to display `{{ example }}` as part of a code snippet or whatever in their book

In both those cases displaying something blank is in my opinion useless. I'm more likely to notice an error of a misspelled variable if it still displays `{{ exmple }}`, so that's useful even when I'm actually wanting to use templating and when I actually don't want to that's good too.

What is there to lose and what's the cost ?

Disabling templating is extreme and not needed when this "fix" is both useful, simple and compatible with what's already out there.

SamyPesse:
With your solution, books about Django, Nunjucks,etc will still have issue, your solution will change nothing to the processing of `{% if .. %}, {% for %}, {% include %}, ...`. Your solution will only handle case of people writing simple `{{ ... }}` expression without knowing about templating.
Basically if the author is writing something else between `{% ... %}`, we'll still have to use `{% raw %}`.

You believe your solution is easy, but you'll have to handle correctly the expression parsing, for example: `{{ file.mtime|date("%d")|uppercase }}` or something like `{{ "Edited +"file.mtime|date("%d") if (file.mtime < now() - (10*3600)) else "No recently updated" }}`.
If your interpolation in the processing pipeline doesn't handle perfectly all the parsing cases, you'll cause more damage that you're trying to fix.

You are considering an hack to solve an UX problem (that you're the only one to encounter yet). I think it's pre-matured. The version `1.x.x` was full of hack of this type, I would prefer avoid going back this way.

Why not wait to get feedback from the authors about this? See how much users are having issue with the `{% raw %}` then setup a proper solution?

But instead of talking about this, you're sure of your solution, that it's clean, then just do a Pull-Request and we'll talk on it.

winniehell:
@AaronO Would disabling template parsing for book content by default and having to explicitly enable it solve your issue? I can imagine that people that make use of the templating syntax refer to the documentation beforehand (other than people using `{`s).

AaronO:
@SamyPesse  Input versions should be validated:

```
~/git/gitbook version/2.0*
❯ gitbook versions:link local .

~/git/gitbook version/2.0*
❯ gitbook versions

/opt/boxen/nodenv/versions/v0.10.17/lib/node_modules/gitbook-cli/node_modules/semver/semver.js:271
    throw new TypeError('Invalid Version: ' + version);
          ^
TypeError: Invalid Version: local
    at new SemVer (/opt/boxen/nodenv/versions/v0.10.17/lib/node_modules/gitbook-cli/node_modules/semver/semver.js:271:11)
    at compare (/opt/boxen/nodenv/versions/v0.10.17/lib/node_modules/gitbook-cli/node_modules/semver/semver.js:424:10)
    at gte (/opt/boxen/nodenv/versions/v0.10.17/lib/node_modules/gitbook-cli/node_modules/semver/semver.js:473:10)
    at cmp (/opt/boxen/nodenv/versions/v0.10.17/lib/node_modules/gitbook-cli/node_modules/semver/semver.js:490:22)
    at Comparator.test (/opt/boxen/nodenv/versions/v0.10.17/lib/node_modules/gitbook-cli/node_modules/semver/semver.js:560:10)
    at testSet (/opt/boxen/nodenv/versions/v0.10.17/lib/node_modules/gitbook-cli/node_modules/semver/semver.js:894:17)
    at Range.test (/opt/boxen/nodenv/versions/v0.10.17/lib/node_modules/gitbook-cli/node_modules/semver/semver.js:886:9)
    at Function.satisfies (/opt/boxen/nodenv/versions/v0.10.17/lib/node_modules/gitbook-cli/node_modules/semver/semver.js:907:16)
    at checkVersion (/opt/boxen/nodenv/versions/v0.10.17/lib/node_modules/gitbook-cli/lib/versions.js:21:16)
    at /opt/boxen/nodenv/versions/v0.10.17/lib/node_modules/gitbook-cli/lib/versions.js:46:10

~/git/gitbook version/2.0*
❯
```

SamyPesse:
@AaronO this is related to gitbook-cli, post an issue there.

PR #694, opened 2015-04-07
japan support added

awrelll:


SamyPesse:
Same as before, please add yourself as an author (for the language) at the same time :)

SamyPesse:
Thanks :+1: :japan:

PR #696, opened 2015-04-08
use the `escape-string-regexp` module

sindresorhus:
no point in hardcoding this.

https://github.com/sindresorhus/escape-string-regexp


SamyPesse:
Thanks :+1: :beers:

PR #700, opened 2015-04-10
Switch from gaze to chokidar.

paulmillr:
Hey there.

This commits switches `gaze` to a better file watcher, chokidar.

To summarize: chokidar is *very* stable, and is much faster on OSX, with less CPU usage because of `fsevents`.

Chokidar has been downloaded over 1.2M times last month and is used by very popular projects.


SamyPesse:
Thanks :+1: :rocket:

PR #743, opened 2015-05-12
Added vk sharing button.

chudaol:
Added sharing button for vk (https://vk.com/). Will only be enabled if links.share.vk is set to true (like weibo).


SamyPesse:
Thanks :+1:

chudaol:
Hi Samy! Is that possible that file "theme/javascript/utils/sharing.js" is not included in current builds?
Or maybe I've missed something in my code?
Because VK sharing button appears ok, but when you click it, the script breaks 'cause it doesn't find the corresponding function in types map, and in fact it is not there:

vkismissing

Thank you!

Issue #476, opened 2014-10-13
Don't set main README.md file as 'introduction' on the index


piranna:
README.md is added automatically, that not something so bad, but it's also added on the index as an introduction, that maybe it's not what you want, for example when using it to show info about how to generate the book. In my case (a degree thesis) I have a separate introduction chapter, so having it twice on the index looks ugly.

petermichaux:
I don't understand why the README.md file is required and why it is automatically included as the first item in the Table of Contents with the string "Introduction". I think it would be better that the README.md file is not included by default. It is very easy for authors to include it in the SUMMARY.md file if they want the README.md file included in some way. Much more flexible too.

piranna:
+1 to @petermichaux

gbonanome:
+1 to this feature, maybe an option inside `book.json`

CristianCantoro:
See also https://github.com/GitbookIO/gitbook/issues/392.

piranna:
I've found this problem also affect the Glossary, being it shown in english while the other entries of my table of contents are in spanish...

SamyPesse:
Just add `"language": "sp"` to your book.json file.

piranna:
Oh, great, thank you! :-)

Issue #416, opened 2014-08-20
readme.md and summary.md must be in capital letters


skilesare:
Ignore case when looking for files:

Tweaking book.json
Building MOBI, will upload
Starting build ...
ENOENT, open '/tmp/book/README.md'

when readme.md exists.  Changing case fixes the problem.

Issue #392, opened 2014-08-03
Add option to ignore README.md


davefp:
My book doesn't use README.md as the introduction. Instead, it's a guide to contributing, how to build the book, etc..

As such, I don't want it included in the summary at all. As far as I can tell this isn't possible. Adding the file to .bookignore causes the following error:

`ENOENT, stat '/Users/david/code/books/webhookbook/_book/index.html'`

My preferred solution would be an option entry in `book.json` called `ignoreReadme` or something similar that when set to true would simply let me skip adding the README to the summary.

gbonanome:
+1 to the feature.

petermichaux:
+1

SamyPesse:
Since version 2.0.0, this is fixed by configuration:

```
{
  "structure": {
    "readme": "OTHER_README.md"
  }
}
```

Abel-Torres:
Gitbook demands the README.md file. I don't want to use it. Why should the editor impose itself on my work?

piranna:
It's not the only place where this happens, it was not easier to use
directly GitHub repos for plugins... According to the authors, this is done
to make it easier to use by people that know nothing about Node.js and npm,
but this way of thinking complicates too much to us advance users and
gitbook architecture and development itself. What a shame :-(

SamyPesse:
@Abel-Torres You can use another file than `README.md` as introduction by modifying the `book.json`:

```
{
  "structure": {
    "readme": "OTHER_README.md"
  }
}
```

The editor should support it, but when opening the book it might try to always open the `README.md`. We are working on improving the editor to fix this kind of UX issues.

@piranna You can use GitHub repos for plugins now (starting with version `2.6.7`, and better support in `3.0.0`):

```js
{
    "plugins": ["test@git+https://github.com/You/myGitBookPlugin.git"]
}
```

piranna:
> @piranna You can use GitHub repos for plugins now (starting with version 2.6.7, and better support in 3.0.0):
> 
> {
>     "plugins": ["test@git+https://github.com/You/myGitBookPlugin.git"]
> }

Ok, good to know, last time I test it was on previous versions (2.0.x or so). SInce now I'm using 2.6.7 I'm going to give it a try :-)

SamyPesse:
@piranna If it's not working as expected, it will in version `3.0.0` (A few fixes have been implemented on this: support for SSH urls, etc).

piranna:
[...]

PR #747, opened 2015-05-14
Update README to clarify book context and options

chasballew:
This just demonstrates how to configure the book context with the changes from 2.0.0 (#562). It should also help clarify how #416 and #476 have been resolved.

I don't know that you can completely leave a structural component out (a la #392).


SamyPesse:
I'll add this chapter to the new documentation: http://toolchain.gitbook.com

PR #757, opened 2015-05-23
Add title for language overview

aschempp:
When having multilingual books, there's a problem with the book title:
- If you do not specify a `title` in your `book.json`, the language overview page will have an empty `<title>` attribute.
- If you do specify a title, it will be applied to all languages (=> untranslatable).

This PR attempts to solve this by
 1. Add a new option `langTitle` to `book.json`. If set, it will be used for the language overview `<title>` instead of the `title` option.
 2. Parse the `LANGS.md` and find the first headline as a title, same as done for the book content.

If neither `langTitle` nor `title` option is set and the `LANGS.md` does not have a headline, the behavior is not changed (=> empty `<title>` in language overview).

### DIFF
--- a/lib/book.js
+++ b/lib/book.js
@@ -416,6 +416,12 @@ Book.prototype.parseLangs = function() {
                     name: err.name || "Langs Parse Error",
                     fileName: that.langsFile
                 });
+            }).then(function(langlist) {
+                return langs.parser.readme(content)
+                .then(function(readme) {
+                    that.options.langTitle = that.options.langTitle || that.options.title || readme.title;
+                    return langlist;
+                });
             });
         });
     })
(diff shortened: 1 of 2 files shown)


SamyPesse:
I fixed it with https://github.com/GitbookIO/gitbook/commit/6dd7f6162f28830c334df8cf4c26926c20e72a02

You can create a `book.json` at the root of your multilingual book with a `title` attribute, this title will be used for the language overview page, then you can define a new title in a language`s book.json

aschempp:
Thanks, I didn't know I can have a books.json for each language too :-)

aschempp:
Oh, but shouldn't it choose the title in preference to the generic translation?

PR #758, opened 2015-05-23
Fix the X-UA-Compatible meta tag

### LINKED ISSUE #755: Incorrect X-UA-Compatible meta tag
I just stumbled upon this because [the html5validator](https://github.com/svenkreiss/html5validator) says
> A “meta” element with an “http-equiv” attribute whose value is “X-UA-Compatible” must have a “content” attribute with the value “IE=edge”

In #501 the respective meta tag was added. But according to my knowledge and [the Microsoft website](https://www.modern.ie/en-us/performance/how-to-use-x-ua-compatible) it should simply be

```html
<meta http-equiv="X-UA-Compatible" content="IE=edge">
```

/cc @tonka3000

aschempp:
This fixes #755

### DIFF
--- a/theme/templates/website/layout.html
+++ b/theme/templates/website/layout.html
@@ -4,7 +4,7 @@
     <head>
         {{ htmlSnippet("head:start")|default("")|safe }}
         <meta charset="UTF-8">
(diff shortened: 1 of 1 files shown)


SamyPesse:
Thanks again :+1:

PR #759, opened 2015-05-23
Move style sheets to the <head> section

### LINKED ISSUE #756: Stylesheets added to <body> instead of <head>
This results in invalid markup. I noticed this because [html5validator](https://github.com/svenkreiss/html5validator) reports
> Element “link” is missing required attribute “property”

fixing should be fairly easy by moving the `style` block:
https://github.com/GitbookIO/gitbook/blob/34fc2831e0cf0fed01c71cec28d93472d87f455b/theme/templates/ebook/layout.html#L16

**Background Info**:
http://stackoverflow.com/questions/18549726/element-link-is-missing-required-attribute-property
> Although a link element is unconditionally invalid in body in older HTML specifications, HTML5 has more permissive rules. According to HTML 5.1 Nightly (which is more or less what the validator tries to keep up with), the link element is allowed in the document body, too (wherever phrasing content is allowed), provided that it has an itemprop attribute.

aschempp:
This fixes #756


SamyPesse:
I'll take a look at all your PRs in the coming days :+1:

SamyPesse:
Thanks :+1:

PR #773, opened 2015-05-29
Aadd option for configFile

friedger:


kylebrowning:
I tried to fix the failing build here, #789

SamyPesse:
Closing this (more than one year after .. sorry), it can be done using `book.js`:

```
module.exports = (process.env.CONFIG_A? require('./book_configa.json') : require('./book_configb.json');
```

I'll advice authors to use a `book.json`, since the GitBook Editor doesn't support `book.js`

BTW: [GitBook.com](https://www.gitbook.com) also supports setting environment variables.

PR #789, opened 2015-06-05
Add option for config file

kylebrowning:
Attempts to fix #773


SamyPesse:
This can be done using a `book.js` which load another configuration file.

PR #806, opened 2015-06-18
Aria roles for accessibility

aschempp:
This pull requests adds semantic information for screen readers and thelike.
I was not sure if it's "allowed" to update the HTML markup to use HTML5 tags (e.g. `<main>` instead of `<div>`)…


SamyPesse:
Thanks :+1: (Sorry for the delay)

PR #813, opened 2015-06-30
Allowing for custom chapter-mark and page-break in PDF.

MrMaksimize:
I'm building a gitbook that needs to compile pretty cleanly to PDF, and the trouble I've been having is with line breaks, and chapter marks.  

I would love to be able able to specify these options in book.json, but because they're hard coded in the generator, I am unable to.  

So I thought I'd make this PR to get the convo started.  The natural progression of this though is be able to take a lot of the ebook-convert PDF options and be able to specify them in book.json.  But I didn't want to go down that route yet before discussing it with the project leads.


SamyPesse:
Please  add the default values to https://github.com/GitbookIO/gitbook/blob/master/lib/configuration.js#L235

MrMaksimize:
Amended the commit :)  Thanks for the heads up!

SamyPesse:
Thanks :+1: :beers:

MrMaksimize:
Tys! :beers:

PR #844, opened 2015-08-04
Fix #843 ignore _book in watch.js

### LINKED ISSUE #843: gitbook server always refreshing
If I put my book directory in Dropbox, and use `gitbook serve`. Initially, it works fine. But it'll always `Restart after change in files` after I change any character in one file.

I try to move book directory out of Dropbox or stop Dropbox, they both solve the issue.

I'm not sure it's issue of `chokidar` itself or Dropbox or gitbook.

xcv58:
Fix #843


SamyPesse:
Thanks you :+1: sorry for the late answer

PR #877, opened 2015-08-23
hebrew support theme

a-moses:
full Hebrew support for GITBOOK (with rtl )

* Sans and Serif may still in original name fonts


SamyPesse:
Please add yourself to the AUTHORS list: https://github.com/GitbookIO/gitbook/blob/master/AUTHORS.md#translators :smile:

PR #878, opened 2015-08-23
Add `npmConfig` field for configuration

jasonslyvia:
**Background**

Somewhere in this world there is a country which is not in favour of freely access of internet, many oversea websites have been blocked for no reason, and unfortunately npm is one of them.

So in this country when developer run `npm install`, they got stuck forever. Some warm hearted companies build mirrors of npm to make developers use npm without pain.

Since gitbook has encapsulated npm internally, the usage of `npm install --registry=http://kind-npm-mirror-url` will be unavailable. 

**What's been changed**

In `book.json`, added a new field call `npmConfig`, which will be passed to `npmi` directly as its `npmLoad` parameter.

### DIFF
--- a/lib/configuration.js
+++ b/lib/configuration.js
@@ -268,6 +268,11 @@ Configuration.DEFAULT = {
 
         //Footer HTML template. Available variables: _PAGENUM_, _TITLE_, _AUTHOR_ and _SECTION_.
         "footerTemplate": ""
+    },
+
+    // Options passed through to npm
+    "npmConfig": {
+        "registry": "https://registry.npmjs.org"
     }
 };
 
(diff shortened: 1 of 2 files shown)


jasonslyvia:
Turns out we can also use environment variable for npm config, so this PR isn't that important I guess.

[https://docs.npmjs.com/misc/config#environment-variables](https://docs.npmjs.com/misc/config#environment-variables)

SamyPesse:
Thanks @jasonslyvia 

@AaronO what do you think ? 

I personally think that if you can have the same result with environment variable, we should not add npm-specific fields in the `book.json`.

aniude:
still no method to resolve this issue? @SamyPesse

PR #919, opened 2015-09-12
Don't react when user wants to open a link in a new tab

### LINKED ISSUE #846: Cmd+Click does not work on navigation
Cmd+Click any item in the navigation to the left.
Expected: it opens in new tab.
Actual: Nah, it eats the click.

gaearon:
This fixes #846. The code is [based on equivalent React Router code](https://github.com/rackt/react-router/blob/bb645ee84e6e84936d5fe12ee6e0e4b01243a3f3/modules/Link.js#L6-L12) which has enjoyed a ton of production usage.

### DIFF
--- a/theme/javascript/core/navigation.js
+++ b/theme/javascript/core/navigation.js
@@ -105,7 +105,19 @@ define([
         events.trigger("page.change");
     };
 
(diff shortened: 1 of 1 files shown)


SamyPesse:
Thanks :+1:

PR #928, opened 2015-09-14
Code highlighting extended by plugins

### LINKED ISSUE #918: Make code highlighter configurable
Currently highlight.js is used, but I don't quite like the way it highlights JS, and in addition, it has multiple problems with JSX. I would much rather use [Prism](http://prismjs.com).

It does not appear that this is configurable. (I don't want to create a plugin with a custom tag because that would break vanilla doc display on the Github repo.)

How would I go about switching to Prism? For now, I'll probably fork, but it would be nice to allow this to be configurable.

SamyPesse:
This PR improves how code highlighting is done in GitBook by using a `code` block instead of `highlight.js`. It fixes #918.

Plugin can extend blocks and define this `code` block to provide their own code highlighting.
GitBook will use by default [plugin-highlight](https://github.com/GitbookIO/plugin-highlight), which is equivalent to the code highlighter used in `gitbook<2.4.0`.

Implementing a code highlighter is really easy, here is a simple example:

```js
module.exports = {
    blocks: {
         'code': function(block) {
                return prims(block.body, block.kwargs.language);
         }
    }
};
```

This also makes it easy for authors to disable code highlighting and reduce css sizes (not really important).


SamyPesse:
@gaearonYou'll soon be able to improve  [gitbook-plugin-prism](https://github.com/gaearon/gitbook-plugin-prism), don't forget to add to future release in your `package.json`:

```js
"engines": {
    "gitbook": ">=2.4.0"
}
```

GitBook (event version < 2.4.0) will take care of downloading the right version of the plugin.

SamyPesse:
@AaronO tests are failing because of that [line](https://github.com/GitbookIO/plugin-highlight/blob/master/package.json#L7), passing locally when changing the version.

I'll merge this PR just before publishing a new release.

gaearon:
Awesome, thank you for working on this!

SamyPesse:
Let me know if you need help adapting the prism plugin. This should be a good example of how to do it: https://github.com/GitbookIO/plugin-highlight/blob/master/index.js

gaearon:
I'd appreciate a PR if you find time!

SamyPesse:
@gaearon Done: https://github.com/gaearon/gitbook-plugin-prism/pull/3

PR #937, opened 2015-09-17
Fixed libpng warning: iCCP with Libpng1.6+ .

iamchenxin:
Fixed libpng warning: iCCP with Libpng1.6+(linux) which caused by apple-touch-icon-precomposed-152.png .
this occurs when use gitbook-cli in console like this(or anything else will use Libpng1.6+)
```sh
gitbook pdf ./ ./test.pdf
```
Simply remove the embedded profile in this png.
More detail see this :https://wiki.archlinux.org/index.php/Libpng_errors.

PS: Maybe the origin png is transparent,i do not familiar with graphic editor in linux,so someone should convert it again. to transparent?


SamyPesse:
I'll take a look at this PNG file in the new external default theme.

PR #1018, opened 2015-11-16
Update README.md

atilacamurca:
There is an extra space in the command line example that enable debug mode.


SamyPesse:
Thanks :+1:

PR #1020, opened 2015-11-17
Fix small font rendering issue on Macs

dylang:
Removed the `antialias` setting which tells Chrome to not use native font rendering and instead use something from the late 2010's that isn't subpixel-friendly and looks very blocky. This used to be required when OS X browsers didn't antialias by default because it made fonts look bolder than on Windows. This has since been fixed and we are advised to not use `-webkit-font-smooth`. 

https://developer.mozilla.org/en-US/docs/Web/CSS/font-smooth

## Before
screen shot 2015-11-17 at 4 14 36 pm

## After
screen shot 2015-11-17 at 4 14 44 pm

As you can see this is very apparent with small fonts. The above is rendered with gitbook-plugin-mermaid, which has the same issue, and I've provided a PR that for that repo as well: https://github.com/JozoVilcek/gitbook-plugin-mermaid/pull/14

*Note: The above example will only be fixed when both repos has removed `-webkit-font-smothing: antialias` from their CSS.*

Thanks!

### DIFF
--- a/theme/stylesheets/website.less
+++ b/theme/stylesheets/website.less
@@ -22,7 +22,6 @@
     -webkit-tap-highlight-color: transparent;
     -webkit-text-size-adjust: none;
     -webkit-touch-callout: none;
-    -webkit-font-smoothing: antialiased;
 }
 
 a {
@@ -39,7 +38,6 @@ html {
 
 body {
     text-rendering: optimizeLegibility;
-    font-smoothing: antialiased;
     font-family: @font-family-base;
     font-size: @font-size-base;
     letter-spacing: .2px;


SamyPesse:
@dylang Thank you,

I'll take a look at it (test the rendering) before the next release :)

SamyPesse:
I'm sorry but we'll keep with `-webkit-font-smoothing: antialiased;` for now.

On OS X with Safari, it render terribly without it; and websites like Medium are still using it.

## With -webkit-font-smoothing: antialiased;

screen shot 2015-11-25 at 10 36 15

## Without

screen shot 2015-11-25 at 10 36 26

dylang:
@SamyPesse Thanks for considering it and giving it a test. I also appreciate that you looked at other sites as well like Medium. I admit I only tested this change in Chrome, and my primary concern was the small text produced by the mermaid plugin. I'll see if there's a better way to go about that particular issue. 

Thanks again!

SamyPesse:
You're welcome :beers: 

You can disable it for this plugin using something like:

```css
.mermaid {
     -webkit-font-smothing: initial;
}
```

dylang:
:beers:

PR #1065, opened 2015-12-20
align lrport on backend & front-end

cades:
this commit contains 2 changes:

- upgrade plugin-livereload to 0.1.0, which supports custom port.
- add config for plugin-livereload, set port to kwargs.lrport

therefore lrport of both backend & front-end are aligned.


cades:
livereload plugin ver 0.1.0 is [on the way](https://github.com/GitbookIO/plugin-livereload/pull/1), let's wait for this plugin's next release.

PR #1082, opened 2016-01-06
add ebook-convert pdf options 'custom-size' and 'unit'

rockitbaby:
description from http://manual.calibre-ebook.com/cli/ebook-convert.html#pdf-output-options

--custom-size
Custom size of the document. Use the form widthxheight EG. 123x321 to specify the width and height. This overrides any specified paper-size.

--unit, -u
The unit of measure for page sizes. Default is inch. Choices are [‘millimeter’, ‘centimeter’, ‘point’, ‘inch’, ‘pica’, ‘didot’, ‘cicero’, ‘devicepixel’] Note: This does not override the unit for margins!


SamyPesse:
Sorry for the long delay ...

Why do you want to set `unit` ? I understand the goal with setting `custom-size`.

PR #1115, opened 2016-01-26
Fix isMobile(): Use .width() instead of obsolete document.width

kliput:
The ``isMobile()`` function for detecting small-width window size was not working on modern browsers, because of obsolete ``document.width``: https://developer.mozilla.org/en-US/docs/Web/API/Document/width

I think that using jQuery is the safest option.

### DIFF
--- a/theme/javascript/platform.js
+++ b/theme/javascript/platform.js
@@ -1,5 +1,5 @@
 module.exports = {
     isMobile: function() {
-        return (document.width <= 600);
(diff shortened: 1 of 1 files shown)


kliput:
Hi, is anything wrong with this fix or there is simply no time for merging this :)?

SamyPesse:
I fixed it in the new theme for GitBook v3.

Thnak you :beers:

PR #1139, opened 2016-02-12
Created a glossaryExclusions config item

Perturbatio:
modified page.js to use this instead of hardcoding the elements that will match the exclusions

used by page.js to determine which elements will not have their glossary matches
marked (useful for excluding headings 'h1', 'h2', 'h3', 'h4', 'h5', 'h6' for instance)


SamyPesse:
Thank you :beers: 

Headings are not annotated anymore in GitBook v3, and I've added a class `.no-glossary` to disable annotations on elements created by plugins.

But I don't think it's a good idea to have an explicit configuration for this.

PR #1109, opened 2016-01-21
Version 3.0.0

SamyPesse:
- [x] Simpler JS APIs (in GitBook itself and plugins)
- [x] Improvements for Glossary (`GLOSSARY.md`)
     - [x] Generate glossary as a normal page
     - [x] Add class `.no-glossary` to avoid annotating in it
     - [x] Don't annotate in headings
- [ ] Improvements for Languages (`LANGS.md`)
    - [ ] Fix #788
    - [x] Fix #744
    - [x] Remove plugins from language books (list of plugins should be global for the entire book)
- [x] Improvements for Summary (`SUMMARY.md`)
     - [x] Allow anchor entries, ex: `* [Entry](test.md#hello)` (Fix #776)
     - [x] Allow multiple parts/sections in SUMMARY
     - [x] Allow external entries
     - [x] No longer required
- [x] Page
     - [x] Optmize urls when links to a `README.md` (Fix #537)
     - [x] Text direction can change between pages (`ltr` to `rtl`)
- [x] Templating
     - [ ] Allow single tag block (exemple for `codesnippet` plugin)
     - [x] Cleanup context
     - [ ] Update to `nunjucks@3`
     - [x] date filters
- [x] Documentation use-case, where book is stored in a `docs` folder
- [x] Theming
     - [x] Default theme should be stored in an external repository (https://github.com/GitbookIO/theme-default)
     - [x] Author can specify a theme in its `book.json`
     - [x] Author can extend his theme by creating a `_layouts` folder
     - [x] i18n templating filter  in the core
     - [x] Merge templating engine with the book's one
- Markdown
     - [x] Switch to `markdown-it` instead of `kramed` ?
            - [ ] `[ALink]:..` no supported (https://gist.github.com/SamyPesse/c2875343878f21338b14)
- Tests
     - [ ] Fix tests on Windows

#### Change Log

https://github.com/GitbookIO/gitbook/blob/3.0.0/CHANGES.md

PR #1149, opened 2016-02-18
Transfer files one by one insted of all at the same time

avastms:
When the amount of pending transfer files is large, copying them
all at the same time cause Node to quit on error
	ENFILE: File table overflow.
Do it slowly, we are not in a hurry, are we?


SamyPesse:
GitBook v3 is doing this (https://github.com/GitbookIO/gitbook/pull/1109)

PR #1159, opened 2016-02-24
Fix #697: add support for internal absolute links

### LINKED ISSUE #697: Is it possible to add internal absolute links?
How can one make an absolute internal link?

If I'm in the file `/foo/bar.md`, and I write the following:

```
[my link](/absolute/link/to/file.md)
```

The link that is created is to:

```
/foo/absolute/link/to/file.html
```

and not:

```
/absolute/link/to/file.html
```

Absolute links are important because they are easier to update site wide than relative links when there are large scale folder restructurings.

mastergenius:
Correctly handle absolute internal links

### DIFF
--- a/lib/utils/page.js
+++ b/lib/utils/page.js
@@ -210,7 +210,8 @@ function normalizeHtml(src, options) {
             var anchor = parts.hash || '';
 
             // Calcul absolute path for this file (without the anchor)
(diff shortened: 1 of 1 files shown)


SamyPesse:
It should be fixed in #1109 (I'll add a test to confirm).

Thank you!

PR #1192, opened 2016-03-24
Limit splitting plugin name for git with authentication

jashaj:
URLs of plugins from a git repo that are fetched via ssh contain an `@` Example plugin config: `ga@git@github.com:GitbookIO/plugin-ga.git#master`. 

This fails to load because the plugin name is split from the "version" on `@` without a delimiter.


SamyPesse:
Awesome, thanks :+1:

PR #1194, opened 2016-03-24
Fix for incorrect splitting the plugin name/version

jashaj:
In the previous pull request I had misinterpreted the second argument of String.prototype.split. The remainder of the splitted string must be joined instead of delimiting the number of splits.
The tests were all passing because the plugin in the test is also available in npmjs. 
My apologies for the extra work.


SamyPesse:
Yes ... I should have take more time to review it, my bad :)

Thank you for the fix :+1:

PR #1221, opened 2016-04-11
Added Dutch translation

alardw:


SamyPesse:
Thank you, but this PR should be done on https://github.com/GitbookIO/theme-default

PR #1275, opened 2016-05-05
Allow summary content references

jpreynat:


SamyPesse:
The current implementation will prevent the use of GitBook in the browser (editor), the parser should **never** generate files (or use the `utils/fs`, `utils/git` modules).

- The generation of these files should be during output in **the output directory** (no file should be written in the input folder).
- Also instead of generating a file containing `{% include ... %}`, we can directly use the `Git` module to access the file content using `git.resolve('...')`.
- The `Git.isUrl` from `utils/git` should be moved to `utils/location` so that it can be used in the browser environment
- The `SummaryArticle` model should have a method `isConRef` or `isGitUrl`
- I think most of the changes should be done in [lib/output/generatePages.js](https://github.com/GitbookIO/gitbook/blob/summary/conrefs/lib/output/generatePages.js), summary articles with `isGitUrl() == true` should also be added as pages, then instead of reading from the local folder, we should read using the `git.resolve` method

jpreynat:
Thanks for the remarks @SamyPesse.
I made the adaptations to actually read the page content from git outside of the parser scope.

Currently missing:
* [ ] Get the actual page's filepath when encoding a summary article which is a content reference in `json/encodeSummaryArticle()`: the filename is generated in `output/readPage()`, and link to the newly created page, but we can't modify the summary to reflect the actual page's file's path.

lib/output/generatePage.js
```diff
@@ -15,13 +15,13 @@ var callPageHook = require('./callPageHook');
     @param {Page} page
     @return {Promise<Page>}
 */
-function generatePage(output, page) {
+function generatePage(output, page, pagePath) {
```
SamyPesse: I don't think this is a good idea to pass the `pagePath` as a second argument to all functions, we can change the data structure for `Page` to more suitable to this new feature instead.

SamyPesse:
1) I'm not sure I understand the precise process when my summary include a conref.

Let's take as an example:

```md
# Summary

* [Normal Page](page.md)
* [Git ref](git+https://mygit.com/repo.git/myPage.md)
```

Can you list the different tasks to generate the book correctly.


2) (Related to my inline comment) I think the cleaner implementer should adapt the `Page` model, the current structure is:

```js
{
    file:           File(),

    // Attributes extracted from the YAML header
    attributes:     Immutable.Map(),

    // Content of the page
    content:        String(),

    // Direction of the text
    dir:            String('ltr')
}
```

It can be rethink to handle/represent conref pages.

jpreynat:
Starting from your example:

```md
# Summary

* [Normal Page](page.md)
* [Git ref](git+https://mygit.com/repo.git/myPage.md)
```

When generating the book, we would first create the actual pages list for the output using `preparePages()` which calls `parse.parsePagesList()`, resulting in:

```js
output.pages = {
  'page.md': PageObject1
};
```

Then add the content references pages using `prepareConRefs()`, resulting in:

```js
output.pages = {
  'page.md': PageObject1,
  'git+https://mygit.com/repo.git/myPage.md': PageObject2
}
```

Finally, actually generating the pages executes:
`generatePages()`

-> Got `PageObject1`
-> `generatePage(output, PageObject1, 'page.md')`
-> `readPage(output, PageObject1, 'page.md')`
`'page.md'` is not a conref
-> `parse.parsePage(book, PageObject1)`
-> continue `generatePage()` for this page (page hooks, template rendering, etc...)
...

-> Got `PageObject2`
-> `generatePage(output, PageObject2, 'git+https://mygit.com/repo.git/myPage.md')``
-> `readPage(output, PageObject2, 'git+https://mygit.com/repo.git/myPage.md')``
`'git+https://mygit.com/repo.git/myPage.md'` is a conref
-> `git.resolve('git+https://mygit.com/repo.git/myPage.md')`
-> Read content of the file and generate a new filename for the output for this file
-> Update `PageObject2` with content and new filename
-> continue `generatePage()` for this page (page hooks, template rendering, etc...)
...

Regarding your inline comment, we can indeed embed the page's ref from the summary into the `Page` object. We can also add a method `Page.isConRef()`, using `location.isGitUrl()` against the page's ref to determine if the page is actually a content reference.

The remaining problem is to actually find a way to point the `SummaryArticle` object's path to the actual file.

PR #1278, opened 2016-05-06
bump bash-color to version that has a license

dylang:
Version currently being used does not have a license making it difficult for companies to use.


SamyPesse:
I merged the other PR very fast, and it seems that `bash-color@0.0.4` is not published on NPM

SamyPesse:
Thanks 👍

Issue #1281, opened 2016-05-07
Passing Other Options to ebook-convert command?


cycomachead:
Is there a recommended way to pass custom options to `ebook-convert`? I'm looking for more advanced controls when generating PDFs.

It seems like the only defined options are in this file?
https://github.com/GitbookIO/gitbook/blob/36b49c66c6b75515bc84dd678fd52121a313e8d2/lib/output/ebook/getConvertOptions.js

Essentially, what I'd like to see is a more generic way of specifying the raw options myself, since some would be useful, and it seems wasteful for GitBook to duplicate all the options. I think that connecting GitBook's options more directly to the ones in `ebook-convert` would clarify what the options do. I also think it'd be helpful to link the docs to the ebook-convert ones. (https://manual.calibre-ebook.com/generated/en/ebook-convert.html#pdf-input-options)

SamyPesse:
I'm wondering if GitBook should allow passing options directly to `ebook-convert`, it will break if we decide to switch to another PDF builder (`ebook-convert` has a lot of issues).

What options are you looking to use?
We can maybe expose them in GitBook, but I'm not a big fan of having a `ebookConvertOptions` property.

cycomachead:
Yeah, I see the point about not having a hard link to the converter... My initial thought was that `ebook-convert` provides some options which are helpful, and could be nice to pass to the web builder. 

I've been experimenting with two options primarily lately, but I might try getting into more:
* embed fonts option
* generate PDF TOC. (The `ebook-convert` one has page numbers which are pretty important for me, even if it's not styled correctly.)

There are a few more options that I might try experimenting with, however.

Another way of thinking about this would be if I could do:
`gitbook pdf --ebook-covert '--....ebook convert options....' ` or something along those lines.

However, the other minor issue is that it would be helpful if the options in `pdf: {}` were documented a little better. (It took me a while to figure out that the margin values were in `point`s, for example.)
I wrote a short bit about that in the PR, but I'd be happy to write a couple sentence about the current options in a (separate) PR if you'd like.

xuv:
@SamyPesse Jumping in, but would love to see this feature added. 
I use a lot of the 'ebook-convert' options to get more control on typography outputs in the pdf. 
As an example, here is a list of the options I use:
```
'--embed-all-fonts': true,
'--keep-ligatures':true,
'--subset-embedded-fonts': true ,
'--disable-font-rescaling': true
```

SeaDude:
[...]

PR #1282, opened 2016-05-07
Allow the ability to pass specific options to ebook-convert

cycomachead:
This is a first pass at #1281, though I haven't had a chance to fully test it on my book yet...


xuv:
Thanks for this. Have not tested this code, but would love to have that feature added to Gitbook.
I use a lot of the 'ebook-convert' options to get more control on typography outputs in the pdf. 
As an example, here is a list of the options I use:
```
'--embed-all-fonts': true,
'--keep-ligatures':true,
'--subset-embedded-fonts': true ,
'--disable-font-rescaling': true
```

PR #1285, opened 2016-05-08
[WIP] Api for plugins to access summary

SamyPesse:
👷 WIP

This PR is related to #392.

It adds a few API for plugins to access the summary tree:

```js
// Walk over all article of the summary, stop when iterator returns false
this.summary.walk(function(article) {
    ...
});

// Article getters
var article = this.summary.getArticleByPath('README.md');
var article = this.summary.getArticleByLevel('1.1.0');
```

`article` is a JSON object with the same property as in the JSON output:

```
{
     "level": "1.0",
     "title": "Introduction",
     "ref": "README.md",

     // Deconstructed from "ref":
     "path": "README.md",
     "url": undefined,
     "anchor": undefined
}
```

### DIFF
--- a/lib/api/encodeGlobal.js
+++ b/lib/api/encodeGlobal.js
@@ -5,11 +5,13 @@ var fs = require('../utils/fs');
 
 var Plugins = require('../plugins');
 var deprecate = require('./deprecate');
-var encodeConfig = require('./encodeConfig');
-var encodeNavigation = require('./encodeNavigation');
 var fileToURL = require('../output/helper/fileToURL');
 var defaultBlocks = require('../constants/defaultBlocks');
 var gitbook = require('../gitbook');
+
+var encodeConfig = require('./encodeConfig');
+var encodeSummary = require('./encodeSummary');
+var encodeNavigation = require('./encodeNavigation');
 
 /**
     Encode a global context into a JS object
@@ -29,6 +31,7 @@ function encodeGlobal(output) {
     var result = {
         log: logger,
         config: encodeConfig(output, book.getConfig()),
+        summary: encodeSummary(output. book.getSummary()),
 
         /**
             Check if the book is a multilingual book
--- a/lib/api/encodeSummary.js
+++ b/lib/api/encodeSummary.js
@@ -10,7 +10,7 @@ var encodeSummaryArticle = require('../json/encodeSummaryArticle');
 function encodeSummary(output, summary) {
     var result = {
         /**
-            Iterate over the summary
+            Iterate over the summary, it stops when the "iter" returns false
 
             @param {Function} iter
         */
--- a/lib/api/encodeSummary.js
+++ b/lib/api/encodeSummary.js
@@ -1,0 +1,40 @@
+var encodeSummaryArticle = require('../json/encodeSummaryArticle');
+
+/**
+    Encode summary to provide an API to plugin
+
+    @param {Output} output
+    @param {Config} config
+    @return {Object}
+*/
+function encodeSummary(output, summary) {
+    var result = {
+        /**
+            Iterate over the summary
+
+            @param {Function} iter
+        */
+        walk: function (iter) {
+            summary.getArticle(function(article) {
+                var jsonArticle = encodeSummaryArticle(article, false);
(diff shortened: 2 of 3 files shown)


SamyPesse:
@todvora and @cycomachead, will it be enough or is there any required method missing?

@todvora, The implementation in https://github.com/todvora/gitbook-plugin-image-captions/blob/82161381edbc1e7ab0d685f188930ae8a8800bb7/src/lib/ImageCaptions3.js#L28 should be adapted, it's using some private method (like `.read()` of a page).
Why do you need to read all pages in the `init` hook? Basically I'm trying to understand the overall workflow of this plugin to find the right implementation using the GitBook v3 APIs.

todvora:
Hi @SamyPesse, the plugin needs to read and parse all images of all ordered pages, to be able to assign them a global index in the book (think captions like 'Figure 3. - database structure.'). The second feature is to build a [registry of all images](https://github.com/todvora/gitbook-plugin-image-captions#image-list), rendered inside any page. 

This cannot be done in the ```page``` hook, because promises doesn't guarantee any order (and is too late for an images list). It would be needed to execute all the ```page``` hooks in order, serialized, to replace the current functionality in ```init``` (and assume, that the image list is rendered after all pages containing images). 

I would actually prefer to replace all the captions on the fly, without the ```init``` hook and collecting images, if it would be possible somehow. 

To the API draft: it looks good, I am missing just this ```.read()``` method, to be able to reach the raw content or parsed HTML of a page inside current ```init``` hook.

Thanks!

cycomachead:
Thanks for this awesome, and quick, work!

Similar to @todvora's plugin (which I'm also using), I'd like to adapt one for references to use markdown citations. (Like `[text][ref]` format.) I'm currently playing around with the [bibtex-citation](https://github.com/manuelmitasch/gitbook-plugin-bibtex-citation) which uses a custom function `{{ x | cite }}` to get around the ordering problem, but this is rather awkward to read (markdown) and write.

SamyPesse:
@todvora There is already a `this.readFileAsString` [method](https://github.com/GitbookIO/gitbook/blob/master/lib/api/encodeGlobal.js#L67).

Here is an example to read the readme:

```js
var article = this.summary.getArticleByLevel('1.1.0');

this.readFileAsString(article.path)
.then(function(content) {
    ...
});
```

So basically you need to index the images from the page before generating each one?

---

@cycomachead You're trying to extend the markdown syntax to avoid using `{{ x | cite }}` ? There is no "correct way to do this in GitBook right now.
But the solution might be to use the `page:before` hook to modify the markdown content:


```js
"page:before": function(page) {
    page.content = page.content.replace(...);

    return page;
}
```

todvora:
Oh, I see, thanks. ```readFileAsString``` is what I am looking for. 

Yes, I need to index all images from all pages, before modifying every page later in the ```page``` hook. Maybe if I had a chance to preprocess all pages one by one, async, parallel) in ```page:before```, where the info contained in the JSON ```article```(level, ref, ...) would be available. But that would require to finish all ```page:before``` calls before first ```page``` hook...

cycomachead:
Thank's I'll take a look at the `page:before` option, but I think I also need the order to be guaranteed. What I'm looking for is a global counter of all links in the book, that could then be formatted properly, and similar to figures can be used to build an index page.

For reference, my (very WIP) book is https://cycomachead.gitbooks.io/thesis/content/index.html

SamyPesse:
@todvora @cycomachead Is there still API method that you need?

Let me know if you need help adapting your plugins.

todvora:
Works for me, thank you!

I have still some doubts about the new levels numbering, but it's mostly the plugin's internal issue. 

These levels are used as indexes in configuration ([example](https://github.com/todvora/gitbook-plugin-image-captions/#image-specific-captions)).  But I is [expected to be fragile](https://github.com/todvora/gitbook-plugin-image-captions/issues/2#issuecomment-159955063) and maybe we should find some other way, how to reference images from config.

Anyway I am OK now with ```walk``` and ```readFileAsString``` methods and I don't need any additional changes now. Thank you for your support and fast implementation.

PR #1302, opened 2016-05-12
Number Single Level books as multiple levels.

### LINKED ISSUE #1301: 3 pre.14: All items start with level 1.x
When I view the level numbers, everything starts at a 1.x level but the summary file doesn't have any nesting. To my eyes it looks just like there's an extra "1." before every item.
screen shot 2016-05-12 at 2 20 45 am

cycomachead:
Fix #1301

This makes it so that a single level book will have chapters show as
"1.", "2." and so on, making single part books appear the same as
before.

Note that the smallest way to make this change causes the `level`
property to be an integer for single part books. I'm not 100% sure this
is desirable, but so far it didn't seem to cause complications.


SamyPesse:
Thanks, there are some indentation issues (we are using 4 spaces).

Also can you add a test in `lib/models/__tests__/summary.js`?

cycomachead:
> Thanks, there are some indentation issues (we are using 4 spaces).

Not sure what happened there... I'll fix that up.

> Also can you add a test in lib/models/__tests__/summary.js?

Sure. I'll probably get to that in a day or two.

Soreine:
Levels are used inside the library to navigate the `Summary`. If we are going this route, we need to make sure the levels are consistent with the `Summary` structure. As long as the articles are within a single, untitled part, the levels should render that.
Your initial issue is not a logic matter, but a display matter. So I'll suggest to fix this at the output generation level. For convenience, we could provide some function with the `Summary` to return a _displayed level_.

```ts
/**
 * @param {Summary} summary
 * @param {SummaryArticle | SummaryPart} entry
 * @return {String} The displayed level (1 instead of 1.1 when adequate)
 */
function displayedLevel(summary, entry) {
    ...
}
```

I wonder what should be the behavior for books with a summary like this:

```md
# Summary

* Chapter 1
* Chapter 2

## Appendix Part

* Chapter 3
```

Something like that ? : 

```
1.1 Chapter 1
1.2 Chapter 2
2. Appendix Part
2.1 Chapter 3
```

noerw:
any progress on this issue?

zq99299:
啥时候能解决哇。现在的 1.1 太难受了
When can solve wow. Now 1.1 too uncomfortable

colingonglin:
Almost one year passed...
When will this issue be resolved?

pablo14:
Dear @cycomachead  do you have a workaround to avoid this issue? Thanks:)

huarse:
+1, `1.1`, `1.2` is so uncomfortable, I prefer to `1`, `2`

crifan:
+1
Hope fix it ASAP！

wajchina:
This issue was not resoled yet, is there any progress on this issue?

Issue #1093, opened 2016-01-11
Is there a way to create plugins that does not require you publishing them to NPM?


dionridley:
I'm trying to create a plugin but would like to avoid publishing to NPM because it is specific to our company.  I tried to create the node_modules folder and build my plugin identical to the way gitbook install sets them up in your project but did not have any success.

The error message is that it must be installed with gitbook install.  Is there any work around for this?

SamyPesse:
You can potentially use a GIT url in the `book.json`, since `gitbook install` is running the NPM command to install it.

Try something like:

```
{
    "plugins": [
        {
            "name": "myplugin",
            "version": "git+https://github.com/YourCompany/myplugin.git"
        }
    ]
}
```

Not sure if it'll work, but I can make the require changes to make it work in the future versions of GitBook.

dionridley:
Thanks, this solution seems to work for us.

AaronO:
@dionridley I just wanted to add that you can also simply version your plugin in the git repo by keeping in under `node_modules`.

You'll have to correctly tweak your `.gitignore` obviously, but it should work.

uliska:
There is another way to use non-published gitbook plugins:

In your plugin's `package.json` there must be a

```json
  "engines": {
    "gitbook": "*"
  },
```
entry. If that's not present Gitbook won't recognize it and ask you to install using `gitbook install`. So if your plugin contains that entry Gitbook won't try to install it.

What I do is *not* to actually have the plugin folder within the book but somewhere else and use a symlink in `path/to/your/book/node_modules`.

PR #1334, opened 2016-05-26
Fixed two issues for handling git URLs for plugins

ryanswanson:
Fixed two issues for handling git URLs for plugins that are published to git repositories rather than on npmjs.org.

Fixed issue in pluginDependency to dereference 'name' and 'version' using Immutable Map.get(...) accessors since config values are now wrapped by Immutable.fromJS(...) in config.js > setValue(...).  Added associated unit test.

Fixed issue in resolveVersion where a plugin may be using a git URL rather than a semver for the version portion of the plugin config definition.  In addition, refactored the resolveVersion function into a new module to allow for unit testing.  Added associated unit test.

### DIFF
--- a/lib/models/__tests__/pluginDependency.js
+++ b/lib/models/__tests__/pluginDependency.js
@@ -42,21 +42,39 @@ describe('PluginDependency', function() {
                 expect(plugin.getVersion()).toBe('git+ssh://samy@github.com/GitbookIO/plugin-ga.git');
             });
         });
-    });
 
-    describe('listToArray', function() {
-        var list = PluginDependency.listToArray(Immutable.List([
-            PluginDependency.createFromString('hello@1.0.0'),
-            PluginDependency.createFromString('noversion'),
-            PluginDependency.createFromString('-disabled')
-        ]));
+        describe('listToArray', function() {
+            it('must create an array from a list of plugin dependencies', function() {
+                var list = PluginDependency.listToArray(Immutable.List([
+                    PluginDependency.createFromString('hello@1.0.0'),
+                    PluginDependency.createFromString('noversion'),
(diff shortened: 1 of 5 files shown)


ryanswanson:
**Background**
In creating my first gitbook plugin, I initially developed locally using `npm link gitbook-plugin-myplugin` for the dependency.  My book.json in the main project could simply list `'myplugin'` as a string.

Once I was ready to publish to a private git URL to share with our team, I found that the [object-based configuration with name/value keys described by Samy](https://github.com/GitbookIO/gitbook/issues/1093#issuecomment-170586843) appeared to no longer function - yielding the following error message:
```shell
info: installing plugin "myplugin"

Error: Registry returned 404 for GET on https://registry.npmjs.org/gitbook-plugin-myplugin
```
This led to the fix in pluginDependency.js.

I also learned that a compound string syntax would also work in the meantime: `myplugin@git+https://github.com/YourCompany/myplugin.git`.

Unfortunately, the same error displayed (404 while attempting to load from npmjs.org), but now due to an issue in the resolveVersion method of installPlugins.js.  This methods uses the nmp `view` command, which is valid when the configured version is a semver; but not when the config version is used to support git URLs.  The semver.validRange(...) method (luckily, already imported in that file) provided the needed conditional to bypass processing non-semver *versions*.

Hope you don't mind the refactor to introduce a new test for resolveVersion.

Cheers,
Ryan

SamyPesse:
Looks good, Thanks 👍

SamyPesse:
I'll just make a few changes on master so that `installPlugin` and `resolvePlugin` use `PluginDependency` instead of `Plugin`.

ryanswanson:
Great!  Glad to help.  Appreciate the quick response and version bump so I can make use of it today.

PR #1338, opened 2016-05-28
Fix TOC links for missing files

### LINKED ISSUE #1331: TOC links are not disabled for non-existing documents
I have gitbook 3.0.2.

Taking a clean repo, I used `gitbook build` to generate the html. The link to chapters that don't exist are actual links, clicking them results in 404.

The render produced by gitbook website generates `<span>`s that are grayed out and not clickable.

Is this a regression or misconfiguration?

jpreynat:
Fix #1331 
Linked with GitbookIO/theme-default#17

⚠️  `theme-default` must be published and upgraded in `package.json` before publishing

### DIFF
--- a/lib/json/encodeSummaryArticle.js
+++ b/lib/json/encodeSummaryArticle.js
@@ -21,7 +21,8 @@ function encodeSummaryArticle(article, recursive) {
         url: article.getUrl(),
         path: article.getPath(),
         ref: article.getRef(),
-        articles: articles
+        articles: articles,
+        exists: article.getExists()
     };
 }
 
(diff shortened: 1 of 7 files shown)


SamyPesse:
I'm not a big fan of this implementation, because:

- I think the summary structure should only contain information related to the summary parsing, no information related to the existence of pages, etc
- I think it should be done only in the theme using: `{% if getPageByPath(article.path) %}` (or we can add a [`hasPageByPath`](https://github.com/GitbookIO/gitbook/blob/master/lib/output/website/createTemplateEngine.js#L84) which will only return a boolean)

In conclusion, we should avoid as much as possible sharing same information across multiple data sets: the existence is already defined in the Page list, it should not be stored in the Summary.

jpreynat:
You have a point about sharing the information across multiple objects.
Closing in favor of GitbookIO/theme-default#18

PR #1363, opened 2016-06-07
Fixed target attribute for external links

todvora:
The attribute key/value for links that should be opened into a new window is ```target="_blank"``` but Gitbook generated ```_target="blank"``` (the underscore should be before attribute value, not name)

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-target


SamyPesse:
Thanks 👍

PR #1397, opened 2016-06-17
add open and browser args to serve command

mpfeil:
@i5ting asked for an open option in https://github.com/GitbookIO/gitbook-cli/issues/37

I have added an ```open``` and ```browser``` option. ```open``` is true by default and by using ```browser``` you can specify the browser for opening the book.

Example:
```$ gitbook serve --browser firefox```

### DIFF
--- a/lib/cli/serve.js
+++ b/lib/cli/serve.js
@@ -1,6 +1,7 @@
 /* eslint-disable no-console */
 
 var tinylr = require('tiny-lr');
+var open = require('open');
 
 var Parse = require('../parse');
 var Output = require('../output');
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by i5ting** [comment]
thanks @mpfeil

File: lib/cli/serve.js [hunk 1]
```diff
@@ -112,6 +119,16 @@ module.exports = {
             description: 'Enable live reloading',
             defaults: true
         },
+        {
+            name: 'open',
+            description: 'Enable opening book in browser',
+            defaults: true
```
SamyPesse: We should disable it by default.

mpfeil:
The ```open``` argument is now disabled by default.

SamyPesse:
Thanks 👍

PR #1431, opened 2016-07-07
Fix extra dot of filePath

bipedd:
`filePath` of "README.md" in root directory should return "/" rather than "./".

### DIFF
--- a/lib/output/helper/fileToURL.js
+++ b/lib/output/helper/fileToURL.js
@@ -21,8 +21,15 @@ function fileToURL(output, filePath) {
 
     filePath = fileToOutput(output, filePath);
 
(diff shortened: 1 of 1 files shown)


SamyPesse:
@bipedd I'm not sure this is the expected result, website generated by GitBook should be able to run under a sub-directory (for example: `test.com/hello/` can be the root of a book).

If links to README are generated as `/` instead of `./` (or `../`, etc) then the link will be invalid.

bipedd:
@SamyPesse I have made a simple test, and it has been proved that you are right. It's the problem of [plugin-sitemap](https://github.com/GitbookIO/plugin-sitemap) that should have handled issue [#8](https://github.com/GitbookIO/plugin-sitemap/issues/8).

PR #1522, opened 2016-09-07
Update examples.md

jdaviescoates:
By adding the Loomio and Enspiral handbook examples


SamyPesse:
Thanks 👍 Awesome books!

PR #1549, opened 2016-10-04
Update examples.md

jdaviescoates:
adding Outlandish documents example


Soreine:
However worthwhile this book is, I don't think it serves well as a comprehensive display of GitBook's possibilities.

jdaviescoates:
Fair point :)

PR #1586, opened 2016-10-31
Update api.md

sta-szek:


SamyPesse:
Thanks @sta-szek 👍  but with GitBook v4 we are working on rewriting the API documentation.

PR #1592, opened 2016-11-07
fix template engine to work with plugin theme

Klortho:
Before this fix, a custom filter that I added to a theme would not be recognized.


SamyPesse:
Thanks @Klortho 👍  but with GitBook v4 we are changing the templating engine for themes / plugins, and this PR is no longer relevant.

PR #1619, opened 2016-12-01
Changes from svgexport cli to programatic api

mikaelbr:
(I'm not sure if there is any historic or other reason why the `svgexport` API was used through spawning a command instead of using the programatic API. I couldn't find any issues or comments on it, but it might be poor research on my part. In that case, sorry about that.)

Spawning a command instead of using dependencies and programatic API causes some unnecessary difficulties when getting started and when building on CIs or automatic build setups. And it's implicit dependencies instead of explicit ones.

In this PR I've set the `svgexport` as a dependency, but I guess it's done like this initially to reduce install time and not installing things like Phantom.js if one doesn't use the PDF-part of Gitbook? But couldn't this also be achieved by using the programatic API, peer depencencies and defering the `require`? In newer NPM versions peer deps aren't installed automatically, so you avoid having to install it if you don't need it. But it makes it easier for consumers to install it and having it as a part of their `package.json` instead of having a global installed tool.

Sorry for the long post, but what do you think?


SamyPesse:
@mikaelbr Sorry for the delay.

> In this PR I've set the svgexport as a dependency, but I guess it's done like this initially to reduce install time and not installing things like Phantom.js if one doesn't use the PDF-part of Gitbook?

Yes since phantomJS is a big dependency, the idea was to make it optional.

I'm not sure defering the `require` and using `peerDependencies` will be a better a solution than the current one. Because it'll show warnings during the installation.

> But it makes it easier for consumers to install it and having it as a part of their package.json instead of having a global installed tool.

Are you using GitBook as a node dependency ?

mikaelbr:
It's been a while, sorry for the late response. I've been thinking about this some more. And the lazy require method would probably also break if at one point you transition to static import statements. (There'll be some way to defer the loading in the module loader there also, but maybe not wanted in this case)

My main concern about this is implicit dependencies. Now you are dependent of a module which you don't declare. Also it's required as a globally installed module. I'd argue this makes it harder for usage on automatic rollout for different environments. Maybe this is a specific case for my part, but it's not always straight forward to add global tools on all CIs.

Another aspect of global installations is the potential collision between projects. Project A could be dependent on a older version of svgexport than Project B, etc.

I see your point about the peer dependency warning. It would probably have been better as an optional. Could an alternative way to solve this be using optional dependencies and local resolve of svgexport from `node_modules/.bin/svgexport`?

Or is this entire thread non-sensical and just splitting hairs? 😄

PR #1642, opened 2016-12-19
Minor grammar fix

marlonbernardes:
This change is so minor that I almost feel embarrassed... 😆 

This PR basically removes an extraneous punctuation in README.md:

image


AaronO:
@marlonbernardes Good catch, thanks !

Issue #990, opened 2015-10-24
whitelist support


azu:
Please support whitelist for files & folders.

GitBook already support [blacklist](https://github.com/GitbookIO/gitbook#ignoring-files--folders), but not support whitelist.

|  | blacklist | whitelist |
| --- | --- | --- |
| npm | [.npmignore](https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package) | ["files" field](https://docs.npmjs.com/files/package.json#files) |
| GitBook | .gitignore/.bookignore/.ignore | ???? |

I want to whitelist for GitBook.

Or 

`gitbook build` should not ignore files under the `.gitignore`
### Use case

I'v generated `*.md` files from `original-one-source-file`.
I want to control `original-one-source-file` under the Git, but not want to control `generated/*.md` files.

**.gitignore**:

```
generated/*.md
```

Unfortunately, `gitbook build` ignore `generated/*.md`.

Example: https://github.com/azu/nodefest-technical-writing

Simran-B:
You can actually do a white list like this in a `.bookignore` file:

```
*
!README.md
!SUMMARY.md
!book.json
!allFilesInHere/
```

\* to ignore everything
! to white-list files (or folders)

martindale:
Thanks for that, @Simran-B!  Very helpful!  Where in the documentation should this go?

Simran-B:
It would be great if it could be added here:
https://toolchain.gitbook.com/structure.html#ignore

martindale:
Thanks @Simran-B!  I've submitted a pull request to add it (see above).  Using this in production now, so thank you!

PR #1667, opened 2017-01-11
Add Whitelist Documentation

martindale:
From the discussion in #990, this pull request adds documentation on how to use whitelisting.  Thanks to @Simran-B for providing the information!


martindale:
It's not clear to me why tests are failing for a simple documentation change, but I'm happy to fix that if someone will give me some guidance here.

Simran-B:
I don't think it's related to this PR. AppVayor:

> Error: Command failed: npm run test 
> 'mocha' is not recognized as an internal or external command, operable program or batch file.

Looks like `mocha` is not installed?

TravisCI:

> /home/travis/build/GitbookIO/gitbook/scripts/bump.js:18
>        if (!packages.includes(key)) {
> TypeError: packages.includes is not a function

Array.prototype.includes() is part of the ES2016 standard, but the build environment does not seem to run a recent enough version of Node to support it natively.

docs/structure.md
```diff
@@ -45,6 +45,22 @@ test.md
 bin/*
 ```
 
+### Whitelisting files & folders 
+
+GitBook can also block all content by default, and only process specific "whitelisted" files and folders.
+
+Use the `*` operator first to block all content, then the `!` operator to create an exemption.
+
```
mathbunnyru: I think it would be better to say in which file to put this list.

Issue #1309, opened 2016-05-18
random errors when using gitbook plugin on running "gitbook serve".


sachetsharma:
We are running gitbook with some plugins. Plugins are loaded fine but at the end it errors out with different error messages. All of them complains about one or the other plugin file missing. Running gitbook version 2.1.3 on windows 7.

info: 7 plugins are installed
info: loading plugin "prism"... OK
info: loading plugin "edit-link"... OK
info: loading plugin "livereload"... OK
info: loading plugin "search"... OK
info: loading plugin "lunr"... OK
info: loading plugin "fontsettings"... OK
info: loading plugin "theme-default"... OK
info: found 66 pages
info: found 83 asset files

Error: ENOENT: no such file or directory, stat 'C:\Projects\myproject_book\gitbook\gitbook-plugin-fontsettings\buttons.js'

Error: ENOENT: no such file or directory, stat 'C:\Projects\myproject_book\gitbook\gitbook-plugin-lunr\lunr.min.js'

Error: ENOENT: no such file or directory, stat 'C:\Projects\myproject_book\gitbook\gitbook-plugin-livereload\plugin.js'

SamyPesse:
Can you paste me the output of `gitbook current` inside the book's directory?

SamyPesse:
And also the output of `node -v` ?

sachetsharma:
here is the output 

C:\Projects\myproject\\_book>gitbook current
GitBook version is 2.6.7

C:\Projects\myproject\\_book>node -v
v4.2.2

SamyPesse:
Run `gitbook current` inside the `C:\Projects\myproject` (I guess this is folder where your content is stored), and paste me the output.

sachetsharma:
Here it is:
C:\Projects\myproject>gitbook current
GitBook version is 3.0.0-pre.15

SamyPesse:
Ok thanks, I'll investigate and fix this.

fractalf:
I get the same..
```
alf@dev [~/gitbook] gitbook init ./foobar
warn: no summary file in this book
info: create README.md
info: create SUMMARY.md
info: initialization is finished
alf@dev [~/gitbook] cd foobar/
alf@dev [~/gitbook/foobar] node -v
v4.4.5
alf@dev [~/gitbook/foobar] gitbook current
GitBook version is 3.0.3
alf@dev [~/gitbook/foobar] gitbook build
info: 6 plugins are installed
[...]

PR #1681, opened 2017-01-18
disabling `cpr`s "confirm" option for fs.copyDir

obogobo:
Seems as if the release branch for 3.2.2 is gone, this should probably be a PR onto that and then artifact 3.2.3 would go to npm?

This fixes: https://github.com/GitbookIO/gitbook/issues/1309

### DIFF
--- a/lib/output/website/copyPluginAssets.js
+++ b/lib/output/website/copyPluginAssets.js
@@ -64,7 +64,7 @@ function copyAssets(output, plugin) {
         {
             deleteFirst: false,
             overwrite: true,
-            confirm: true
+            confirm: false // fixes Issue #1309, don't use graceful-fs.stat
         }
     );
 }


obogobo:
Right, but unfortunately line 112 is gone on master (long gone!). This fix is for v3.3.2 and while master is at 4.0.0, but I still think v3.* should be supported.

if you checkout the tag for 3.3.2 and apply the patch to line 112 you *should* see that it works?

zekedroid:
@insanum: this is a fix for v3 which is likely to stay around for a while, but either way it should be fixed on v4 as well. And it still seems to be a much deeper problem with `graceful-fs` running `stat` on systems that don't support it.

insanum:
Ah yes, ok. I'm using v3.2.2 and changing line 112 as previously mentioned fixes the issue. Sorry for the noise. :-)

zekedroid:
Bump.

ctron:
Is there any update on this one?

zekedroid:
@ctron None that we know so we resolved to publishing the @obogobo's fork to our internal registry. Hopefully this is addressed soon.

PR #1683, opened 2017-01-19
Calibre App is in /Application (no tild)

borisschapira:


borisschapira:
Duplicate of #1656, sorry.

PR #1694, opened 2017-01-29
Add custom size

dusty:
I added customSize for pdf support to the 3.2.2 tag.

### DIFF
--- a/lib/output/ebook/getConvertOptions.js
+++ b/lib/output/ebook/getConvertOptions.js
@@ -50,7 +50,7 @@ function getConvertOptions(output) {
         .spread(function(headerTpl, footerTpl) {
             var pdfOptions = config.getValue('pdf').toJS();
 
-            return options = extend(options, {
+            options = extend(options, {
                 '--chapter-mark':           String(pdfOptions.chapterMark),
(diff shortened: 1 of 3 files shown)


lib/output/ebook/getConvertOptions.js
```diff
@@ -65,6 +65,10 @@ function getConvertOptions(output) {
                 '--pdf-header-template':    headerTpl,
                 '--pdf-footer-template':    footerTpl
             });
+            if (pdfOptions.customSize) {
+                options['--custom-size'] = String(pdfOptions.customSize);
```
annegentle: Then what happens? How does the generator know what to do next?

dusty: All this does is allows --custom-size to pass through to calibre.  You can then add what size you want to your book.json file.

eg:

```
{
  "pdf": {
    "customSize": "8.12x10.25"
  }
}
```

Then when you run 'gitbook pdf' it will create the size you want.

If you are on OSX, you can just patch your currently installed version.  Replace the version number with whatever you are running.

```
vi ~/.gitbook/versions/3.2.2/lib/output/ebook/getConvertOptions.js
```

annegentle: Yep, this approach works, thanks!

PR #1695, opened 2017-01-30
remove this regex constraint since config.structure.readme can have a path

grinich:
If you set config.root.


grinich:
Perhaps there's a better way to do this, but this constraint was breaking my gitbook builds

AaronO:
@SamyPesse Want to take a look at this ?

Issue #1703, opened 2017-02-07
GLOSSARY don't work with Chinese characters


yuchuanxi:
Hi, I'm trying the GLOSSARY.md . It works well. but it no work with chinese.

glossary.md:
```md
## Term
Definition for this term

## Another term
With it's definition, this can contain bold text
and all other kinds of inline markup ...

## 小土豆
小土豆是个很土很逗的逗比
```

article md:
```md
# Article 1

这是一个小土豆的故事

这是一个*小土豆*的故事

这是另一个Term的故事

这是另一个**Term**的故事

这是另一个~~Term~~的故事

这是另一个*Term*的故事

这是另一个`Term`的故事
```

thank you very much

kogakure:
I can confirm, all non-ascii characters get ignored, e.g. the German word **Ära** or **Fuß**. In my case I have a term of **Fu** in my glossary and my word will be marked with boundaries inside the word, like this: _Fu_ß.

indoorsNB:
same problem :(

jfoboss:
same problem with Russian Language :(

EternallLight:
For Russian Language, it is still not working in Gitbook 3.2.3.

juhuan:
hope to resolve

juhuan:
@yuchuanxi 我连英文都不能用诶，兄弟

yuchuanxi:
@juhuan 我好久没用这个了。。可以尝试着自己fork debug改下代码

kntmr:
same problem in the Japanese language in GitBook 3.2.3.

PR #1704, opened 2017-02-07
fix bug: gloassy can not distinguish the chinese character word

yuchuanxi:
#1703 fix the bug


yuchuanxi:
the code base `tag 3.2.2`

edwardtoday:
This fix works, thank you.

PR #1733, opened 2017-03-05
Resolves #1015, enable the include directive (using server safe mode)

### LINKED ISSUE #1015: Asciidoc's `include` macro is parsed as link
I have a book which is spread over many files. I need to be able to build with normal `asciidoc` or `asciidoctor` into a single file, too. I hence use statements like `include::was-ist-debian.adoc[]`.

Gitbook shows this statement as `<p><a href="was-ist-debian.html" class="bare">was-ist-debian.adoc</a></p>` though. I would have expected that it includes the given file at this place instead of placing a link or completely ignored at all (in case it needs to be listed in SUMMARY.adoc separately).

I even tried adding a title parameter (locally) to make it look nicer, but no success: `include::was-ist-debian.adoc[Was ist Debian?]`

(The according code can be found in the gitbook feature branch of dpmb/dpmb at https://github.com/dpmb/dpmb/blob/gitbook/konzepte/linux-dschungel/linux-dschungel.adoc, the book on Gitbook is currently private due to its unfinished Gitbook support at https://dpmb.gitbooks.io/debian-paketmanagement/content/v/gitbook/konzepte/linux-dschungel/linux-dschungel.html)

Mogztter:
Read more about include directive: http://asciidoctor.org/docs/user-manual/#include-directive

NOTE: The path is resolved relatively to where the command where run. We could define the `base_dir` on each page to resolve relatively to the page.

In other words, given:
```
chapter-1/
  index.adoc
  include.adoc
```
The following **will** work:

```
include::chapter-1/include.adoc[]
```

The following **will not** work:

```
include::include.adoc[]
```

Resolves #1015

### DIFF
--- a/packages/gitbook/src/parsers/__tests__/asciidoc.js
+++ b/packages/gitbook/src/parsers/__tests__/asciidoc.js
@@ -1,0 +1,9 @@
+const AsciidocParser = require('../asciidoc');
+
+describe('AsciidocParser', () => {
+
+    it('should convert to HTML and include file', () => {
+        const result = AsciidocParser.toHTML('= GitBook User Manual\n\n== Usage\n\ninclude::src/parsers/__tests__/usage.adoc[]');
+        expect(result).toMatch(/To use GitBook/);
+    });
+});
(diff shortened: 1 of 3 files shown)


Mogztter:
Closing, GitBook dropped support for AsciiDoc: https://docs.gitbook.com/v2-changes/important-differences#asciidoc-support

PR #1734, opened 2017-03-05
Use font icons by default in AsciiDoc

Mogztter:


packages/gitbook/src/parsers/asciidoc.js
```diff
@@ -36,7 +36,7 @@ function toDocument(text) {
  */
 function toHTML(text) {
     return asciidocjs.convert(text, {
-        attributes: 'showtitle'
+        attributes: ['showtitle', 'icons=font@']
```
Mogztter: `@` symbol means that the value can be overridden

Mogztter:
Closing, GitBook dropped support for AsciiDoc: https://docs.gitbook.com/v2-changes/important-differences#asciidoc-support

PR #1735, opened 2017-03-05
Fix Asciidoctor.js version in package.json

Mogztter:
Asciidoctor.js does not use Semantic Versioning as the release versions are aligned on Asciidoctor (Ruby).
It’s highly recommended to define the exact version in your package.json file (ie. without ^).


Mogztter:
Closing, GitBook dropped support for AsciiDoc.

https://docs.gitbook.com/v2-changes/important-differences#asciidoc-support

PR #1741, opened 2017-03-08
Ignore empty articles, and be more flexible in parsing articles

Soreine:
Once integrated in the Editor, this would 

- Fix https://github.com/GitbookIO/feedback/issues/372 and fix https://github.com/GitbookIO/feedback/issues/411

  The reason is links in the parsed Slate.Document are surrounded by empty texts in Asciidoc, so considered as empty articles. And empty articles were not ignored.

- Fix https://github.com/GitbookIO/editor/issues/571

### DIFF
--- a/packages/gitbook/package.json
+++ b/packages/gitbook/package.json
@@ -43,7 +43,7 @@
     "json-schema-defaults": "0.1.1",
     "jsonschema": "1.1.0",
     "juice": "2.0.0",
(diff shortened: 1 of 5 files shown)


manngo:
Whatever this issue fixes it does not appear to resolve  GitbookIO/feedback#372 GitbookIO/feedback#411.

Both issues are causing problems with asciidoc. Version 7.0.11.

PR #1744, opened 2017-03-10
Summary files operations

Soreine:
Adds two summary modifiers, required for https://github.com/GitbookIO/editor/pull/573

- `deleteByPath` which updates a summary in response to a path being deleted.
- `movePath` which updates a summary in response to a path being moved.


Soreine:
I'll try to link `gitbook` locally and work on the Editor. Just in case I realize I need to add something else.

PR #1764, opened 2017-03-28
Simpler summary export

Soreine:
This always default to a link export for articles.
This helps having a more reliable markdown syntax for the summary, and allows writing numbered articles:

```
# Summary

* [Introduction](README.md)
* [Empty]()
* [1. Numbered]()
```

Which is otherwised parsed as a numbered sublist

Related to this editor bug: https://gitbook.zendesk.com/agent/tickets/5478


Soreine:
Another fix will be made in the editor. The editor will always create articles with a default filename as ref.

PR #1722, opened 2017-02-19
Catch Error: ENOENT: no such file or directory

shigemk2:
I add error handling for symbolic link file that links impalpable file.

https://github.com/GitbookIO/gitbook/issues/1500

gitbook serve would fail because fs.statSync in Node.js read Emacs lock file. It is because .#filename.md is symbolic link file and it is impalpable file.

This is debug log when gitbook serve fails. 

```
Error: ENOENT: no such file or directory, stat '/home/me/dir/.#filename.md'
    at Error (native)
    at Object.fs.statSync (fs.js:987:18)
    at /home/me/.gitbook/versions/3.2.2/lib/fs/node.js:17:31
    at /home/me/.gitbook/versions/3.2.2/node_modules/immutable/dist/immutable.js:3018:46
    at List.__iterate (/home/me/.gitbook/versions/3.2.2/node_modules/immutable/dist/immutable.js:2208:13)
    at IndexedIterable.mappedSequence.__iterateUncached (/home/me/.gitbook/versions/3.2.2/node_modules/immutable/dist/immutable.js:3017:23)
    at seqIterate (/home/me/.gitbook/versions/3.2.2/node_modules/immutable/dist/immutable.js:606:16)
    at IndexedIterable.IndexedSeq.__iterate (/home/me/.gitbook/versions/3.2.2/node_modules/immutable/dist/immutable.js:322:14)
    at IndexedIterable.forEach (/home/me/.gitbook/versions/3.2.2/node_modules/immutable/dist/immutable.js:4383:19)
    at /home/me/.gitbook/versions/3.2.2/node_modules/immutable/dist/immutable.js:2071:14
```

Issue #1500, opened 2016-08-23
Gitbook serve crashes if temporary files created by emacs disappear


austinjp:
### What is the current behavior?

Occasionally my locally-hosted `gitbook serve` will fail with an error similar to:

```
Error: ENOENT: no such file or directory, stat '/home/me/dir/.#filename.md'
```

Files beginning with `.#` are left lying around by emacs. Removing those files manually fixes the issue, if they in fact exist. Occasionally it seems that these files have temporarily existed, then have been automatically cleaned by emacs, leading `gitbook server` to get confused about what files exist. It then deliberately shuts down.

Output from `gitbook version`:

```
CLI version: 2.3.0
GitBook version: 3.2.0
```

Running on Ubuntu 14.04.1.
### What is the expected behavior?

Perhaps generate a warning and ignore the error, carrying on if possible?

austinjp:
Hmm. Manually creating a file of that format with e.g. `touch '.#whatever.md'` fails to recreate this issue.

When it happens again I'll investigate and add more detail.

It's extremely low priority give how easy it is to work around :smile:

shigemk2:
Me neither.

I think there are two solutions.

First, put `(setq create-lockfiles nil)` in our init.el. Then, we can avoid Gitbook serve crash.
Second, fix gitbook's code and send PR. But I am not sure how to debug/build gitbook locally. https://github.com/GitbookIO/gitbook/issues/1435

shigemk2:
This is complex problem.

First, current gitbook would not ignore files whose names start with ".#" just like ".#filename.md" even if we write ".\#*" on .gitignore/.bookignore/.ignore because watcher might watch all markdown files (in watch.js).

Second, current gitbook would fail when gitbook read Emacs lock file because Emacs lock file(like ".#readme.md") is symbolic link file and it doesn't have entity (in node.js).

I sent Pull Request #1722 that fixes second problem.

austinjp:
👍 Nice. I haven't used gitbook for a few months but may need to do so again soon.

PR #1889, opened 2017-10-20
Ignore symlink

shigemk2:
Refer #1500 and #1722 

gitbook crashes when gitbook serves symbolic files (e.g. `Error: ENOENT: no such file or directory, stat '/path/to/symlink_file'`.

I will ignore symlink files.


cdegroot:
Bump. I'm having the same issue ;-)

PR #1992, opened 2018-03-03
improve support for linking SUMMARY.md to sources outside root dir

warren-bank:
* this does NOT fix the fact that HTML files are saved outside of the $TEMP directory
  * root directory is mapped to: `$TEMP/tmp-12345`
  * relative paths can escape this directory:
    * `../foo.md` => `$TEMP/foo.html`
    * `../../bar.md` => `$TEMP/../bar.html`
    * `../../../baz.md` => `$TEMP/../../baz.html`
    * `../../foo/bar/baz.md` => `$TEMP/../foo/bar/baz.html`
      * directory structure is created
      * `foo` directory becomes a sibling to `$TEMP`

* this DOES fix the relative paths linking these HTML files to CSS stylesheets that ARE stored inside of $TEMP

### DIFF
--- a/lib/cli/buildEbook.js
+++ b/lib/cli/buildEbook.js
@@ -25,6 +25,7 @@ module.exports = function(format) {
 
             // Create temporary directory
             var outputFolder = tmp.dirSync().name;
+            options.outputFolder = outputFolder;
 
             var book = getBook(args, kwargs);
             var logger = book.getLogger();
@@ -71,6 +72,7 @@ module.exports = function(format) {
                 logger.info.ok(count + ' file(s) generated');
 
                 logger.debug('cleaning up... ');
+                delete options.outputFolder;
                 return logger.debug.promise(fs.rmDir(outputFolder));
             });
         }
--- a/lib/output/website/createTemplateEngine.js
+++ b/lib/output/website/createTemplateEngine.js
@@ -91,6 +91,23 @@ function createTemplateEngine(output, currentFile) {
         return JSONUtils.encodePage(page, summary);
     }
 
+    function resolveFileNotWithinRoot(filePath) {
+        if (currentFile.substr(0,2) === '..') {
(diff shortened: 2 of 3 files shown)


warren-bank:
Hi. This PR is the direct result of the discussion in this other [PR](https://github.com/serverless/guide/pull/80).

Long-story-short:
* I wanted to add the ability to generate PDSf to this documentation repo
* their structure supports it nicely
* the root of the repo has a `README.md` and it links to other pages inside a `./sources/*.md` subdirectory
* they wanted all files pertaining to ebooks in a separate subdirectory: `./ebook/`
* I wrote a bunch of scripts.. but ultimately the gitbook root folder is at: `./ebook/src`
* the ebook documents that are generated by gitbook are saved to: `./ebook/dist`
* `./ebook/src/SUMMARY.md` contains links to: `../../sources/*.md`
* as is, ebooks are generated and saved to the dist directory.. exactly as desired
* the problem is.. the ebooks are unstyled
* after some digging, I see that:
  * the markdown files are converted to HTML and saved into a directory: `$TEMP/tmp-12345`
  * the problem is.. markdown files that are identified by a relative link (ie: `../../`).. the generated HTML is stored in the parent of $TEMP
    * this is crazy dangerous!
    * it also breaks links to CSS stylesheets

warren-bank:
oh.. looks like the javascript needs to be ES4 or ES5

sorry.. I just assumed it would be transpiled from ES6

warren-bank:
the notes I took while tracing the code and testing the effects of changes are [here](https://github.com/warren-bank/serverless-guide/raw/5bfe2dcfd588fbd6773736b92ebab9709e7c75c3/ebook/.etc/2.%20notes%20-%20gitbook%20-%20tracing%20issues%2C%20tests%2C%20and%20PR.txt).. but I'll warn you that they're excessively long

PR #2022, opened 2018-04-16
Fix typo breaking monospacing

Lupie:
`gitbook --> `gitbook`


Lupie:
@atilacamurca I don't have write access, so I can't merge!

PR #2036, opened 2018-05-06
Update validatePlugin messages and tests

fdesjardins:
These commits add detailed error messages for plugin validation and expand on the plugin validation tests. This change was motivated by the infamous `Error: Error loading plugin ...` error message generated in `validatePlugin()`.

While developing a new theme, we forgot to include `gitbook` in the `engines` field in the plugin's `package.json` file. Unfortunately, it took a little while to track down the issue because of the previous generic error message, even with the use of `--debug --log=debug`.

This should hopefully help others avoid making the same mistake .

### DIFF
--- a/lib/plugins/__tests__/validatePlugin.js
+++ b/lib/plugins/__tests__/validatePlugin.js
@@ -1,16 +1,80 @@
+var Immutable = require('immutable');
 var Promise = require('../../utils/promise');
 var Plugin = require('../../models/plugin');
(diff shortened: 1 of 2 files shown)


fdesjardins:
@SamyPesse Sorry to have to tag you directly. What's the process of getting fixes merged into Gitbook at this stage? I see some PRs merged recently while others have remained open for quite some time. Has all legacy support ended, including accepting PRs? Thanks in advance for any more info

PR #2222, opened 2022-07-06
Feature Request: Random Page

oddrabbit:
Hello there,

Over at the carlospolop / hacktricks project (https://github.com/carlospolop/hacktricks), we would like to add in a feature to Gitbook so that users can retrieve a random page. We don't know how to do this in our community. Would this be possible to add into Gitbook?

Here is the issue: https://github.com/carlospolop/hacktricks/issues/339

Thanks.

### DIFF
--- a/book.js
+++ b/book.js
@@ -21,3 +21,4 @@ module.exports = {
         }
     }
 };
+


oddrabbit:
Hello there,

We have resolved this Issue on our end so no need for this feature to be added then.

Thanks.

PR #2228, opened 2024-03-24
Fix zoom on small images with custom zoom component

SamyPesse:
- [x] Adapt to all images
- [x] Prevent zoom on images that will not be expanded
- [x] Prevent zoom on mobile since user can zoom with fingers


src/components/utils/Image.tsx
```diff
@@ -271,16 +271,14 @@ async function ImagePicture(
...
-        <img
-            alt={alt}
-            style={style}
-            loading={loading}
-            fetchPriority={fetchPriority}
-            {...rest}
-            {...attrs}
-        />
-    );
+    const imgProps: React.ComponentPropsWithoutRef<'img'> & { src: string } = {
```
gregberge, line 274:
```suggestion
    // https://www.totaltypescript.com/react-apps-ts-performance
    interface ImgPropsWithSrc extends React.ComponentPropsWithoutRef<'img'> {
      src: string
    }
    const imgProps: ImgPropsWithSrc = {
```

src/components/utils/ZoomImage.module.css
```diff
@@ -0,0 +1,19 @@
+html:has(.zoomModal) {
```
gregberge, line 1: I thought we were using Tailwind CSS?

src/components/utils/ZoomImage.tsx
```diff
@@ -0,0 +1,259 @@
...
+    const [zoomable, setZoomable] = React.useState(false);
+    const [active, setActive] = React.useState(false);
+    const [opened, setOpened] = React.useState(false);
+
+    // Only allow zooming when image will not actually be larger and on mobile
+    React.useEffect(() => {
+        if (isTouchDevice()) {
+            setZoomable(false);
+            return;
+        }
```
gregberge, line 33:
```suggestion
        if (isTouchDevice()) {
            return;
        }
```

src/components/utils/ZoomImage.tsx
```diff
@@ -0,0 +1,259 @@
...
+        if (!resizeObserver) {
+            // When resizeObserver is available, it'll take care of calling the changelog as soon as the element is observed
+            onChange();
+        }
+
+        return () => {
+            resizeObserver?.disconnect();
+            mediaQueryList.removeEventListener('change', onChange);
+        };
+    }, [imgRef, width, setZoomable]);
```
gregberge, line 73: There is two responsibilities in this, I think it should be splitted in several useEffect, but anyway I think it's fine.

src/components/utils/ZoomImage.tsx
```diff
@@ -0,0 +1,259 @@
...
+    const onClose = React.useCallback(() => {
+        startViewTransition(
+            () => {
+                setOpened(false);
+            },
+            () => {
+                setActive(false);
+            },
+        );
+    }, [setOpened]);
```
gregberge, line 103: no need

src/components/utils/ZoomImage.tsx
```diff
@@ -0,0 +1,259 @@
...
+                onClose();
+            }
+        };
+
+        document.addEventListener('keydown', handleKeyDown);
+
+        return () => {
+            document.removeEventListener('keydown', handleKeyDown);
+        };
+    }, [onClose]);
```
gregberge, line 165: It's stable because contained in the same component, yep but still it's hard to know. You should probably wrap `ZoomImageModal` in a `memo` then we can guess that it awaits stable props.

src/components/utils/ZoomImage.tsx
```diff
@@ -0,0 +1,259 @@
...
+                <IconMinimize />
+            </button>
+        </div>
+    );
+}
+
+function startViewTransition(callback: () => void, onEnd?: () => void) {
+    // @ts-ignore
+    if (document.startViewTransition) {
+        // @ts-ignore
```
gregberge, line 237: We should get rid of `ts-ignore`

gregberge (changes requested):

src/components/utils/ZoomImage.tsx
SamyPesse: This API is still experimental, it's fine

src/components/utils/ZoomImage.tsx
SamyPesse: It's fine, nothing says that useEffect should have only one responsibility, especially when controlling a single state based on multiple conditions.

src/components/utils/ZoomImage.module.css
SamyPesse: We are, but it doesn't mean that we don't do CSS, we use both.

For example for the code-blocks, I want to use more CSS to avoid repetitive long class names that create a huge HTML.

In this case, Tailwind doesn't have a way to do `has()` without going through the config and I don't want a custom selector just for this small CSS.

gregberge: My advice is to only put CSS in one single file at the root because you could end up with some issues with style loading.

Also we could do most of it using Tailwind and for what is not supported do it in the CSS file at the root.

src/components/utils/ZoomImage.tsx
```diff
@@ -104,11 +112,25 @@ export function ZoomImage(
...
+                <>
+                    {placeholderRect ? (
+                        // Placeholder to keep the layout stable when the image is removed from the DOM
+                        <span
+                            style={{
+                                display: 'block',
+                                width: placeholderRect.width,
+                                height: placeholderRect.height,
+                            }}
+                        />
```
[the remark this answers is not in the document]
SamyPesse: We also use it for inline images that could be wrapped in `<a>` where a `<div>` will throw an error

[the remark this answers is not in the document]
gregberge: You are so HTML 4.01 https://stackoverflow.com/a/1828032

PR #2230, opened 2024-03-25
Fix `fullPage` screenshots

gregberge:


SamyPesse (commented):
I've tested this last week and I finally think it's better to default to `false` and set it to `true` on case by case basis.

gregberge:
> I've tested this last week and I finally think it's better to default to `false` and set it to `true` on case by case basis.

OK, I changed it then.

SamyPesse (approved):

PR #2234, opened 2024-03-25
Support images & files in synced blocks

gregberge:


src/components/DocumentView/BlockSyncedBlock.tsx
```diff
@@ -13,22 +14,34 @@ export async function BlockSyncedBlock(props: BlockProps<DocumentBlockSyncedBloc
...
-            document={result.document}
+            nodes={syncedBlock.document.nodes}
+            document={syncedBlock.document}
             ancestorBlocks={[...ancestorBlocks, block]}
-            context={context}
+            context={{
+                ...context,
+                resolveContentRef: async (ref, options) => {
+                    if (!syncedBlock?.files) {
+                        return context.resolveContentRef(ref, options);
```
SamyPesse, line 36: This is incorrect and will not match the behaviour in the editor.

If I copy a link `{ kind: "page", page: 'something' }` in a sync block, it'll do a broken link in the app, but will work in published content if the page `something` exists in the space.

Resolution of content ref in the synced block should be isolated from the one for the rest of the content (clean context). Except if you tell me the logic has changed in the app.

SamyPesse (approved):
Looks good!

src/components/DocumentView/BlockSyncedBlock.tsx
```diff
@@ -13,22 +14,34 @@ export async function BlockSyncedBlock(props: BlockProps<DocumentBlockSyncedBloc
...
-            nodes={result.document.nodes}
-            document={result.document}
+            nodes={syncedBlock.document.nodes}
+            document={syncedBlock.document}
             ancestorBlocks={[...ancestorBlocks, block]}
-            context={context}
+            context={{
+                ...context,
+                resolveContentRef: async (ref, options) => {
+                    if (!syncedBlock?.files) {
```
emmerich, line 35:
```suggestion
                    if (!syncedBlock?.files || ref.kind !== 'file') {
```

I'd probably check whether the content ref is a file here and then I'd have more specific typing below, but it's not a big deal.

src/components/DocumentView/BlockSyncedBlock.tsx
```diff
@@ -13,22 +14,34 @@ export async function BlockSyncedBlock(props: BlockProps<DocumentBlockSyncedBloc
...
         <Blocks
-            nodes={result.document.nodes}
-            document={result.document}
+            nodes={syncedBlock.document.nodes}
+            document={syncedBlock.document}
             ancestorBlocks={[...ancestorBlocks, block]}
-            context={context}
+            context={{
+                ...context,
+                resolveContentRef: async (ref, options) => {
```
emmerich, line 34: i'm actually curious how other content refs are going to be resolved. Like if you do an `#anchor` ref to a heading in the page, it'll probably come out as `SpaceName#anchor` because the Synced Block and the Space are different.

emmerich (approved):

src/components/DocumentView/BlockSyncedBlock.tsx
gregberge: This is the way it should work. If a customer use a link in a synced block and use its synced block in only one space it's better.

src/components/DocumentView/BlockSyncedBlock.tsx
gregberge: It's handled at low level, no need of that.

src/components/DocumentView/BlockSyncedBlock.tsx
gregberge: We will resolve in the space, but anyway we didn't decide yet about what we are doing about relative refs. See discussion with @SamyPesse in other thread.

PR #2257, opened 2024-04-01
Remove mask on page cover

SamyPesse:
This PR removes the mask on page cover that is impacting content such as https://docs.tuum.com/api-cookbooks.
The mask is a difference with the GitBook editor that doesn't seem justified.

### DIFF
--- a/src/components/PageBody/PageCover.tsx
+++ b/src/components/PageBody/PageCover.tsx
@@ -69,14 +69,7 @@ export async function PageCover(props: {
                     // Maximum size of the cover
                     { width: 1248 },
                 ]}
-                className={tcls(
-                    'w-full',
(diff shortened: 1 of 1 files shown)


src/components/PageBody/PageCover.tsx
```diff
@@ -69,14 +69,7 @@ export async function PageCover(props: {
                     // Maximum size of the cover
                     { width: 1248 },
                 ]}
-                className={tcls(
-                    'w-full',
-                    'object-cover',
-                    'object-center',
-                    as === 'full'
```
SamyPesse, line 76: @sebastiangraz I can't remember or trace back a reason for this mask. I'm going to merge this PR but let me know async if you have any concern.

PR #2267, opened 2024-04-04
Fix an issue where OpenAPI blocks could not be fetched.

emmerich:
Whitespace in a URL is handled properly in the browser/Node, but not in Cloudflare. We've raised it on their side but in the meantime we'll wrap in a URL to make sure we parse correctly.


SamyPesse (approved):

SamyPesse:
The issue is related to https://github.com/cloudflare/workerd/issues/1957

PR #2268, opened 2024-04-04
small typo

scazan:


src/lib/cache/cache.ts
```diff
@@ -126,6 +126,7 @@ export function cache<Args extends any[], Result>(
             // Try the memory backend, independently of the other backends as it doesn't have a network cost
             const memoryEntry = await memoryCache.get(key);
             if (memoryEntry) {
+                console.log(`${key} memcache ${Date.now()} setAt=${memoryEntry.meta.setAt} expiresAt=${memoryEntry.meta.expiresAt} revalidatesAt=${memoryEntry.meta.revalidatesAt}`)
```
SamyPesse, line 129: What is this log? do we need it? Cloudflare has some limits on logs

scazan: Sorry, this was from debugging with Steven yesterday. We kinda messed this PR with logs in order to deploy logging to a Cloudflare deployment (since we can't see it locally). I'll create a new one for the typo.

PR #2283, opened 2024-04-13
Avoid crashing Cloudflare Worker when syntax highlighting too many lines of code.

emmerich:
For pages with 100s of lines of code, the Cloudflare worker crashes.  After a lot of investigation, it seems that Shiki uses more memory as the rendering goes on.

Code blocks are rendered in parallel thanks to Next, so it's possible to have multiple Shiki tokenizations happening at once. But even after having code blocks render sequentially (no more than one tokenization operation at a time), a page with 100+ blocks will still crash from memory.

I've shipped a hotfix that stops code highlighting after 50 blocks just so that the pages don't crash anymore. Otherwise longer term fixes would be:
- A more powerful Worker :pray:
- Doing syntax highlighting client-side (no support yet in Shiki as the onigasm library is not supported in browsers yet)
- More investigation into the memory leak, it's possible that only certain lines of code cause a memory explosion

Future work tracked here: https://linear.app/gitbook-x/issue/RND-3588/gitbook-open-code-syntax-highlighting-runs-out-of-memory-after-a

### DIFF
--- a/src/components/DocumentView/CodeBlock/highlight.ts
+++ b/src/components/DocumentView/CodeBlock/highlight.ts
@@ -36,7 +36,10 @@ type PositionedToken = ThemedToken & { start: number; end: number };
  * This is done per invocation of the Cloudflare worker, so we can store it in-memory.
  */
 let lineCount = 0;
-const LINE_LIMIT = 750;
+let tokenCount = 0;
+const LINE_LIMIT = 10000;
+
+const runner = asyncMutexFunction();
 
 /**
  * Highlight a code block while preserving inline elements.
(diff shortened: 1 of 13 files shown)


src/components/DocumentView/CodeBlock/highlight.ts
```diff
@@ -27,16 +29,31 @@ type InlineIndexed = { inline: any; start: number; end: number };
 
 type PositionedToken = ThemedToken & { start: number; end: number };
 
+/**
+ * Due to a combination of memory limitations of Cloudflare workers and the memory
+ * cost of shiki, we need to set a limit on the number of blocks we can highlight
+ * in a single page.
+ */
+let blockCount = 0;
+const BLOCK_LIMIT = 50;
```
emmerich, line 38: it's not ready to merge yet, storing this in a global variable means it's shared across executions. So one request for a page with a lot of blocks could impact the block loading of another, unrelated page

valentin0h (approved):

PR #2288, opened 2024-04-17
Support GitHub-flavoured Markdown when parsing OpenAPI descriptions.

emmerich:
Some GFM syntax is not supported in our OpenAPI blocks, but it's relatively easy to support using Unified plugins.


scazan (approved):

src/components/Header/Header.tsx
```diff
@@ -33,7 +33,9 @@ export function Header(props: {
 
     const isCustomizationDefault =
         customization.header.preset === CustomizationHeaderPreset.Default;
-    const isMultiVariants = parent?.object === 'collection' || (parent && parent.object === 'site' && spaces.length > 1)
+    const isMultiVariants =
```
spastorelli, line 36: Did you get a format error for this? Can't see any error on the commit before that change:
https://github.com/GitbookIO/gitbook/actions/runs/8722329404/job/23927959633

What settings do you use locally for format? We should probably add a config for formatting.

PR #2293, opened 2024-04-18
Pass a nonce to the script for cdn.iframe.ly/embed.js

SamyPesse:
To fix:

Screenshot 2024-04-18 at 08 25 25

It looks like host based allowlist is disabled.


SamyPesse:
It's not working well. Next tries to preload the script but doesn't pass the `nonce` to the `<link>` tag.

PR #2298, opened 2024-04-19
Investigate an issue where images with 127.0.0.1 in the URL are not correctly resolved.

emmerich:
Due to a bug in somewhere in the Cloudflare/Nextjs/next-on-pages stack, `127.0.0.1` is converted to `localhost` in URLs. We haven't been able to find the root cause, this PR fixes it temporarily.

### DIFF
--- a/src/lib/urls.ts
+++ b/src/lib/urls.ts
@@ -65,3 +65,21 @@ export function getPDFUrlSearchParams(
 
     return searchParams;
 }
+
+/**
+ * Because of a bug in Cloudflare, 127.0.0.1 is replaced by localhost.
+ * We protect against it by converting to a special token, and then parsing
+ * the token in the image API.
+ */
+const GITBOOK_LOCALHOST_TOKEN = '$GITBOOK_LOCALHOST$';
+
+/**
+ * Prepare a URL for the GitBook Open Image API.
+ */
+export function getImageAPIUrl(url: string): string {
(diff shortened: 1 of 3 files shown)


src/lib/urls.ts
```diff
@@ -65,3 +65,21 @@ export function getPDFUrlSearchParams(
...
+/**
+ * Prepare a URL for the GitBook Open Image API.
+ */
+export function getImageAPIUrl(url: string): string {
+    return url.replaceAll('127.0.0.1', GITBOOK_LOCALHOST_TOKEN);
+}
+
+export function parseImageAPIURL(url: string): string {
+    return url.replaceAll(GITBOOK_LOCALHOST_TOKEN, '127.0.0.1');
+}
```
clairechabas, line 85: nit: personal preference for camel case applied to acronyms as well

```suggestion
export function getImageApiUrl(url: string): string {
    return url.replaceAll('127.0.0.1', GITBOOK_LOCALHOST_TOKEN);
}

export function parseImageApiUrl(url: string): string {
    return url.replaceAll(GITBOOK_LOCALHOST_TOKEN, '127.0.0.1');
}
```

clairechabas (approved):
Lgtm but you might want the eyes of someone more versed with this scope

src/lib/urls.ts
emmerich: it's `API` everywhere else in GBO so I'll keep it that way.

PR #2303, opened 2024-04-26
feat: update scalar deps

marclave:
In this PR we:

- updated the scalar dependencies (api-client)
- this included our new theming changes - while they are backwards compatible, we updated to the new scalar-variables. This was done to avoid collisions with users already using --theme-variables
- fixed the body auto populating fields bug


emmerich (approved):
The changes look good, our repo doesn't look correctly configured to run workflows on PRs from forks but hopefully this approval will fix that.

emmerich:
Argh, looks like the repo is not correctly configured to handle `pull_request` events (which would allow forks to run actions). It's running on `push`. I'd change it but it'd also affect how we release the code.

In the meantime I've re-opened your PR in the repo here: https://github.com/GitbookIO/gitbook/pull/2304

PR #2309, opened 2024-05-06
Revert "Fix VA cookie to be only set if basePaths match (#2300)"

gregberge:
This reverts commit 9a86965ba139329e2061e0396f1bd2ef34e200c1.


rodrcastro (approved):

gregberge:
Not the cause of problems.

PR #2312, opened 2024-05-14
Fix an issue where attackers could use the Nextjs image endpoint to redirect to any URL.

emmerich:
Due to a vulnerability in Nextjs, attackers could use the internal Next image API to present any content under the URL of a published GitBook space.

This should prevent it, by only allowing GitBook URLs through the image endpoint.


SamyPesse (approved):
Looks good as we use our own `~gitbook/image` anyway.

Maybe we can even disable next/image entirelly 🤔

PR #2313, opened 2024-05-14
Add support for examples in the OpenAPI block

vibhanshub:
Currently on Prod
Screen Shot 2024-05-14 at 5 11 56 PM

With my changes
Screen Shot 2024-05-14 at 5 11 38 PM

Spec file in this ticket https://linear.app/gitbook-x/issue/RND-3554/show-examples-in-published-api-block-for-parameters

PR #2314, opened 2024-05-14
Add support for examples in OpenAPI block

vibhanshub:
Same description as https://github.com/GitbookIO/gitbook/pull/2313

### DIFF
--- a/packages/react-openapi/src/OpenAPISchema.tsx
+++ b/packages/react-openapi/src/OpenAPISchema.tsx
@@ -96,7 +96,7 @@ export function OpenAPISchemaProperty(
                     {schema.example ? (
                         <Markdown
                             source={'Example: ' + schema.example}
-                            className="openapi-schema-description"
+                            className="openapi-schema-example"
                         />
                     ) : null}
                 </div>
--- a/packages/react-openapi/src/OpenAPISchema.tsx
+++ b/packages/react-openapi/src/OpenAPISchema.tsx
@@ -94,10 +94,7 @@ export function OpenAPISchemaProperty(
                         />
                     ) : null}
                     {schema.example ? (
-                        <Markdown
-                            source={'Example: ' + schema.example}
-                            className="openapi-schema-example"
-                        />
+                        <span className="openapi-schema-example">Example: {schema.example} </span>
                     ) : null}
                 </div>
             }
--- a/packages/react-openapi/src/OpenAPISchema.tsx
+++ b/packages/react-openapi/src/OpenAPISchema.tsx
@@ -96,7 +96,7 @@ export function OpenAPISchemaProperty(
                         />
                     ) : null}
                     {shouldDisplayExample(schema) ? (
-                        <span className="openapi-schema-example">Example: <code>{schema.example}</code></span>
+                        <span className="openapi-schema-example">Example: <code>{JSON.stringify(schema.example)}</code></span>
                     ) : null}
                 </div>
             }
--- a/packages/react-openapi/src/OpenAPISchema.tsx
+++ b/packages/react-openapi/src/OpenAPISchema.tsx
@@ -14,6 +14,7 @@ interface OpenAPISchemaPropertyEntry {
     propertyName?: string;
     required?: boolean;
     schema: OpenAPIV3.SchemaObject;
+    example?: string;
 }
 
 /**
@@ -89,6 +90,12 @@ export function OpenAPISchemaProperty(
                     {schema.description ? (
                         <Markdown
                             source={schema.description}
+                            className="openapi-schema-description"
+                        />
+                    ) : null}
+                    {schema.example ? (
+                        <Markdown
+                            source={'Example: ' + schema.example}
                             className="openapi-schema-description"
                         />
                     ) : null}
--- a/packages/react-openapi/src/OpenAPISchema.tsx
+++ b/packages/react-openapi/src/OpenAPISchema.tsx
@@ -48,7 +48,6 @@ export function OpenAPISchemaProperty(
         : getSchemaAlternatives(schema, new Set(circularRefs.keys()));
 
     const shouldDisplayExample = (schema: OpenAPIV3.SchemaObject): boolean => {
-        schema.example;
         return (typeof schema.example === 'string' || typeof schema.example === 'number' || typeof schema.example === 'boolean')
     }
     return (
--- a/packages/react-openapi/src/OpenAPISchema.tsx
+++ b/packages/react-openapi/src/OpenAPISchema.tsx
@@ -14,7 +14,6 @@ interface OpenAPISchemaPropertyEntry {
     propertyName?: string;
     required?: boolean;
     schema: OpenAPIV3.SchemaObject;
-    example?: string;
 }
 
 /**
@@ -48,6 +47,10 @@ export function OpenAPISchemaProperty(
         ? null
         : getSchemaAlternatives(schema, new Set(circularRefs.keys()));
 
+    const shouldDisplayExample = (schema: OpenAPIV3.SchemaObject): boolean => {
+        schema.example;
+        return (typeof schema.example === 'string' || typeof schema.example === 'number' || typeof schema.example === 'boolean')
+    }
     return (
         <InteractiveSection
             id={id}
@@ -93,8 +96,8 @@ export function OpenAPISchemaProperty(
                             className="openapi-schema-description"
                         />
                     ) : null}
-                    {schema.example ? (
-                        <span className="openapi-schema-example">Example: {schema.example} </span>
+                    {shouldDisplayExample(schema) ? (
+                        <span className="openapi-schema-example">Example: <code>{schema.example}</code></span>
                     ) : null}
                 </div>
             }
--- a/packages/react-openapi/src/OpenAPISpec.tsx
+++ b/packages/react-openapi/src/OpenAPISpec.tsx
@@ -48,6 +48,7 @@ export function OpenAPISpec(props: { rawData: any; context: OpenAPIClientContext
                                 // Description of the parameter is defined at the parameter level
                                 // we use display it if the schema doesn't override it
                                 description: parameter.description,
+                                example: parameter.example,
                                 ...(noReference(parameter.schema) ?? {}),
                             },
                             required: parameter.required,
--- a/src/components/DocumentView/OpenAPI/style.css
+++ b/src/components/DocumentView/OpenAPI/style.css
@@ -173,7 +173,7 @@
 }
 
 .openapi-schema-example {
-    @apply mt-2;
+    @apply mt-2 text-base text-dark/10 dark:text-light/10;
 }
 
 /** Authentication */
--- a/src/components/DocumentView/OpenAPI/style.css
+++ b/src/components/DocumentView/OpenAPI/style.css
@@ -172,8 +172,8 @@
     @apply prose-sm;
 }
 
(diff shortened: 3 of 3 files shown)


Eloutmanixx (approved):

Eloutmanixx (approved):
xxxxxxxxxx

packages/react-openapi/src/OpenAPISchema.tsx
```diff
@@ -92,6 +93,12 @@ export function OpenAPISchemaProperty(
                             className="openapi-schema-description"
                         />
                     ) : null}
+                    {schema.example ? (
```
SamyPesse, line 96: `example` is not supposed to be markdown so why use markdown here? (https://swagger.io/docs/specification/adding-examples/)

It can result in problems, ex: if an `example` contains a markdown string, it'll be processed there when the user expects to see the raw string.

We want to render a value instead of markdown text.

SamyPesse: Also the spacing with the description is not great, we should use a gap between the 2

SamyPesse (changes requested):

packages/react-openapi/src/OpenAPISchema.tsx
vibhanshub: This is what it looks like now
Screen Shot 2024-05-15 at 2 20 31 PM

vibhanshub: Using
Screen Shot 2024-05-15 at 2 28 24 PM
 margin instead of padding now

vibhanshub: Also, argos is complaining for the right reasons as Examples are not present in argos. Can I approve it or do I need to update the images argos is using as a baseline? (first time working with argos where errors are justified)

scazan: If the change is expected and the new version you see in Argos is the correct one, simply approve it and it'll update the baseline.

taranvohra: The font styling looks off to me. Have you checked that?

scazan (approved):

packages/react-openapi/src/OpenAPISchema.tsx
taranvohra: The text color seems to stand about a bit too much. Is it black (#000)? Is it the same text color as the property name (`iso-id` in your screenshot)

vibhanshub: This is what it looks like with the same color as property name

```
.openapi-schema-example {
    @apply mt-2 text-base text-dark/10 dark:text-light/10;
}
```

Screen Shot 2024-05-20 at 4 50 00 PM

Looks the same to me

taranvohra: Idk why but this looks a bit better to me haha. It's subtle but it doesn't come across as fully black. Anyway, lgtm otherwise

taranvohra (approved):

packages/react-openapi/src/OpenAPISchema.tsx
```diff
@@ -14,6 +14,7 @@ interface OpenAPISchemaPropertyEntry {
     propertyName?: string;
     required?: boolean;
     schema: OpenAPIV3.SchemaObject;
+    example?: string;
```
SamyPesse, line 17: I don't understand why we need this prop. Isn't `example` included in the `schema` prop (https://github.com/kogosoftwarellc/open-api/blob/main/packages/openapi-types/index.ts#L166) ?

SamyPesse: Ok it looks like `example` is for parameters and `examples` is for schema fields. I think we should try to handle both the same way.

packages/react-openapi/src/OpenAPISchema.tsx
```diff
@@ -92,6 +93,9 @@ export function OpenAPISchemaProperty(
                             className="openapi-schema-description"
                         />
                     ) : null}
+                    {schema.example ? (
+                        <span className="openapi-schema-example">Example: {schema.example} </span>
```
SamyPesse, line 97: We should use a `<code>` or `<pre>` around the value.

Also, how does it work if my example is not a primitive?

ex:

```
example:
  propA: a
  propB: b
```

My take is that it'll render `[object Object]` which is bad.

packages/react-openapi/src/OpenAPISchema.tsx
SamyPesse: It looks like you don't actually use this

```suggestion
```

and you use `schema.example`.

src/components/DocumentView/OpenAPI/style.css
```diff
@@ -172,6 +172,10 @@
     @apply prose-sm;
 }
 
+.openapi-schema-example {
+    @apply mt-2 text-base text-dark/10 dark:text-light/10;
```
SamyPesse, line 176: Why are we using a larger text font than the description?

It doesn't look good in the screenshot.

SamyPesse (changes requested):
IMO it can't be merged in the current state and it requires more work:
- Properly support non-primitive values?
  - should we represent with `JSON.stringify`? If yes, it'll not work well for XML/etc API
  - should we skip non-primitives using `typeof schema.example === 'string' || typeof schema.example === 'number' || typeof schema.example === 'boolean'`?
- Should we handle `examples` as well by selecting the first example or the one matching the media-type?

packages/react-openapi/src/OpenAPISchema.tsx
vibhanshub: @SamyPesse it looks like you got it reversed. `example` is for schema fields, and `examples` is for parameters
Screen Shot 2024-05-21 at 2 56 09 PM

https://swagger.io/docs/specification/adding-examples/

vibhanshub: I also think we should keep this PR just for the singular `example` and not support `examples` for now. What do you think?

SamyPesse: Yes, but it doesn't change the fact that this prop is actually not used:

```suggestion
```

packages/react-openapi/src/OpenAPISchema.tsx
```diff
@@ -92,6 +93,9 @@ export function OpenAPISchemaProperty(
                             className="openapi-schema-description"
                         />
                     ) : null}
+                    {schema.example ? (
```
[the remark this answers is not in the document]
SamyPesse: It's about the type. `schema.description` is always a string and if it's empty we don't want to show an empty div.

`schema.example` can be any value, depending on the schema.

packages/react-openapi/src/OpenAPISpec.tsx
```diff
@@ -48,6 +48,7 @@ export function OpenAPISpec(props: { rawData: any; context: OpenAPIClientContext
                                 // Description of the parameter is defined at the parameter level
                                 // we use display it if the schema doesn't override it
                                 description: parameter.description,
+                                example: parameter.example,
```
SamyPesse, line 51: Why do we need this? Isn't `example` on the `parameter.schema` itself.

packages/react-openapi/src/OpenAPISchema.tsx
SamyPesse: Are you doing the changes here?

packages/react-openapi/src/OpenAPISpec.tsx
vibhanshub: To maintain compatibility with Swagger 2.0 (in our docs we say we support 2.0). In 2.0, `example` is not on `parameter.schema` but on `parameter` itself.

In 3.0, `example` is on `parameter.schema`, so the destructuring happening in the line below will take of it

packages/react-openapi/src/OpenAPISchema.tsx
vibhanshub: removed

packages/react-openapi/src/OpenAPISchema.tsx
vibhanshub: done. this is what it looks like now
Screen Shot 2024-05-22 at 11 49 45 AM

src/components/DocumentView/OpenAPI/style.css
vibhanshub: fixed. see the screenshot above

packages/react-openapi/src/OpenAPISchema.tsx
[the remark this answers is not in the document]
vibhanshub: fixed

packages/react-openapi/src/OpenAPISpec.tsx
vibhanshub: https://swagger.io/docs/specification/2-0/adding-examples/

daskinw:
@SamyPesse this is blocking for an ICP customer going live with their API docs. I think Vib made the requested changes. Can we get it to move?

packages/react-openapi/src/OpenAPISchema.tsx
```diff
@@ -92,6 +95,9 @@ export function OpenAPISchemaProperty(
                             className="openapi-schema-description"
                         />
                     ) : null}
+                    {shouldDisplayExample(schema) ? (
+                        <span className="openapi-schema-example">Example: <code>{schema.example}</code></span>
```
SamyPesse, line 99: I mentioned it a few times already; but until we solve this part we should not merge this PR: this line is wrong.

`schema.example` can be any value; and the way you display it will not work with non-primitive values.

Example taken from the doc https://swagger.io/docs/specification/adding-examples/:

```yaml
paths:
  /users:
    post:
      summary: Adds a new user
      requestBody:
        content:
          application/json:
            schema:      # Request body contents
              type: object
              properties:
                id:
                  type: integer
                name:
                  type: string
              example:   # Sample object
                id: 10
                name: Jessica Smith
      responses:
        '200':
          description: OK
```

With this case, it'll display `Example: [object Object]` which is shit 💩 

This is a very common use case, so we need to handle it properly. I can help over a 30-minute pair programming session, or you can iterate on it.

SamyPesse (changes requested):

packages/react-openapi/src/OpenAPISchema.tsx
SamyPesse: To explain more and why we can't merge it as is: you are solving it for one customer by handling only primitives; every customers with non-primitive examples (like in the one above) will be `Example: [object Object]` and will consider worst than not having it. Basically, this PR will be a net-negative for all users.

SamyPesse: Haa I see that you added the non-prmitive... ok let's merge but fix the boolean case first:

Otherwise it'll display `Example: `

SamyPesse, line 99:
```suggestion
                        <span className="openapi-schema-example">Example: <code>{JSON.stringify(schema.example)}</code></span>
```

SamyPesse (approved):

packages/react-openapi/src/OpenAPISchema.tsx
vibhanshub: Done. Will merge the PR now

PR #2337, opened 2024-06-10
Remove cache client-side

gregberge:
Since we have to invalidate the content, it's not OK to have a cache client-side because we can't invalidate this one.

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -261,7 +261,7 @@ export async function middleware(request: NextRequest) {
         );
     } else {
         if (resolved.cacheMaxAge) {
-            const cacheControl = `public, max-age=0, s-maxage=${resolved.cacheMaxAge}, stale-while-revalidate=60, stale-if-error=0`;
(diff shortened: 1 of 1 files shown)


src/middleware.ts
```diff
@@ -261,7 +261,7 @@ export async function middleware(request: NextRequest) {
         );
     } else {
         if (resolved.cacheMaxAge) {
-            const cacheControl = `public, max-age=60, s-maxage=${resolved.cacheMaxAge}, stale-while-revalidate=60, stale-if-error=0`;
+            const cacheControl = `public, max-age=0, s-maxage=${resolved.cacheMaxAge}, stale-while-revalidate=60, stale-if-error=0`;
```
SamyPesse, line 264: Do you want to remove the `stale-while-revalidate` as well?

Otherwise it might still result in 60s, in the edge, but also on the client

SamyPesse (approved):

PR #2339, opened 2024-06-12
Pass site url as context to `listSiteSpaces` API

taranvohra:
This PR fixes the Space switcher for share-links sites. Currently the spaces rendered in the switcher does not have a published URL and breaks navigation. It is done on purpose to avoid leaking share-links. The `listSiteSpaces` API has support for populating published URLs for site-spaces if the site URL is provided as `context` to the API.


gregberge (approved):

src/lib/api.ts
```diff
@@ -736,12 +740,28 @@ export const getSite = cache(
...
 export const getSiteSpaces = cache(
     'api.getSiteSpaces',
-    async (organizationId: string, siteId: string, options: CacheFunctionOptions) => {
+    async (
+        organizationId: string,
+        siteId: string,
+        /**
+         * Additional site URL that can be used as context to resolve site space published urls
+         */
+        siteUrlContext: string | null,
```
emmerich, line 753: By passing as an arg here it means it'll form part of the cache key so multiple requests with the same orgId/siteId will not be cached:

```
getSiteSpaces(':orgId', ':siteId', 'https://docs.gitbook.com') => CACHE MISS
getSiteSpaces(':orgId', ':siteId', 'https://docs.gitbook.com/getting-started') => CACHE MISS, even though data is the same
```

I'm not sure the best solution for it. If we don't include it in the cache key it could mean caching the same content between different share links, but including it will mean effectively no cache.

gregberge: It's not good. We have to fix that.

PR #2343, opened 2024-06-17
Add `getContentTitle` to properly render the title for sites & legacy published content

taranvohra:


spastorelli (approved):
LGTM. Looks like there are other places where the title seems to be used, can you update them as well if needed, e.g:
https://github.com/GitbookIO/gitbook/blob/main/src/app/(space)/(core)/~gitbook/icon/route.tsx#L57
https://github.com/GitbookIO/gitbook/blob/main/src/app/(space)/(core)/~gitbook/ogimage/%5BpageId%5D/route.tsx#L34

gregberge (approved):

PR #2347, opened 2024-06-18
Share memory cache between middleware and request handler

jpreynat:
~~This PR reduces the work done in the middleware, since some of it is useless.~~

~~The most certain one is that we call either `get(Space|CurrentSite)LayoutData` that fetches both the scripts and customization for the space/site. However, we only need the scripts for the current content to compute the CSP. Fetching and waiting for the customization just blocks the whole request uselessly.~~

~~Then we also prefetch the space/site content data in the middleware, but I'm pretty sure that the middleware execution is isolated from the actual route handler, meaning that this prefetch via `waitUntil`, even if not blocking the request, is useless.~~

EDIT:
Since I've checked in logs that `globalThis` isn't shared between the middleware and the request handlers, this PR instead makes sure that the memory cache is now properly shared between both using Cloudflare's request context.

### DIFF
--- a/src/lib/cache/memory.ts
+++ b/src/lib/cache/memory.ts
@@ -1,5 +1,6 @@
 import { CacheBackend, CacheEntry } from './types';
 import { NON_IMMUTABLE_LOCAL_CACHE_MAX_AGE_SECONDS, isCacheEntryImmutable } from './utils';
+import { getGlobalContext } from '../waitUntil';
 
 export const memoryCache: CacheBackend = {
     name: 'memory',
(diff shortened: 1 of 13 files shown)


src/middleware.ts
```diff
@@ -169,23 +166,9 @@ export async function middleware(request: NextRequest) {
...
         async () => {
-            // Start fetching everything as soon as possible, but do not block the middleware on it
-            // the cache will handle concurrent calls
-            await waitUntil(
-                getSpaceContentData({
-                    spaceId: resolved.space,
-                    changeRequestId: resolved.changeRequest,
-                    revisionId: resolved.revision,
-                }),
-            );
```
SamyPesse, line 180: I don't understand why you removed it and called it useless. I measured it at the time, and it was efficient.

It doesn't block anything because: we don't await it, we only start the execution asap, and since the cache is maintaining only one operation per cache key, it's being reused when needed during rendering

src/lib/cache/memory.ts
```diff
@@ -66,13 +66,21 @@ export const memoryCache: CacheBackend = {
...
+    let globalThisForMemoryCache: any = globalThis;
+
+    if (process.env.NODE_ENV !== 'test') {
+        // We lazy-load the next-on-pages package to avoid errors when running tests because of 'server-only'.
+        const { getOptionalRequestContext } = await import('@cloudflare/next-on-pages');
+        const cloudflare = getOptionalRequestContext();
+        if (cloudflare) {
+            globalThisForMemoryCache = cloudflare.ctx;
+        }
+    }
```
SamyPesse, line 79: Can we use https://github.com/GitbookIO/gitbook/blob/main/src/lib/waitUntil.ts#L7?

or at least keep things centralize on this file

SamyPesse (approved):
Nice!

src/lib/cache/memory.ts
jpreynat: Ah right, good call 👍

PR #2349, opened 2024-06-19
Parse cache tags returned from the API set them in response

jpreynat:
This will allow the response from GitBook Open to set the `cache-tag` and `x-gitbook-cache-tag` headers, allowing Cloudflare to actually use them and to purge them.


scazan (approved):
Don't know the full context here but approving cause it seems correct in that we weren't passing the tags at all previously.

PR #2352, opened 2024-06-21
Forward content resolution API response tags to local caches

jpreynat:
This will also ensure that content cache be refreshed on demand in "local" (meaning this application's) caches (memory, Cloudflare cache, KV, ...).

### DIFF
--- a/src/lib/api.ts
+++ b/src/lib/api.ts
@@ -243,6 +243,7 @@ export const getPublishedContentByUrl = cache(
             };
             return {
                 tags: [
+                    ...parsed.tags,
                     getAPICacheTag({
                         tag: 'url',
                         hostname: parsedURL.hostname,


src/lib/api.ts
```diff
@@ -243,6 +243,7 @@ export const getPublishedContentByUrl = cache(
             };
             return {
                 tags: [
+                    ...parsed.tags,
```
SamyPesse, line 246: I removed it on purpose as otherwise it creates too many caches (more expensive on KV, redis as it requires a lot of write); for tags that we don't use.

So now the tags are not related to the API server tags.

jpreynat: But in the case of URLs, it means that if a content is unpublished, unless we manually try to revalidate the tags for all of its possible URLs from a remote, they will always continue to serve this content until the cache period ends.
I don't think that it's manageable for this part, and we should instead at least use the content type and key as cache tags for the local GBO caches to be able to invalidate them.

SamyPesse: I'm not sure to understand, we have a tag for the published url resoution (the domain); we just need to invalidate it when needed.

jpreynat: Yeah ok 🤔 
I'm just thinking that this can add a bit of overhead, but I understand the logic here.
Let's keep it like this for now then.

jpreynat:
Closed as not necessary
https://github.com/GitbookIO/gitbook/pull/2352#pullrequestreview-2132272427

PR #2354, opened 2024-06-21
Set preconnect links to main domains on render

jpreynat:
Preconnect to the main external domains used (API, iframely for embeds, ...) during rendering to send hints to the browser.

### DIFF
--- a/src/components/DocumentView/Embed.tsx
+++ b/src/components/DocumentView/Embed.tsx
@@ -1,5 +1,6 @@
 import * as gitbookAPI from '@gitbook/api';
 import Script from 'next/script';
+import ReactDOM from 'react-dom';
 
 import { Card } from '@/components/primitives';
 import { api } from '@/lib/api';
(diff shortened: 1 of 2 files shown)


src/components/DocumentView/Embed.tsx
```diff
@@ -11,6 +12,8 @@ import { IntegrationBlock } from './Integration';
 
 export async function Embed(props: BlockProps<gitbookAPI.DocumentBlockEmbed>) {
     const { block, context, ...otherProps } = props;
+    
+    ReactDOM.preconnect('https://cdn.iframe.ly');
```
taranvohra, line 16: nit: Maybe define a constant like we have for `assetsDomain`?

taranvohra (approved):

src/components/DocumentView/Embed.tsx
SamyPesse: Why ?

SamyPesse (approved):

PR #2357, opened 2024-06-24
Bump next@14.2.4

jpreynat:
It has a bunch of build improvements that we could benefit from:
https://nextjs.org/blog/next-14-2#build-and-production-improvements


jpreynat:
Switching the PR to draft for now, as it seems to create important changes for rendering, especially with header links

PR #2359, opened 2024-06-25
Tree shake unused code for Sentry

jpreynat:
Noticed that one of the JS chunks for the app was quite big, and it included a whole lot of code dedicated to Sentry only, which we don't use to a lot of extent client-side (no integrations).

Searched and found [this GitHub issue](https://github.com/getsentry/sentry-javascript/issues/7680), talking about [this tree shaking guide](https://docs.sentry.io/platforms/javascript/guides/nextjs/configuration/tree-shaking/).

Applying the suggestions led to removing ~40KB of gzipped code, or ~136KB of uncompressed JS, which is ~10% of the first load JS code size.

### Before (main)
image

### After
image

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -25,6 +25,7 @@ module.exports = withSentryConfig(
                 config.plugins.push(
                     new webpack.DefinePlugin({
                         __SENTRY_DEBUG__: false,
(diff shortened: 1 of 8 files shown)


scazan (approved):
nice improvement!

PR #2361, opened 2024-06-25
Use latest bun@1.1.17 in CI

jpreynat:


SamyPesse:
It was changed already https://github.com/GitbookIO/gitbook/blob/main/.github/workflows/ci.yaml#L23

PR #2362, opened 2024-06-27
Prevent SVG images to be resized and rendered on the /~gitbook/image endpoint

jpreynat:
The `Image` component endpoint prevents generating resize URLs for SVG files already.

However, the `/~gitbook/image` resize endpoint doesn't recheck if the URL can (and should) be resized, which includes SVG files.

This PR improves the `checkIsSizableImageURL` helper to ensure that the URL can be parsed, and that the targeted filename, via the pathname, ends in `.svg`, instead of checking the URL that could have a query string parameter.
It also uses this helper in the image resizing endpoint to prevent resizing and thus rendering SVGs on the domain used to serve the docs.

### DIFF
--- a/src/app/(global)/~gitbook/image/route.ts
+++ b/src/app/(global)/~gitbook/image/route.ts
@@ -1,6 +1,6 @@
 import { NextRequest } from 'next/server';
 
-import { verifyImageSignature, resizeImage, CloudflareImageOptions } from '@/lib/images';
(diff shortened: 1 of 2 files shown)


emmerich (approved):
Looks good. I wonder if it fixes the vulnerability because Next itself has an image resizing UI, but they should handle it on their side.

PR #2364, opened 2024-07-02
Use published URL for site when displaying results for a Site space

scazan:
Published URLs for spaces are used in search results but these are not relevant when searching within Sites. This causes a bug on Sites currently where the search result for a different variant takes you to a non-existent page (as it is missing the /v/ portion of the space's published URL).
We need to, instead, use the SiteSpace published URL. This fixes the issue for the moment by adding the published URL for the SiteSpace into the Space result.


gregberge (commented):
This URL is not always present (not published sites or share links).

taranvohra (approved):

PR #2367, opened 2024-07-03
Don't use customized site title for space in search selector

scazan:
This fixes an issue where the site's customized title is used in place of the space title in the search modal when filtering by space id.


src/components/SpaceLayout/SpaceLayout.tsx
```diff
@@ -114,7 +114,7 @@ export function SpaceLayout(props: {
                 <SearchModal
                     spaceId={contentTarget.spaceId}
                     revisionId={contentTarget.revisionId}
-                    spaceTitle={customization.title ?? space.title}
+                    spaceTitle={space.title}
```
emmerich, line 117: I think it's a breaking change for users who have set the title in customization.

What we really need is:

```siteSpaceCustomization.title ?? space.title```

but do we have access to it?

**Review by emmerich — DISMISSED** [review]

File: packages/gitbook/src/app/(space)/fetch.ts [hunk 2]
```diff
@@ -87,11 +88,12 @@ export async function fetchSpaceData() {
     );
 
     const parent = await (parentSite ?? fetchParentCollection(space));
+    const spaceRelativeToParent = parent?.spaces.find((space) => space.id === content.spaceId);
```
taranvohra, line 91: Why do we need to do this?

packages/gitbook/src/app/(space)/fetch.ts
```diff
@@ -195,23 +197,34 @@ async function fetchParentSite(args: {
     siteShareKey: string | undefined;
 }) {
     const { organizationId, siteId, siteShareKey } = args;
-    const [site, siteSpaces] = await Promise.all([
+    const [site, siteSpaces, siteParentCustomizations] = await Promise.all([
         getSite(organizationId, siteId),
         getSiteSpaces({ organizationId, siteId, siteShareKey }),
+        siteId ? getCurrentSiteCustomization({ organizationId, siteId }) : null,
```
taranvohra, line 203: `siteId` will always be defined in this method so no need to do a ternary

packages/gitbook/src/lib/api.ts
```diff
@@ -722,7 +722,7 @@ export const getSiteSpaces = cache(
             organizationId: string;
             siteId: string;
             /** Site share key that can be used as context to resolve site space published urls */
-            siteShareKey: string | undefined;
+            siteShareKey?: string;
```
taranvohra, line 725: We should not make it optional. It causes bugs. It was typed like this so we always specify a `shareKey` when we have it otherwise the urls in the object is not properly resolved (for share link sites)

packages/gitbook/src/lib/api.ts
```diff
@@ -818,7 +818,7 @@ export async function getCurrentSiteLayoutData(args: {
 export async function getCurrentSiteCustomization(args: {
     organizationId: string;
     siteId: string;
-    siteSpaceId: string | undefined;
+    siteSpaceId?: string;
```
taranvohra, line 821: Same comment about making it optional

packages/gitbook/src/lib/utils.ts
```diff
@@ -12,14 +12,14 @@ import {
 export function getContentTitle(
     space: Space,
     customization: CustomizationSettings | SiteCustomizationSettings,
-    parent: Site | Collection | null,
+    parent?: Site | Collection | null,
```
taranvohra, line 15: Why add an optional here?

taranvohra (commented):
Can we add some e2e tests to assert title is being rendered correctly for a bunch of cases?

packages/gitbook/src/lib/api.ts
scazan: It's just a syntactical change to make it more standard. This is already optional as it is typed as `string | undefined`. If we don't want it to be optional we should remove undefined from the type.

packages/gitbook/src/lib/api.ts
scazan: Why are typing it as potentially undefined in that case?

packages/gitbook/src/app/(space)/fetch.ts
scazan: We do this because when we query the site we override all sitespaces with customizations. When we query the space alone we don't (these are sitespace-level customizations as opposed to space-level so I assume that is why and it makes sense).

Perhaps at this point we can do the overrides in the space query as well since everyone is on sites but there is a bit more work and testing to be done to achieve that I think.

wdyt?

packages/gitbook/src/lib/api.ts
scazan: my guess is that what we really want here then is `string | null`

packages/gitbook/src/lib/utils.ts
scazan: ah! I think this is no longer necessary and was related to old published spaces.

packages/gitbook/src/lib/api.ts
taranvohra: Yes `string | null` is fine but we absolutely need to not make it optional

packages/gitbook/src/lib/api.ts
taranvohra: Yep `string | null` should be good

packages/gitbook/src/app/(space)/fetch.ts
taranvohra: Sorry but I still don't understand it fully. The spaces coming from `fetchParentSite` don't have any customizations. The `space` is used from the resolved `pointer`

scazan: it does after the changes we made earlier here: https://github.com/GitbookIO/gitbook/blob/bb83e05d06c1035517fd10557c460c89ce2cb4e4/packages/gitbook/src/app/(space)/fetch.ts#L211

taranvohra: Ah okay makes sense 👍

packages/gitbook/src/lib/utils.ts
```diff
@@ -17,9 +17,9 @@ export function getContentTitle(
     // When we are rendering a site, always give priority to the customization title first
     // and then fallback to the site title
     if (parent?.object === 'site') {
-        return customization.title ?? parent.title ?? space.title;
+        return parent.title ?? space.title;
```
taranvohra, line 20: Can you add a comment that the `parent.title` would already be overridden with the customization tittle while fetching the site

taranvohra (approved):

PR #2370, opened 2024-07-05
Use schema.example if defined, even if value is falsy

SamyPesse:


sdirosa (approved):

taranvohra (approved):

packages/react-openapi/src/generateSchemaExample.ts
```diff
@@ -19,7 +19,7 @@ export function generateSchemaExample(
         return undefined;
     }
 
-    if (schema.example) {
+    if (typeof schema.example !== 'undefined') {
```
gregberge, line 22: `schema.example !== undefined` is fine

PR #2372, opened 2024-07-10
Fix issue where nav route/path is not retained when switching variants

spastorelli:


src/components/Header/SpacesDropdown.tsx
```diff
@@ -1,11 +1,16 @@
-import { Collection, Space } from '@gitbook/api';
+'use client';
```
emmerich, line 1: The downside to having `use client` on this component is that the whole dropdown will no longer be a server component, even though it's only the individual menu items that need the current layout segment.

I'd create a new component called `SpacesDropdownItem`:

```
'use client';

export function SpacesDropdownItem(props: { space: Space; }) {
  const currentPathname = useSelectedLayoutSegment() ?? '';
  return <DropdownItem ... />
}
```

so we have Client Components further down the stack

src/components/Header/SpacesDropdown.tsx
```diff
@@ -29,15 +30,9 @@ export function SpacesDropdown(props: { space: Space; spaces: Space[] }) {
...
-                    <DropdownMenuItem
-                        key={otherSpace.id}
-                        href={otherSpace.urls.published ?? otherSpace.urls.app}
-                        active={otherSpace.id === space.id}
-                    >
-                        {otherSpace.title}
-                    </DropdownMenuItem>
-                ))}
+                {spaces.map((otherSpace, index) => 
+                    <SpacesDropdownMenuItem key={`${otherSpace.id}-${index}`} currentSpace={space} variantSpace={otherSpace} />
```
emmerich, line 34:
```suggestion
                    <SpacesDropdownMenuItem key={otherSpace.id} currentSpace={space} variantSpace={otherSpace} />
```

the id should be unique I guess?

src/components/Header/SpacesDropdownMenuItem.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { Space } from '@gitbook/api';
+import { useSelectedLayoutSegment } from 'next/navigation';
+
+import { DropdownMenuItem } from './Dropdown';
+
+export function SpacesDropdownMenuItem(props: { variantSpace: Space; currentSpace: Space }) {
+    const { variantSpace, currentSpace } = props;
+    const currentPathname = useSelectedLayoutSegment() ?? '';
+    const targetUrl = new URL(variantSpace.urls.published ?? variantSpace.urls.app);
+    targetUrl.pathname += `/${currentPathname}`;
```
emmerich, line 12: Screenshot 2024-07-10 at 13 58 36

Will it create double slashes?

emmerich (approved):
Approve with a couple of comments to check

src/components/Header/SpacesDropdownMenuItem.tsx
emmerich: It relies on the URL implementation which I think is a standard so we should be fine.. BUT because this code could run on the server, client etc. I think it's best to be safe here.

src/components/Header/SpacesDropdown.tsx
spastorelli: I have set the key to `${otherSpace.id}-${index}` instead of just `otherSpace.id` because with sites there could be 2 site spaces that points to the same space so using the space id will not make it unique.

src/components/Header/SpacesDropdownMenuItem.tsx
emmerich: But you're appending to the pathname:

```
const targetUrl = new URL(variantSpace.urls.published ?? variantSpace.urls.app); // targetUrl.pathname is "/"
targetUrl.pathname += `/${currentPathname}`; // targetUrl.path is "/" + "/hello-world"
```

Shouldn't it be:

```suggestion
    targetUrl.pathname = `/${currentPathname}`;
```

spastorelli: No we need to append it because `variantSpace.urls.published ?? variantSpace.urls.app` can already contain a pathname that we need to preserve, e.g
```
https://gitbook-open-e2e-sites.gitbook.io/api-multi-versions/v/2.0
```

or for share links:

```
https://gitbook-open-e2e-sites.gitbook.io/api-multi-versions-share-links/bRfQbzwsK8rbN1GRxx7K/v/2.0
```

btw if I fed to new URL the value returned by `variantSpace.urls.published ?? variantSpace.urls.app` I'm not getting doubleslashes 
Screenshot 2024-07-10 at 15 22 48

I wondering if it's because, in this case urls.published is returned with a non root path unlike your example where you didn't provide one. 

Going to sanitize the pathname anyway but we definitely need to prepend it instead of overwritting it.

emmerich: Makes sense 👍  - yes I think a quick sanitization will help as we can never be 100% sure of the URL implementation we're dealing with. Anyways PR is approved so feel free to merge when you're happy with it.

src/components/Header/SpacesDropdownMenuItem.tsx
```diff
@@ -0,0 +1,30 @@
...
+function useVariantSpaceHref(variantSpace: Space) {
+    const currentPathname = useSelectedLayoutSegment() ?? '';
+    const targetUrl = new URL(variantSpace.urls.published ?? variantSpace.urls.app);
+    targetUrl.pathname += `/${currentPathname}`;
+    targetUrl.pathname = targetUrl.pathname.replace(/\/{2,}/g, '/').replace(/\/$/, '');
+
+    return targetUrl.toString();
+}
+
+export function SpacesDropdownMenuItem(props: { variantSpace: Space; currentSpace: Space }) {
```
SamyPesse, line 17: @spastorelli keep in mind that when you do a client component like this, all props will have to be serialized in the HTML. So instead of passing the whole object when you use only 1/2 properties, instead pass the few properties as dedicated props.

In that case `active`, `targetUrl` could have been computed in the parent servr component.

spastorelli: I can definitely reduce the props to only what's needed from space and variantSpace instead of the full object but not sure about `targetUrl`.

spastorelli: actually nevermind sorry I just realized that by targetUrl you meant the value of `variantSpace.urls.published ?? variantSpace.urls.app` 🤦 Sorry about this

spastorelli: @SamyPesse https://github.com/GitbookIO/gitbook/pull/2373

PR #2374, opened 2024-07-11
Remove Page/Space label from BlockContentRef

BrettJephson:
In this change, we:

- adds emoji to Space content ref
- remove the kind (Page/Space) pre title
- increases size of the leading emoji in the block

### Preview

Screenshot 2024-07-11 at 10 53 48

### DIFF
--- a/src/components/DocumentView/BlockContentRef.tsx
+++ b/src/components/DocumentView/BlockContentRef.tsx
@@ -1,12 +1,11 @@
-import { CustomizationThemedURL, DocumentBlockContentRef } from '@gitbook/api';
+import { DocumentBlockContentRef } from '@gitbook/api';
 
+import { LogoIcon } from '@/components/icons';
 import { Card, Emoji } from '@/components/primitives';
 import { getSpaceCustomization, ignoreAPIError } from '@/lib/api';
 import { ResolvedContentRef } from '@/lib/references';
 
 import { BlockProps } from './Block';
-import { Image } from '../utils';
-
 
 export async function BlockContentRef(props: BlockProps<DocumentBlockContentRef>) {
     const { block, context, style } = props;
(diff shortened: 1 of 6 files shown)


src/lib/references.ts
```diff
@@ -176,6 +176,7 @@ export async function resolveContentRef(
                 href: targetSpace.urls.published ?? targetSpace.urls.app,
                 text: targetSpace.title,
                 active: true,
+                emoji: targetSpace.emoji
```
emmerich, line 179: `space.emoji` is a pretty internal thing to show on published docs. I'm not sure it's what users would expect.

At the very least they should expect to see the logo they configure in customization to take precedence over the emoji.

You can see how we decide to render the space logo in `LogoFallback` in `HeaderLogo` and I think we should use the same logic here. I'd argue that the emoji shouldn't be shown at all (falling back to the GitBook logo just like HeaderLogo does)

emmerich (changes requested):

src/lib/references.ts
BrettJephson: Sure, I added `targetSpace.urls.logo` but I'll take a look at the logic in `LogoFallback` instead. Thanks for the pointer!

src/lib/references.ts
```diff
@@ -176,6 +178,7 @@ export async function resolveContentRef(
                 href: targetSpace.urls.published ?? targetSpace.urls.app,
                 text: targetSpace.title,
                 active: true,
+                emoji: targetSpace.emoji,
```
emmerich, line 181: I'd remove this in case other components are displaying this property in some way. I think the emoji should be an entirely hidden property and never shown in public content. 

```suggestion
```

src/components/DocumentView/BlockContentRef.tsx
```diff
@@ -1,27 +1,77 @@
...
         />
     );
 }
+
+function BlockContentRefIcon(props: { icon: CustomizationThemedURL }) {
+    const { icon } = props;
+            
+    return icon ? <Image 
+        priority='lazy' 
+        alt=""
```
emmerich, line 63:
```suggestion
        alt={`Logo for ${resolved.text}`}
```

Let's set an alt text

src/lib/references.ts
BrettJephson: Using the customization icon instead now. And resolving it in the component.

src/components/DocumentView/BlockContentRef.tsx
```diff
@@ -1,27 +1,77 @@
...
+
+    const spaceCustomization = await ignoreAPIError(getSpaceCustomization(spaceId));
+    const customFavicon = spaceCustomization?.favicon;
+    const customEmoji = customFavicon && 'emoji' in customFavicon ? customFavicon.emoji : null;
+    const customIcon = customFavicon && 'icon' in customFavicon ? customFavicon.icon : null;
+
     return (
         <Card
-            leadingIcon={resolved.emoji ? <Emoji code={resolved.emoji} /> : null}
+            leadingIcon={customIcon ? <BlockContentRefIcon icon={customIcon} /> : customEmoji ? <Emoji code={customEmoji} style="text-xl" /> : null}
```
emmerich, line 50: If there's no icon or emoji, we fallback to null. That's different to `HeaderLogo` which falls back to the GitBook logo:

https://github.com/GitbookIO/gitbook/blob/main/src/components/Header/HeaderLogo.tsx#L114

Should we make sure that both act in the same way, so that the icon a user sees in the ContentRef card also matches the icon they'll see at the top of the page if they follow the link?

emmerich (changes requested):

src/components/DocumentView/BlockContentRef.tsx
BrettJephson: I'm not sure that is better. 

An empty alt text tends to be the way to handle decorative images, in other words, images that don't provide any useful information. They get ignored by a screen reader.  Similar example from [W3C](https://www.w3.org/WAI/tutorials/images/decorative/#example-2-decorative-image-as-part-of-a-text-link).

Knowing that there is a logo there doesn't really have any value if you can't see the logo. Unless we describe the logo it's just adding noise.

emmerich: Okay, I did see that in `HeaderLogo` we use an alt text to maybe worth removing that too.

src/components/DocumentView/BlockContentRef.tsx
BrettJephson: Ok, I've made a `LogoIcon` that makes sure the logic is consistent between both. I included the customization emoji because it seems like it should show if you've selected that customization and not an icon image.

emmerich (approved):

PR #2376, opened 2024-07-11
Fallback redirect to root when switching variants

spastorelli:


src/app/(space)/(content)/[[...pathname]]/page.tsx
Comment (spastorelli) [inline src/app/(space)/(content)/[[...pathname]]/page.tsx:1]: @SamyPesse @emmerich Can I please get your eyes and help on this? I know the PR is still wip/draft but I'm still unsure about what would be the best approach here. Sorry very new to NextJs.

I've implemented 2 approaches so far (see current changes now for approach 2 and [this commit](https://github.com/GitbookIO/gitbook/pull/2376/commits/82781adc84185fdd5f2824c161d4e4276ea62de5) for approach 1) and also thinking about another one (handle fallback at middleware level).

Mainly what I'm unsure is whether to handle the redirect/fallback at the middleware level or at the `Page` route handler.

- The **advantage** I see to handle it at the middleware level is that we could handle it as a proper request redirect (e.g 307) and allow us to rewrite the request URL at the server side (helping to remove the `fallback` param to avoid being sticky). The **disadvantage** is that we would need to fetch the page data which we don't do right now (we do it only further down the line in the `Page` route handler) and could mean a performance impact.

- The **advantage** I see to handle it at the `Page` route handler level is that we already fetch page data there so it should not have a perf impact and we can avoid a redirect (see current change) whilst writing the correct metadata/canonical. We can also use a redirect approach (see [this commit](https://github.com/GitbookIO/gitbook/pull/2376/commits/82781adc84185fdd5f2824c161d4e4276ea62de5)) but triggering a redirect at this level seem to translate the a html meta refresh instead of proper 307 redirects which is not great and could be a disadvantage I think.

Another approach I had thought was to handle this logic in `getPublishedContentByUrl` API handler but we don't fetch revision data there so didn't think we should do it there for perf reason and so discarded it.

There could be another level where we could handle this logic but couldn't think of any other ones sorry.

Comment (emmerich) [reply src/app/(space)/(content)/[[...pathname]]/page.tsx:1]: Approach 2 feels like the right way to do things.

I'm surprised the redirect approach didn't work with something like: 

```
if (!page && fallback) {
  redirect('/');
}
```

Comment (emmerich) [reply src/app/(space)/(content)/[[...pathname]]/page.tsx:1]: In fact, with this approach, aren't you going to redirect anyways? From line 57:

```
} else if (getPagePath(pages, page) !== rawPathname) {
  redirect(pageHref(pages, page, linksContext));
}
```

Because your new page path isn't the pathname you started with, you're going to redirect.

Comment (emmerich) [reply src/app/(space)/(content)/[[...pathname]]/page.tsx:1]: Feels like a redirect is even nicer because you can strip the query param off in the redirect too! Instead of doing a client-side query param cleanup

Comment (spastorelli) [reply src/app/(space)/(content)/[[...pathname]]/page.tsx:1]: > I'm surprised the redirect approach didn't work with something like:

It's not that it's not working it's just that it doesn't produce HTTP redirects like 307, at the page route handler level doing a `redirect(...)` translate to a HTML meta refresh which feels off. Almost like we load the html and then have to refresh it again instead of handling it at the HTTP layer. At least from what I noticed.

> In fact, with this approach, aren't you going to redirect anyways? From line 57:

Yes good point. Yeah I could also trigger the redirect at line 45:
https://github.com/GitbookIO/gitbook/pull/2376/files#diff-29898e446e07057bae008ab5942b02aca79c24fb31ecfb29d9b789a8fa6dfff5R45

 > actually thinking about it we'd still need to strip off the query param in the case where we find a page post-variant
 Exactly in the good scenario where the requested path exist I don't think we should redirect (as it's unecessary) and so we have to remove it client side so it doesn't stick.

Comment (spastorelli) [reply src/app/(space)/(content)/[[...pathname]]/page.tsx:1]: Ha that would explain the meta tag:

> When used in a [streaming context](https://nextjs.org/docs/app/building-your-application/routing/loading-ui-and-streaming#what-is-streaming), this will insert a meta tag to emit the redirect on the client side. When used in a server action, it will serve a 303

https://nextjs.org/docs/app/api-reference/functions/redirect

Comment (spastorelli) [reply src/app/(space)/(content)/[[...pathname]]/page.tsx:1]: @emmerich changed it to explicitly redirect when the page is not found and fallbacking PTAL.

Also please see my 2 comments above.

src/app/(space)/(content)/[[...pathname]]/page.tsx
```diff
@@ -30,11 +33,18 @@ export default async function Page(props: { params: PagePathParams }) {
         parent,
         customization,
         pages,
-        page,
+        page: resolvedPage,
```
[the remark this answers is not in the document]
Comment (spastorelli) [reply src/app/(space)/(content)/[[...pathname]]/page.tsx:36]: Yes, no point now. Sorry it was a residue from the impl. prio to me making [this commit](https://github.com/GitbookIO/gitbook/pull/2376/commits/b15c70fb850c19a3bce2872c0868270ef894b170), didn't clean up in the process.

src/app/(space)/(content)/[[...pathname]]/page.tsx
```diff
@@ -114,24 +124,38 @@ export async function generateViewport({ params }: { params: PagePathParams }):
...
-    if (!page) {
+    const canFallback = searchParams.fallback?.toLocaleLowerCase() === 'true';
+
+    let targetPage = page;
+    if (!page && canFallback) {
+        const rootPage = resolveFirstDocument(pages, []);
+        targetPage = rootPage?.page;
+    }
+
+    if (!targetPage) {
```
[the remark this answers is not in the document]
Comment (spastorelli) [reply src/app/(space)/(content)/[[...pathname]]/page.tsx:143]: Yes good point. Reverting this change.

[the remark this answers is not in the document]
Comment (spastorelli) [reply src/app/(space)/(content)/[[...pathname]]/page.tsx:143]: Actually this is needed otherwise during the first `fallback=true` request to the non-existing page, `generateMetadata` will short-circuit and return a Page not found here:
https://github.com/GitbookIO/gitbook/blob/81b372f372e52d59ad944451c03f7d33f8c303c1/src/app/(space)/(content)/%5B%5B...pathname%5D%5D/page.tsx#L119

And we still need to generate the metadata for that request I think hence why I use the fallback page for it. Yes they will be generated again anyway in the next request (redirect) but I think we still need to generate the metadata for that first request.

src/app/(space)/(content)/[[...pathname]]/PageClientLayout.tsx
```diff
@@ -1,13 +1,34 @@
 'use client';
 
+import { usePathname, useRouter, useSearchParams } from 'next/navigation';
+import React from 'react';
+
 import { useScrollToHash } from '@/components/hooks';
 
+function useStripFallbackQueryParam() {
```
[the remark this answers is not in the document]
Comment (spastorelli) [reply src/app/(space)/(content)/[[...pathname]]/PageClientLayout.tsx]: done.

src/app/(space)/(content)/[[...pathname]]/PageClientLayout.tsx
```diff
@@ -1,13 +1,34 @@
...
+
 /**
  * Client component to initialize interactivity for a page.
  */
 export function PageClientLayout(props: {}) {
     // We use this hook in the page layout to ensure the elements for the blocks
     // are rendered before we scroll to the hash.
     useScrollToHash();
+
+    useStripFallbackQueryParam();
```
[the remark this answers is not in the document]
Comment (spastorelli) [reply src/app/(space)/(content)/[[...pathname]]/PageClientLayout.tsx:20]: As chatted this is because in the case of a `fallback=true` request where the path match in the switched variant we are not redirecting which means we need to remove it client side.

src/app/(space)/(content)/[[...pathname]]/page.tsx
```diff
@@ -35,6 +38,12 @@ export default async function Page(props: { params: PagePathParams }) {
     } = await fetchPageData(params);
     const linksContext: PageHrefContext = {};
 
+    const canFallback = !!searchParams.fallback;
+    if (!page && canFallback) {
+        const rootPage = resolveFirstDocument(pages, []);
+        rootPage?.page ? redirect(pageHref(pages, rootPage?.page, linksContext)) : notFound();
```
Comment (SamyPesse) [inline src/app/(space)/(content)/[[...pathname]]/page.tsx:44]: Maybe we should not call the `notFound` here and let the logic below handle it.

```suggestion
        if (rootPage?.page) {
            redirect(pageHref(pages, rootPage?.page, linksContext));
        }
```

src/app/(space)/(content)/[[...pathname]]/page.tsx
```diff
@@ -114,24 +123,38 @@ export async function generateViewport({ params }: { params: PagePathParams }):
     };
 }
 
-export async function generateMetadata({ params }: { params: PagePathParams }): Promise<Metadata> {
+export async function generateMetadata({
```
Comment (SamyPesse) [inline src/app/(space)/(content)/[[...pathname]]/page.tsx:126]: I think the whole part here it not needed.

Comment (SamyPesse) [reply src/app/(space)/(content)/[[...pathname]]/page.tsx:126]: Nevermind I read the other comment. we should then try to reuse a function in between the 2 to avoid different logic growing over time.

Comment (spastorelli) [reply src/app/(space)/(content)/[[...pathname]]/page.tsx:124]: Done https://github.com/GitbookIO/gitbook/pull/2376/commits/871f9642adf94d37cae8be3fd50f36df8d2fa542

src/components/Header/SpacesDropdownMenuItem.tsx
```diff
@@ -10,6 +10,7 @@ function useVariantSpaceHref(variantSpaceUrl: string) {
     const targetUrl = new URL(variantSpaceUrl);
     targetUrl.pathname += `/${currentPathname}`;
     targetUrl.pathname = targetUrl.pathname.replace(/\/{2,}/g, '/').replace(/\/$/, '');
+    targetUrl.search = '?fallback=true';
```
SamyPesse, line 13:
```suggestion
    targetUrl.searchParams.set('fallback', 'true');
```

src/app/(space)/(content)/[[...pathname]]/PageClientLayout.tsx
```diff
@@ -9,5 +12,25 @@ export function PageClientLayout(props: {}) {
     // We use this hook in the page layout to ensure the elements for the blocks
     // are rendered before we scroll to the hash.
     useScrollToHash();
+
+    // When the user switches variants using the space dropdown, we pass a fallback=true parameter.
+    // This parameter indicates that we should redirect to the root page if the path from the
+    // previous variant doesn't exist in the new variant. If the path does exist, no redirect occurs,
+    // so we need to remove the fallback parameter.
```
Comment (SamyPesse) [inline src/app/(space)/(content)/[[...pathname]]/PageClientLayout.tsx:19]: Not important: It find it better to document/comment at the function definition level, as it can be visualized in VS code and also more accessible in cases of multiple uses.

SamyPesse (approved):

PR #2378, opened 2024-07-14
Fix render of one of when alternatives are primitives

SamyPesse:
## Before:
Screenshot 2024-07-14 at 20 51 03

## After:

Screenshot 2024-07-14 at 20 51 16

### DIFF
--- a/packages/react-openapi/src/OpenAPISchema.tsx
+++ b/packages/react-openapi/src/OpenAPISchema.tsx
@@ -182,6 +182,8 @@ export function OpenAPIRootSchema(props: {
 
 /**
  * Render a tab for an alternative schema.
+ * It renders directly the properties if relevant;
+ * for primitives, it renders the schema itself.
  */
 function OpenAPISchemaAlternative(props: {
     schema: OpenAPIV3.SchemaObject;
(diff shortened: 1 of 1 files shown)


packages/react-openapi/src/OpenAPISchema.tsx
```diff
@@ -190,12 +192,13 @@ function OpenAPISchemaAlternative(props: {
...
     const id = useId();
+    const subProperties = getSchemaProperties(schema);
 
     return (
         <OpenAPISchemaProperties
             id={id}
-            properties={getSchemaProperties(schema) ?? []}
-            circularRefs={new Map(circularRefs).set(schema, id)}
+            properties={subProperties ?? [{ schema }]}
+            circularRefs={subProperties ? new Map(circularRefs).set(schema, id) : circularRefs}
```
scazan, line 201: This feels smelly that we are calling new here directly in the prop. It would be cleaner to pull this up and out of the JSX I think.

scazan (approved):

packages/react-openapi/src/OpenAPISchema.tsx
SamyPesse: Why? it's done in a bunch of places here, it makes sure we create a new reference without impacting the other ones used in the react tree (map as an immutable map without using a library for this).

Pulling this up will make it even less clear imo as you'll lose track of where it's used and why we need a new.

scazan: The thinking is that it should be more clear because you'll need to name this logic. It's also fairly un-Reacty. 
this function here was named so was a bit easier to understand immediately: `getSchemaProperties(schema)`.
Maybe a style difference? not sure but it seems less readable for me as I have to stop and look closely at it to be sure it's not gonna cause an optimization issue.

PR #2380, opened 2024-07-16
Control tracking of page views with env/header variables

valentin0h:
Adds environment and header variables to skip tracking page views in cases/environments where we don't want to, such as previews of published content.


src/lib/seo.ts
```diff
@@ -28,6 +30,12 @@ export function shouldIndexSpace({
         return false;
     }
 
+    // Prevent indexation of gitbook instances other than `multi`
+    const gitbookMode = getGitbookMode();
+    if (gitbookMode !== 'multi') {
+        return false;
+    }
```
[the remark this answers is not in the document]
valentin0h: Hmm 🤔 While it'll be cleaner for this case, how are we going to handle similar cases where we need a different behaviour based on the instance `mode`? E.g., SEO indexing as you mentioned. More headers?

src/components/PageBody/PageBody.tsx
```diff
@@ -11,6 +11,7 @@ import { getSpaceLanguage } from '@/intl/server';
 import { t } from '@/intl/translate';
 import { ContentPointer, ContentTarget, SiteContentPointer, api } from '@/lib/api';
 import { hasFullWidthBlock, isNodeEmpty } from '@/lib/document';
+import { getGitbookMode } from '@/lib/middleware';
```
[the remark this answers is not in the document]
valentin0h: Ha, the naming is a bit mixed up; there's a `GITBOOK_MODE` and a `URLLookupMode`. I'll try to rework the naming of the "gitbook instance mode"

src/lib/seo.ts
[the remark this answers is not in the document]
valentin0h: Ha, okay, I jus saw 19-22. We already rely on separate environments/headers for each. Will adapt.

src/components/PageBody/PageBody.tsx
```diff
@@ -141,14 +141,12 @@ export function PageBody(props: {
                     ) : null}
                 </div>
             </main>
-            <React.Suspense fallback={null}>
```
valentin0h, line 144: Removed the Suspense as I didn't see why it was necessary. Let me know if I missed something.

src/components/PageBody/TrackPageViewConditional.tsx
```diff
@@ -0,0 +1,31 @@
...
+    sitePointer?: Pick<SiteContentPointer, 'siteId' | 'organizationId'>;
+    spaceId: string;
+    pageId: string | undefined;
+}) {
+    const { apiHost, sitePointer, spaceId, pageId } = props;
+
+    const headerSet = headers();
+    if (
+        process.env.GITBOOK_BLOCK_PAGE_VIEWS_TRACKING &&
+        !headerSet.has('x-gitbook-track-page-views')
```
valentin0h, line 18: This makes it on by default. I'm conflicted if that's best.

SamyPesse: I think it should be on by default 👍  just like SEO indexation

src/components/PageBody/server-actions.ts
SamyPesse, line 1: Why are you doing this change? this is a BIG change that will cause a LOT more requests: each page view will require a Cloudflare worker execution to proxy it to the API.

Also, it means the IP address will be the wrong one, as in that case, it'll be the IP of the Cloudflare datacenter.

src/components/PageBody/PageBody.tsx
SamyPesse: It should be fine, I'm not sure if Next suspends when loading a client component

src/components/PageBody/server-actions.ts
valentin0h: Ah, okay. I didn't understand the impact very well. Reverting.

src/components/PageBody/TrackPageViewConditional.tsx
```diff
@@ -0,0 +1,34 @@
+import { headers } from 'next/headers';
+
+import { SiteContentPointer } from '@/lib/api';
+
+import { TrackPageView } from './TrackPageView';
+
+/**
+ * Track page view in GitBook analytics if not disabled
+ */
+export function TrackPageViewConditional(props: {
```
SamyPesse, line 10: It seems a bit too much to have a component just for this instead of javing the condition in a function in `PageBody.tsx`, it's a bit of an indirection

SamyPesse (approved):

PR #2381, opened 2024-07-16
Show additional context when linking to a page group & a page in another space

spastorelli:
This PR aims to change how we display content ref link cards for:

### Page group

e.g: https://docs.vitally.io

| **Before**    | **After** |
| -------- | ------- |
| We used to show just the title of the first page in the group.  | We now show the page group title instead.    |
| Screenshot 2024-07-16 at 18 01 06  | Screenshot 2024-07-16 at 18 00 47    |

### Page in another Space

e.g: https://docs.rekor.ai

| **Before**    | **After** |
| -------- | ------- |
| We used to show only the title of the page without more context  | We now show the card with the space title and page title.   |
|Screenshot 2024-07-16 at 18 03 53  | Screenshot 2024-07-16 at 18 05 20 |

also fixes RND-4521


src/lib/pages.ts
```diff
@@ -141,7 +141,15 @@ function resolvePageDocument(
...
-            return firstDocument;
+            const { page: firstPageInGroup, ancestors } = firstDocument;
+            return {
+                // Keep the page group title but use the first page in the group for the rest.
+                page: {
+                    ...firstPageInGroup,
+                    title: page.title,
+                },
+                ancestors,
+            };
```
SamyPesse, line 152: It seems a bit risky to transform data on the fly in a low function like this one that is not used only for resolving content refs.

Do we have alternatives?

src/lib/references.ts
```diff
@@ -245,12 +247,22 @@ async function resolveContentRefInSpace(spaceId: string, contentRef: ContentRef)
...
     });
+
+    if (!resolved) {
+        return null;
+    }
+
+    return {
+        ...resolved,
+        text: space.title,
+        subText: resolved.text,
```
SamyPesse, line 264: Why not the reverse? using `text: space.title` works well hen you point to a first page where the title doesn't matter, but it's bit worse when pointin to another page.

```suggestion
        subText: space.title,
```

src/lib/pages.ts
spastorelli: Good point. I've changed it to handle it in resolveContentRef:
https://github.com/GitbookIO/gitbook/pull/2381/commits/ce569ca218165f8a16a8ae98602994e4a3325f48

Let me know what you think.

src/components/DocumentView/BlockContentRef.tsx
```diff
@@ -18,8 +18,12 @@ export async function BlockContentRef(props: BlockProps<DocumentBlockContentRef>
         return null;
     }
 
+    const isContentInOtherSpace =
+        context.contentRefContext?.space &&
+        'space' in block.data.ref &&
+        context.contentRefContext.space.id !== block.data.ref.space;
     const kind = block?.data?.ref?.kind;
-    if (resolved.active && kind === 'space') {
+    if ((resolved.active && kind === 'space') || isContentInOtherSpace) {
```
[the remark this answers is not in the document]
spastorelli: Because in cases it's a page content ref in another site and we want to show it as a Space card with space icon/emoji....

PR #2382, opened 2024-07-17
Update Site search endpoint

scazan:
This updates the endpoint used for sites to enable specific multi-space search support as well as proper analytics per site-space as opposed to per space (which can be used in multiple sites).

It also centralizes the logic into a function that can replace both `searchSpaceContent` and `searchParentContent` in the future.

### DIFF
--- a/src/components/Search/server-actions.tsx
+++ b/src/components/Search/server-actions.tsx
@@ -54,9 +54,9 @@ export interface AskAnswerResult {
 export async function searchSiteContent(args: {
     query: string;
     siteSpaceIds?: string[];
-    cacheKey?: string;
+    cacheBust?: string;
 }): Promise<OrderedComputedResult[]> {
-    const { siteSpaceIds = [], query, cacheKey } = args;
+    const { siteSpaceIds = [], query, cacheBust } = args;
     const pointer = getContentPointer();
 
     if ('siteId' in pointer && 'organizationId' in pointer) {
@@ -65,7 +65,7 @@ export async function searchSiteContent(args: {
             pointer.siteId,
             query,
             siteSpaceIds,
-            cacheKey,
+            cacheBust,
         );
 
         // resolve all SiteSpaces so we can match up with the spaceId
@@ -124,7 +124,7 @@ export async function searchSpaceContent(
 
         // This is a site so use a different function which we can eventually call directly
         // We also want to break cache for this specific space if the revisionId is different so use it as a cache busting key
-        return await searchSiteContent({ siteSpaceIds, query, cacheKey: revisionId });
+        return await searchSiteContent({ siteSpaceIds, query, cacheBust: revisionId });
     }
 
     const data = await api.searchSpaceContent(spaceId, revisionId, query);
--- a/src/components/Search/server-actions.tsx
+++ b/src/components/Search/server-actions.tsx
@@ -55,6 +55,7 @@ export interface AskAnswerResult {
 export async function searchSiteContent(
     siteSpaceIds: string[],
     query: string,
+    cacheKey: string,
 ): Promise<OrderedComputedResult[]> {
     const pointer = getContentPointer();
 
@@ -65,6 +66,7 @@ export async function searchSiteContent(
             pointer.siteId,
             query,
             siteSpaceIds,
+            cacheKey,
         );
 
         // resolve all SiteSpaces so we can match up with the spaceId
@@ -99,7 +101,8 @@ export async function searchSpaceContent(
         const siteSpaceIds = pointer.siteSpaceId ? [pointer.siteSpaceId] : []; // if we don't have a siteSpaceID search all content
 
         // This is a site so use a different endpoint
-        return await searchSiteContent(siteSpaceIds, query);
+        // We also want to break cache for this specific space if the revisionId is different so use it as a cache key
+        return await searchSiteContent(siteSpaceIds, query, revisionId);
     }
 
     const data = await api.searchSpaceContent(spaceId, revisionId, query);
--- a/src/components/Search/server-actions.tsx
+++ b/src/components/Search/server-actions.tsx
@@ -6,6 +6,7 @@ import {
     SearchAIAnswer,
     SearchPageResult,
     Site,
+    SiteSpace,
     Space,
 } from '@gitbook/api';
 
@@ -82,23 +83,13 @@ export async function searchSiteContent(args: {
 
         if (!siteSpaceIds) {
             // We are searching all of this Site's content
-
-            const processedSiteSpaces = allSiteSpaces.map((siteSpace) => {
-                // replace the published url for the "space" with the site's published url
-                siteSpace.space.urls.published = siteSpace.urls.published;
-                return siteSpace;
-            });
-
-            // We pass a space to the transform since this search is relative to the Site
             return searchResults.items
                 .map((spaceItem) => {
-                    const siteSpace = processedSiteSpaces.find(
+                    const siteSpace = allSiteSpaces.find(
                         (siteSpace) => siteSpace.space.id === spaceItem.id,
                     );
 
-                    return spaceItem.pages.map((item) =>
-                        transformPageResult(item, siteSpace?.space),
-                    );
+                    return spaceItem.pages.map((item) => transformSitePageResult(item, siteSpace));
                 })
                 .flat(2);
         }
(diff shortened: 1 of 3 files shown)


src/components/Search/server-actions.tsx
```diff
@@ -52,6 +51,64 @@ export interface AskAnswerResult {
...
+    const { siteSpaceIds = [], query, cacheKey } = args;
+    const pointer = getContentPointer();
+
+    if ('siteId' in pointer && 'organizationId' in pointer) {
+        const searchResults = await api.searchSiteContent(
+            pointer.organizationId,
+            pointer.siteId,
+            query,
+            siteSpaceIds,
+            cacheKey,
```
SamyPesse, line 68: What is `cacheKey`?

src/components/Search/server-actions.tsx
```diff
@@ -52,6 +51,64 @@ export interface AskAnswerResult {
...
+}): Promise<OrderedComputedResult[]> {
+    const { siteSpaceIds = [], query, cacheKey } = args;
+    const pointer = getContentPointer();
+
+    if ('siteId' in pointer && 'organizationId' in pointer) {
+        const searchResults = await api.searchSiteContent(
+            pointer.organizationId,
+            pointer.siteId,
+            query,
+            siteSpaceIds,
```
SamyPesse, line 67: Why do we need to pass the space IDs since the API backend should know them?

src/components/Search/server-actions.tsx
```diff
@@ -52,6 +51,64 @@ export interface AskAnswerResult {
...
+            cacheKey,
+        );
+
+        // resolve all SiteSpaces so we can match up with the spaceId
+        const allSiteSpaces = await api.getSiteSpaces({
```
SamyPesse, line 72: Can we do this in parallel of the first call to get a response faster?

src/components/Search/server-actions.tsx
```diff
@@ -52,6 +51,64 @@ export interface AskAnswerResult {
...
+            siteId: pointer.siteId,
+            siteShareKey: pointer.siteShareKey,
+        });
+
+        if (siteSpaceIds.length === 0) {
+            // We are searching all of this Site's content
+
+            const processedSiteSpaces = allSiteSpaces.map((siteSpace) => {
+                // replace the published url for the "space" with the site's published url
+                siteSpace.space.urls.published = siteSpace.urls.published;
```
SamyPesse, line 83: Why?

src/lib/api.ts
```diff
@@ -1046,6 +1046,41 @@ export const searchParentContent = cache(
...
+ */
+export const searchSiteContent = cache(
+    'api.searchSiteContent',
+    async (
+        organizationId: string,
+        siteId: string,
+        query: string,
+        siteSpaceIds: string[] = [],
+        /** A cache bust key to avoid revalidating lot of cache entries by tags */
+        cacheKey?: string,
```
SamyPesse, line 1060: Can we then name it `cacheBust`? it's not actually the cache key here

src/lib/api.ts
```diff
@@ -1046,6 +1046,41 @@ export const searchParentContent = cache(
...
+        options?: CacheFunctionOptions,
+    ) => {
+        const response = await api().orgs.searchSiteContent(
+            organizationId,
+            siteId,
+            {
+                query,
+                siteSpaceIds,
```
SamyPesse, line 1068: I didn't review the API design of this endpoint, but I have mixed feeling at first sight. What does it mean when an empty array is passed? Why not pass `undefined`? Is it a filter?

SamyPesse (commented):
Looks good, but I have a few questions :)

src/components/Search/server-actions.tsx
SamyPesse: Answered below, but we should change the name to `cacheBust`

src/components/Search/server-actions.tsx
SamyPesse: Understanding the rest, I think we should have it typed as `siteSpaceIds: string[] | undefined`; the case `[]` to indicate all is weird

src/lib/api.ts
```diff
@@ -1046,6 +1046,41 @@ export const searchParentContent = cache(
...
+            },
+            undefined,
+            {
+                ...noCacheFetchOptions,
+                signal: options?.signal,
+            },
+        );
+
+        return cacheResponse(response, {
+            ttl: 60 * 60,
```
jpreynat, line 1078: 1h seems a lot for search results. We should make it shorter, like 5 minutes top.

src/components/Search/server-actions.tsx
scazan: This isn't all siteSpaceIds.. it's just the spaceIds that we want to search in.

src/lib/api.ts
jpreynat: From the API spec, it looks like a filter yeah, so we should simply not pass it when searching across all.

src/components/Search/server-actions.tsx
scazan: Basically this adds a new feature that is not being used.. we can select specific spaces using this endpoint rather than one or all.

src/lib/api.ts
```diff
@@ -1046,6 +1046,41 @@ export const searchParentContent = cache(
...
+/**
+ * Search content in a Site or specific SiteSpaces.
+ */
+export const searchSiteContent = cache(
+    'api.searchSiteContent',
+    async (
+        organizationId: string,
+        siteId: string,
+        query: string,
+        siteSpaceIds: string[] = [],
```
jpreynat, line 1058: Given that it's a filter, we should rename it and have it either be filled or `undefined` indeed.

src/lib/api.ts
```diff
@@ -1046,6 +1046,41 @@ export const searchParentContent = cache(
...
+        /** A cache bust param to avoid revalidating lot of cache entries by tags */
+        cacheBust?: string,
```
jpreynat, line 1060: It doesn't look like this is used anywhere in the code?

src/lib/api.ts
SamyPesse: There is the cache bust now, so 1h is fine, 5min will not even work with KV

src/lib/api.ts
SamyPesse: It's not used, but passed; which is then computed in the cacheKey

src/components/Search/server-actions.tsx
scazan: This is code ported over but it would seem we do not clear out the old published url for a space when migrating to sites. Usually it's the same, but since a space can belong to multiple sites it is always the site URL that we want. To support old published spaces and SiteSpaces it seems like this was done so they can coexist.

src/lib/api.ts
scazan: yeah, it's a filter. I'll update the API.

scazan: ah, I already did that in the API as you can pass undefined for the same thing. just didn't do that here. fixing...

src/lib/api.ts
jpreynat: Sorry but I can't see where this is even passed?
https://github.com/GitbookIO/gitbook/blob/47fbf27ff2a25d1a0dbc12b26ed63f838b510b85/src/lib/api.ts#L1052-L1082

src/lib/api.ts
jpreynat: If it's covered by the cache bust then yeah.

src/components/Search/server-actions.tsx
jpreynat: It looks like it's due to `transformPageResult` accepting only a space for now, so we're looking for the `space.urls.published` value:
https://github.com/GitbookIO/gitbook/blob/aa8bbaddc46aa90ef30c970a057b329cd7a694ec/src/components/Search/server-actions.tsx#L186-L191

But we should adapt that for sites or even create a new transform function for them instead of manipulating the data to match the implementation for spaces I think.

src/lib/api.ts
scazan: Do you mean how it's passed down from the frontend to the API or do you mean how it acts as a cache breaking arg? The latter is here: https://github.com/GitbookIO/gitbook/blob/main/src/lib/cache/cache.ts#L214

src/components/Search/server-actions.tsx
scazan: ok, made an update that pulls the space-specific logic into a function and common functionality into another.

src/components/Search/server-actions.tsx
```diff
@@ -52,6 +52,59 @@ export interface AskAnswerResult {
...
+    if ('siteId' in pointer && 'organizationId' in pointer) {
+        const [searchResults, allSiteSpaces] = await Promise.all([
+            api.searchSiteContent(
+                pointer.organizationId,
+                pointer.siteId,
+                query,
+                siteSpaceIds,
+                cacheBust,
+            ),
+            api.getSiteSpaces({
```
SamyPesse, line 77: It looks like it's used only when `!siteSpaceIds` later, so we can skip it here in the other case:

```suggestion
            siteSpaceIds ? null : api.getSiteSpaces({
```

SamyPesse (approved):
Looks good!

PR #2383, opened 2024-07-17
Hide pages from table of contents when set as hidden

SamyPesse:


taranvohra (approved):

src/lib/pages.ts
```diff
@@ -75,12 +75,13 @@ export function resolvePageId(
 
 /**
  * Resolve the next/previous page before another one.
+ * It ignores hidden pages as this used for navigation purpose.
```
scazan, line 78:
```suggestion
 * It ignores hidden pages as this is used for navigation purpose.
```

scazan (approved):

PR #2388, opened 2024-07-19
Skip hidden pages in sitemap.xml

SamyPesse:


spastorelli (approved):
Should these hidden pages have their metadata include a noindex to avoid to be crawled directly? I wondering if excluding them from the sitemap is enough for it not to be indexed.

gregberge (approved):

PR #2405, opened 2024-07-24
Add an nvmrc file

emmerich:


SamyPesse:
Why? it doesn't use npm/nvm, but it uses `bun`

emmerich:
> Why? it doesn't use npm/nvm, but it uses `bun`

It's still required to run locally:
Screenshot 2024-07-25 at 11 33 18

spastorelli (approved):

PR #2409, opened 2024-07-26
Pass shareKey as context while fetching space

taranvohra:
To make sure the `space.urls.published` are properly resolved for share-link sites. Otherwise navigation between references etc start breaking.

Also the code might seem messy but all of this will be properly refactored when we are fully on sites and can get rid of `ContentPointer` ternaries from places.


packages/gitbook/src/middleware.ts
```diff
@@ -172,11 +172,14 @@ export async function middleware(request: NextRequest) {
...
-                    changeRequestId: resolved.changeRequest,
-                    revisionId: resolved.revision,
-                }),
+                getSpaceContentData(
+                    {
+                        spaceId: resolved.space,
+                        changeRequestId: resolved.changeRequest,
+                        revisionId: resolved.revision,
+                    },
+                    'site' in resolved ? resolved.shareKey : undefined,
```
jpreynat, line 181: I'm wondering if we shouldn't have a helper for that, given how many times it's done..!

jpreynat (approved):
Looking good

packages/gitbook/src/middleware.ts
taranvohra: I agree but all of this will be refactored and cleaned up soon when sites are 100% GA

PR #2417, opened 2024-08-02
Fix running repository without access to Pro token for @gitbook/icons

SamyPesse:


package.json
```diff
@@ -15,7 +15,8 @@
         "unit": "turbo run unit",
         "e2e": "turbo run e2e",
         "changeset": "changeset",
-        "release": "turbo run release && changeset publish"
+        "release": "turbo run release && changeset publish",
+        "download:env": "op read op://gitbook-x-dev/gitbook-open/.env.local > .env.local"
```
valentin0h, line 19:
```suggestion
        "download:env": "op read op://gitbook-x-dev/gitbook-open/.env.local >> .env.local"
```

SamyPesse: It means it could only be run once 🤔  I don't know if it's better but ok

valentin0h (approved):

PR #2418, opened 2024-08-02
Fix preview of customization

SamyPesse:


emmerich (approved):
you do something like an `as` cast to avoid the ts-ignore (it would've cause this error)

PR #2425, opened 2024-08-09
Automatically scroll to the active page in table of contents

spastorelli:
**Before**
e.g: https://docs.gitbook.com/account-management/sso-and-saml/sso-and-saml

https://github.com/user-attachments/assets/967ae1f8-28c2-4e5f-8a23-c611c07b75f2

**After**

https://github.com/user-attachments/assets/7a621e99-ed09-4002-bf15-0212238a27ee


packages/gitbook/src/components/TableOfContents/useScrollToActiveTOCItem.tsx
```diff
@@ -0,0 +1,55 @@
...
+import React from 'react';
+
+const TOCScrollContainerContext = React.createContext<string | null>(null);
+
+const SCROLL_OFFSET = 200;
+
+/**
+ * Provider that supplies the ID of the scroll container for the table of contents.
+ */
+export function TOCScrollContainerProvider(
```
SamyPesse, line 12: Instead of just rendering the react context, you can extend it to: create a ref, render the div, render the contxt provider with the ref as value.

It'll avoid the use of an ID for it.

packages/gitbook/src/components/TableOfContents/useScrollToActiveTOCItem.tsx
```diff
@@ -0,0 +1,55 @@
...
+                const containerBottom = containerTop + container.clientHeight;
+
+                // Only scroll if the TOC item is outside the viewable area of the container
+                if (itemTop < containerTop || itemBottom > containerBottom) {
+                    container.scrollTo({
+                        top: itemTop - SCROLL_OFFSET,
+                        behavior: 'smooth',
+                    });
+                }
+            }
```
SamyPesse, line 52: Have you tried using `scrollIntoView` / `scrollIntoViewIfNeeded` https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView ?

SamyPesse: When I think about it, I had a similar thing in the app last week. And you might not need to know the parent scroll div as `scrollIntoView` will automaticaly detect it, so maybe the link ref is enough (to be tested)

spastorelli: Yeah I tried `scrollIntoView` using just the `linkRef` but that would not work for some reason. I think that because in that case the scrolling container is not the immediate ancestor. That's why I ended up with the solution of getting the element of the actual scroll container and trigger the scroll to the specific position.

But I'll give it another try maybe missing something.

packages/gitbook/src/components/TableOfContents/useScrollToActiveTOCItem.tsx
spastorelli: Yeah I had tried with passing a ref via the provider but that would not work because of the TableOfContents component is server side. But maybe because I was creating it directly in the TOC component instead of doing it in the Provider itself. I'll try what you suggested.

packages/gitbook/src/components/TableOfContents/useScrollToActiveTOCItem.tsx
spastorelli: Ok not sure what I was doing wrong but works when I retried 😅. Made the change. Thanks

spastorelli: @SamyPesse Ok so reverted back to using scrollTo in the end as I was hitting when using scrollIntoView specifically that depending on how far down the TOC item was it could potentially also scroll the page content too. I think this has to do with the behavior described [in this issue](https://github.com/w3c/csswg-drafts/issues/9452) which is that scrollIntoView will run the scroll against each ancestor element and because in our case page content and TOC are sharing one.

Let me know what you think. I would be happier to use the simpler scrollIntoView but couldn't find a way to make it work without hitting the issue above.

packages/gitbook/src/components/TableOfContents/useScrollToActiveTOCItem.tsx
```diff
@@ -0,0 +1,84 @@
...
+    props: React.PropsWithChildren<{ withHeaderOffset: boolean; withTrademarkEnabled: boolean }>,
+) {
+    const { withHeaderOffset, withTrademarkEnabled, children } = props;
+    const scrollContainerRef = React.createRef<HTMLDivElement>();
+
+    return (
+        <TOCScrollContainerContext.Provider value={scrollContainerRef}>
+            <div
+                ref={scrollContainerRef}
+                className={tcls(
```
SamyPesse, line 19: Instead of passing the props here and computing the `className` here, you can pass the className as a prop or even accept `props: React.HTMLAttributes<HTMLDivElement>` and pass them all to the div

packages/gitbook/src/components/TableOfContents/useScrollToActiveTOCItem.tsx
```diff
@@ -0,0 +1,84 @@
...
+                const containerBottom = containerTop + tocContainer.clientHeight;
+
+                // Only scroll if the TOC item is outside the viewable area of the container
+                if (
+                    tocItemTop < containerTop + TOC_ITEM_OFFSET ||
+                    tocItemTop > containerBottom - TOC_ITEM_OFFSET
+                ) {
+                    tocContainer.scrollTo({
+                        top: tocItemTop - TOC_ITEM_OFFSET,
+                        behavior: 'smooth',
```
SamyPesse, line 78: Is it best to not have a smooth scroll? when looking at the video, it distracts from the content, maybe we should just position the scroll as soon as possible.

packages/gitbook/src/components/TableOfContents/useScrollToActiveTOCItem.tsx
```diff
@@ -0,0 +1,84 @@
...
+ * but only if the item is outside the viewable area of the container.
+ */
+export function useScrollToActiveTOCItem(tocItem: {
+    isActive: boolean;
+    linkRef: React.RefObject<HTMLAnchorElement>;
+}) {
+    const { isActive, linkRef } = tocItem;
+
+    const scrollContainerRef = React.useContext(TOCScrollContainerContext);
+    React.useEffect(() => {
```
SamyPesse, line 61: Should we use a `useLayoutEffect` here? combined with no `behaviour`, it should avoid a flash of the toc with the wrong scroll position

packages/gitbook/src/components/TableOfContents/useScrollToActiveTOCItem.tsx
```diff
@@ -0,0 +1,84 @@
...
+                )}
+            >
+                {children}
+            </div>
+        </TOCScrollContainerContext.Provider>
+    );
+}
+
+// Offset to scroll the table of contents item by.
+const TOC_ITEM_OFFSET = 200;
```
SamyPesse, line 48: What does this offset represent? why 200?

SamyPesse (approved):
Looks good! minor comments on my side

packages/gitbook/src/components/TableOfContents/useScrollToActiveTOCItem.tsx
spastorelli: Yeah I wasn't sure about it. Went for smooth but indeed could be distracting. Changing it to instant.

packages/gitbook/src/components/TableOfContents/useScrollToActiveTOCItem.tsx
spastorelli: It's an offset to position the scrolled item away from the very top of the scroll container. I choose 200 abritrarily.

PR #2427, opened 2024-08-13
Fix mermaid / contentkit webframe not updating properly

SamyPesse:


packages/react-contentkit/src/ElementWebframe.tsx
```diff
@@ -61,7 +62,7 @@ export function ElementWebframe(props: ContentKitClientElementProps<ContentKitWe
 
             // For security reasons, only iframe from our integrations domains are allowed
             // to send and receive messages
-            if (!renderer.security.firstPartyDomains.includes(origin.host) && 0) {
+            if (!renderer.security.firstPartyDomains.includes(origin.host)) {
```
BrettJephson, line 65: Glad you removed that `&& 0` because I'm not sure what that was doing.

BrettJephson (approved):
LGTM

packages/react-contentkit/src/ElementWebframe.tsx
SamyPesse: It looks like it's my mistake, probably while developing and forgot to remove it 😬

PR #2434, opened 2024-08-19
RND-3532: drop down menu for hidden links at small screen size

BrettJephson:
Adds a `HeaderLinkMore` button when links are hidden which shows a drop down menu with the hidden links.

This PR extends the container query based css solution so the components don't need to 'use client'.

Screenshot 2024-08-19 at 09 49 17

Screenshot 2024-08-19 at 09 49 38

### DIFF
--- a/packages/gitbook/src/components/Header/Dropdown.tsx
+++ b/packages/gitbook/src/components/Header/Dropdown.tsx
@@ -1,7 +1,7 @@
 import { Icon } from '@gitbook/icons';
 import { DetailedHTMLProps, HTMLAttributes, useId } from 'react';
 
-import { tcls } from '@/lib/tailwind';
+import { ClassValue, tcls } from '@/lib/tailwind';
 
 import { Link } from '../primitives';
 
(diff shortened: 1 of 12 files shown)


packages/gitbook/src/components/Header/Dropdown.tsx
```diff
@@ -18,8 +18,10 @@ export function Dropdown<E extends HTMLElement>(props: {
     button: (buttonProps: DropdownButtonProps<E>) => React.ReactNode;
     /** Content of the dropdown */
     children: React.ReactNode;
+    /** Custom styles */
+    className?: string;
```
SamyPesse, line 22: We use the name `styles?: ClassValue` in this repo (I know the name is terrible).

Maybe we should start switching to `className`, but you can do:

```suggestion
    className?: ClassValue;
```

packages/gitbook/src/components/Header/HeaderLinkMore.tsx
```diff
@@ -0,0 +1,46 @@
+import { CustomizationHeaderLink } from '@gitbook/api';
+import React from 'react';
+
+import { ContentRefContext, resolveContentRef } from '@/lib/references';
+
+import { Dropdown, DropdownChevron, DropdownMenu, DropdownMenuItem } from './Dropdown';
+import styles from './headerLinks.module.css';
+
+export function HeaderLinkMore(props: {
```
SamyPesse, line 9: Add a comment please to explain the goal. It'll make this code more welcome to contributions

SamyPesse (approved):
Looks good!

PR #2435, opened 2024-08-19
RND-3113: expand when expandable anchor 

BrettJephson:
With this PR, anchor ids for expandable sections and also the content within expandable sections are now matched to the url hash. 

If there's a match, expand the expandable and scroll into view.

### DIFF
--- a/packages/gitbook/src/components/DocumentView/Expandable/Details.tsx
+++ b/packages/gitbook/src/components/DocumentView/Expandable/Details.tsx
@@ -1,44 +1,55 @@
 'use client'
 
-import { useParams } from 'next/navigation';
 import React from 'react';
 
 import { ClassValue, tcls } from '@/lib/tailwind';
+import { useParams } from 'next/navigation';
 
-
+function useHash() {
+    const params = useParams();
+    const [hash, setHash] = React.useState<string>(global.location?.hash?.slice(1));
+    React.useEffect(() => {
+        function updateHash() {
+            setHash( global.location?.hash?.slice(1));
+        }
+        global.addEventListener('hashchange', updateHash);
+        updateHash();
+        return () => global.removeEventListener('hashchange', updateHash); 
+    }, [params]);
+    return hash;
+}
 /**
  * Details component rendered on client so it can expand dependent on url hash changes.
  */
 export function Details(props: { children : React.ReactNode; id: string; contentIds?: string[]; open?: boolean; className?: ClassValue; }) {
-    const { children, id, contentIds = [], open, className } = props;
-    const params = useParams();
-    const [hash, setHash] = React.useState<string | null>(null);
+    const { children, id, open, className } = props;
     
+    const detailsRef = React.useRef<HTMLDetailsElement>(null);
+
+    const [anchorElement, setAnchorElement] = React.useState<Element | null | undefined>();
+
+    const hash = useHash();
     React.useEffect(() => {
-        function updateHash() {
-            const urlHash = global?.location?.hash;
-            if (urlHash !== hash) {
-                setHash(urlHash.slice(1));
-                const element = document.getElementById(urlHash.slice(1));
-                if (element) {
-                    element.scrollIntoView({
-                        block: 'start',
-                        behavior: 'smooth',
-                    });
-                }
-            }
+        if (!hash) { return; }
+        const descendant = hash === id ? detailsRef.current : detailsRef.current?.querySelector(`#${hash}`);
+        setAnchorElement(descendant);
+    }, [hash, id]);
+
+
+    React.useLayoutEffect(() => {
+        if (anchorElement) {
+            anchorElement.scrollIntoView({
+                block: 'start',
+                behavior: 'smooth',
+            });
         }
-        addEventListener('hashchange', updateHash);
-        updateHash();
-        return () => removeEventListener('hashchange', updateHash); 
-    }, [params, hash, open])
-    
-    const isOpen = open || [id, ...contentIds].some(id => hash === id);
+    }, [anchorElement]);
 
     return (
         <details
+            ref={detailsRef}
             id={id}
-            open={isOpen}
+            open={open || Boolean(anchorElement)}
             className={tcls(
                 className,
                 'group/expandable',
--- a/packages/gitbook/src/components/DocumentView/Expandable/Details.tsx
+++ b/packages/gitbook/src/components/DocumentView/Expandable/Details.tsx
@@ -1,4 +1,4 @@
-'use client'
+'use client';
 
 import React from 'react';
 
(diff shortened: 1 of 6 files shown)


packages/gitbook/src/components/DocumentView/Expandable/Details.tsx
```diff
@@ -0,0 +1,78 @@
...
+                        block: 'start',
+                        behavior: 'smooth',
+                    });
+                }
+            }
+        }
+        addEventListener('hashchange', updateHash);
+        updateHash();
+        return () => removeEventListener('hashchange', updateHash); 
+    }, [params, hash, open])
```
SamyPesse, line 34: Is there no hook from NPM or Next we can use here?

It looks like we'll need this in a bunch of places

packages/gitbook/src/components/DocumentView/Expandable/Details.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { useParams } from 'next/navigation';
+import React from 'react';
+
+import { ClassValue, tcls } from '@/lib/tailwind';
+
+
+/**
+ * Details component rendered on client so it can expand dependent on url hash changes.
+ */
+export function Details(props: { children : React.ReactNode; id: string; contentIds?: string[]; open?: boolean; className?: ClassValue; }) {
```
SamyPesse, line 12: Instead of accumulating the `contentIds` and potentially build a big array that needs to be passed client side.

I think the logic should be:

```
const hash = useHash()

React.useEffect(() => {
  if (hash) {
     const descendant = document.querySelectorAll(`#${hash}`); // carefully you might need to handle encoding a better approach here ex with getById
     if (descendant) {
        // open it
     }
  }
}, [hash]);

```

SamyPesse: Another approach could be to have one single hook in the layout that everytime the hash changes, lookup all the ancestors of the DOM element and set open to them if they are `<detail>`, it'll not be react-ish but more DOM-ish API

packages/gitbook/src/components/DocumentView/Expandable/Expandable.tsx
```diff
@@ -1,13 +1,28 @@
...
 
+import { Details } from './Details';
 import { BlockProps } from '../Block';
 import { Blocks } from '../Blocks';
 import { Inlines } from '../Inlines';
 
+/**
+ * Find nodes with ids within Expandable section's content.
+ */
+function getIdsInNodes(nodes: DocumentBlocksEssentials[]): string[] {
```
SamyPesse, line 15: See my comment about this approach being expensive

packages/gitbook/src/components/DocumentView/Expandable/Details.tsx
BrettJephson: Thanks for the pointers! I definitely wasn't very confident in my solution performance wise. Will take another look.

packages/gitbook/src/components/DocumentView/Expandable/Details.tsx
```diff
@@ -0,0 +1,98 @@
...
+    const params = useParams();
+    const [hash, setHash] = React.useState<string>(global.location?.hash?.slice(1));
+    React.useEffect(() => {
+        function updateHash() {
+            setHash(global.location?.hash?.slice(1));
+        }
+        global.addEventListener('hashchange', updateHash);
+        updateHash();
+        return () => global.removeEventListener('hashchange', updateHash);
+    }, [params]);
```
SamyPesse, line 18: Why do we need to have `params` as dependency here since it's not used? If it's for a specific reason, can you please comment it so that we don't forget?

packages/gitbook/src/components/DocumentView/Expandable/Details.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+    const [anchorElement, setAnchorElement] = React.useState<Element | null | undefined>();
+
+    const hash = useHash();
+    React.useEffect(() => {
+        if (!hash) {
+            return;
+        }
+        const descendant =
+            hash === id ? detailsRef.current : detailsRef.current?.querySelector(`#${hash}`);
```
SamyPesse, line 44:
```suggestion
            hash === id ? detailsRef.current : detailsRef.current?.getElementById(hash);
```

I had issues in the past with using hash in query selectors as it needs to be properly encoded.

packages/gitbook/src/components/DocumentView/Expandable/Details.tsx
```diff
@@ -0,0 +1,98 @@
...
+        setAnchorElement(descendant);
+    }, [hash, id]);
+
+    React.useLayoutEffect(() => {
+        if (anchorElement) {
+            anchorElement.scrollIntoView({
```
SamyPesse, line 50: I just noticed we also have `useScrollToHash`, I'm slightly concerned of the conflict between the 2 into trying to scroll.

packages/gitbook/src/components/DocumentView/Expandable/Details.tsx
SamyPesse: Looks like it's documented in useScrollToHash, we should merge the 2

BrettJephson: Thanks. Looks like I can get it to work with `useScrollToHash` with some tweaks to that hook. Then I just need to check for the active element in Details to know to set the open attribute.

packages/gitbook/src/components/DocumentView/Expandable/Details.tsx
BrettJephson: `getElementById` is only on the document but can combine it with contains for the context of the element.

packages/gitbook/src/components/DocumentView/Expandable/Details.tsx
BrettJephson: I've separated out the useHash hook. I couldn't find anything that did this exactly, with the params dependency.

packages/gitbook/src/components/DocumentView/Expandable/Details.tsx
```diff
@@ -0,0 +1,84 @@
...
+    id: string;
+    contentIds?: string[];
+    open?: boolean;
+    className?: ClassValue;
+}) {
+    const { children, id, open, className } = props;
+
+    const detailsRef = React.useRef<HTMLDetailsElement>(null);
+
+    const [anchorElement, setAnchorElement] = React.useState<Element | null | undefined>();
```
SamyPesse, line 22: Useless but now you can have only `const [open, setOpen] = React.useState(false);`

SamyPesse (approved):

PR #2440, opened 2024-08-27
Filter out short searches less then 2 chars, bigger debounce

scazan:
This filters out searches that are too short for any kind of useful results and adjusts the search debounce time to better match typing speeds.

### DIFF
--- a/packages/gitbook/src/components/Search/SearchResults.tsx
+++ b/packages/gitbook/src/components/Search/SearchResults.tsx
@@ -98,7 +98,7 @@ export const SearchResults = React.forwardRef(function SearchResults(
                     : searchSpaceContent(spaceId, revisionId, query));
 
                 setResults(withAsk ? withQuestionResult(fetchedResults, query) : fetchedResults);
-            }, 250);
+            }, 400);
 
             return () => {
                 if (debounceTimeout.current) {
--- a/packages/gitbook/src/components/Search/SearchResults.tsx
+++ b/packages/gitbook/src/components/Search/SearchResults.tsx
@@ -98,7 +98,7 @@ export const SearchResults = React.forwardRef(function SearchResults(
                     : searchSpaceContent(spaceId, revisionId, query));
 
                 setResults(withAsk ? withQuestionResult(fetchedResults, query) : fetchedResults);
-            }, 400);
+            }, 350);
 
             return () => {
                 if (debounceTimeout.current) {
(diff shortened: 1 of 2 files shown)


packages/gitbook/src/components/Search/SearchResults.tsx
```diff
@@ -98,7 +98,7 @@ export const SearchResults = React.forwardRef(function SearchResults(
                     : searchSpaceContent(spaceId, revisionId, query));
 
                 setResults(withAsk ? withQuestionResult(fetchedResults, query) : fetchedResults);
-            }, 250);
+            }, 400);
```
taranvohra, line 101: Are you sure about this? The search request already take 300-500ms so an increase debounced on top of this might degrade the perceived performance

scazan: no not 100% as it's a delicate balance. Typing speed dictates the ~300-500ms debounce time though. Otherwise we fire off a lot of useless search queries which also causes some flashing results and pollutes analytics. 250ms is definitely too short. Let's trial it at 350 and see how it feels if this feels too drastic? Also keeping in mind that 250ms is about near instant reaction from a human perspective. Some other good news, the longer we allow the user to type before searching the faster the response will likely be. Wdyt?

taranvohra: Yeah let's do that

packages/gitbook/src/components/Search/server-actions.tsx
```diff
@@ -60,6 +60,10 @@ export async function searchSiteContent(args: {
     const { siteSpaceIds, query, cacheBust } = args;
     const pointer = getContentPointer();
 
+    if (query.length < 3) {
+        return [];
+    }
```
valentin0h, line 65: `C#` 😅

valentin0h: Or lexigraphs could be an issue.

Wouldn't the increase in debounce time be an improvement already?

scazan: yes, both would be an improvement really. a search like C# or a 2 char lexigraph could be argued that it wouldn't produce any useful results most of the time but I see your point. a higher debounce would help even more. At least, maybe we can agree that single char searches are not useful then and we can reduce that to 2. We are chipping away at the effectiveness of the improvement though.

scazan: switched it to one for now. The debounce increase will improve it a bit already so let's see how much over the next few days.

PR #2441, opened 2024-08-28
Fix spaces list of a site

taranvohra:
With sites we can have same space linked multiple times but it would show only once in the picker.

### DIFF
--- a/packages/gitbook/src/app/(space)/fetch.ts
+++ b/packages/gitbook/src/app/(space)/fetch.ts
@@ -224,7 +224,7 @@ async function fetchParentSite(args: {
 
     return {
         parent,
-        spaces: Object.values(spaces),
+        spaces,
     };
 }
 


packages/gitbook/src/app/(space)/fetch.ts
```diff
@@ -200,18 +200,18 @@ async function fetchParentSite(args: {
         getSiteSpaces({ organizationId, siteId, siteShareKey }),
     ]);
 
-    const spaces: Record<string, Space> = {};
+    const spaces: Array<Space> = [];
     siteSpaces.forEach((siteSpace) => {
-        spaces[siteSpace.space.id] = {
+        spaces.push({
```
spastorelli, line 205: One thing is that the space dropdown uses the Space title as label of each item and not the siteSpace customized title. So you end up with a list that looks like it has duplicate entries when technically it is a different site space.

I think we should do this only when (if we decide to have) the space dropdown use the site space "title" instead of the space title.

taranvohra: Scott has been working on a different PR where he fixes titles. https://github.com/GitbookIO/gitbook/pull/2367/files#diff-f15239843f1025db3421fea34edd5abd23b7cbc2d42abf78017d6158b6e5a733R221

spastorelli: How far is he from merging his changes though? I think we shouldn't have your changes shipped without part of his changes (i.e the part you linked that use the customized title in the variant dropdown) merged.

Even if it is to incorporate your changes in his or his in yours.
If it's a case where his are going to merged very soon after yours then all good. But I think we should not wait to long.

taranvohra: scott has merged his PR so I guess we can merge this one

spastorelli (approved):

PR #2442, opened 2024-08-30
Filter undefined values from schema.allOf when generating schema example

vibhanshub:


packages/react-openapi/src/generateSchemaExample.ts
```diff
@@ -138,9 +138,10 @@ export function generateSchemaExample(
             new Set(ancestors).add(schema),
         );
     }
-
-    if (schema.allOf && schema.allOf.length > 0) {
-        return schema.allOf.reduce(
+    
+    const filteredAllOf = schema.allOf && schema.allOf.filter(item => item !== undefined)
```
spastorelli, line 142: Can you give some more context on the issue you are fixing?

How can we end up with allOf items being undefined? Does the spec allow it?
I thought that the allOf should be a list of schema object or references to schema object, typed like this for example:
```
(ReferenceObject | SchemaObject)[]
```

Is this an issue in resolving a schema reference maybe? To me sounds more like something we should fix at the schema parsing level rather than having to filter undefined at this level.

PR #2445, opened 2024-09-03
Render mandatory headers in the code sample

vibhanshub:


packages/react-openapi/src/OpenAPICodeSample.tsx
```diff
@@ -19,6 +19,19 @@ export function OpenAPICodeSample(props: {
...
     const { data, context } = props;
 
+    const requiredHeaders = data.operation.parameters
+        ?.map(noReference)
+        .filter((param) => param.in === 'header' && param.required);
+
+    const headersObject: { [k: string]: string } = {};
+    requiredHeaders?.forEach((header) => {
+        const example = header.schema && generateSchemaExample(noReference(header.schema));
+        if (example) {
```
SamyPesse, line 29: You should be careful with `&&` I see you use it a lot and you sometimes miss what it means when the function can also return something negative.

```suggestion
        const example = header.schema ? generateSchemaExample(noReference(header.schema)) : undefined;
        if (example !== undefined) {
```

`generateSchemaExample` can return `null` | `false`, etc

SamyPesse (approved):

packages/react-openapi/src/OpenAPICodeSample.tsx
vibhanshub: Thank you @SamyPesse! Noted! Won't happen again.

PR #2448, opened 2024-09-04
Response examples with status code tabs

BrettJephson:
This PR adds tabs to responses examples section allowing access to all the example status codes.

It does not currently connect the status code selection in the Response section with the example in the Response example section as suggested in the original ticket (RND-3387).

Contributes to RND-3387


SamyPesse (approved):
Looks good! To synchronize the 2, we could imagine a global context managing the state and a `stateKey` passed to  the client component`<InteractiveSection stateKey="response" />`

Issue #2432, opened 2024-08-19
"Test it" tool URL not prepopulated with gitbook API and must be manually edited


benjamincburns:
As the title states, whenever I go to use the "test it" tool, the first thing I have to do is edit the URL by hand. It would reduce friction if I didn't have to do this every time. See the attached video for a quick demo that shows the issue.

https://github.com/user-attachments/assets/24692293-55e5-46f6-a80b-1441abd1df25

PR #2452, opened 2024-09-05
RND-2311: enums for server url in openapi block

BrettJephson:
With this PR, open api schemas with an array of servers and variables for server urls will allow selection of the variables and choice of server. 

Updating code samples and scalar modal's server url when updated. Which fixes [Issue #2432](https://github.com/GitbookIO/gitbook/issues/2432)

https://github.com/user-attachments/assets/22631b8c-e25e-4842-9ad3-5d9d1314f7b9


packages/react-openapi/src/OpenAPIServerURLForm.tsx
```diff
@@ -0,0 +1,63 @@
+'use client';
+
+import * as React from 'react';
+import { useRouter } from 'next/navigation';
```
SamyPesse, line 4: We can't use `next` modules in `react-openapi` as the main application doesn't use `next`.

We need to find a way to pass these by props to the high level component.

Or to use a client context:

```
import { OpenAPI, OpenAPIContext } from '@gitbook/react-openapi';

return (
  <OpenAPIContext onUpdateEnums={} enums={..}>
)
```

packages/gitbook/src/app/(space)/(content)/[[...pathname]]/page.tsx
```diff
@@ -89,6 +89,7 @@ export default async function Page(props: {
                         // Display the page feedback in the page footer if the aside is not visible
                         withPageFeedback && !page.layout.outline
                     }
+                    searchParams={searchParams}
```
Comment (SamyPesse) [inline packages/gitbook/src/app/(space)/(content)/[[...pathname]]/page.tsx:92]: Related to the approach I'm suggesting with a react context, we could avoid passing these search search params everywhere, and instead wrap it all in a client component that uses `useSearchParams` from next

packages/react-openapi/src/OpenAPIServerURLVariable.tsx
```diff
@@ -1,16 +1,63 @@
...
+ * Render a select if there is an enum for a Server URL variable
+ */
+function EnumSelect(props: {
+    value?: number;
+    name: string;
+    variable: OpenAPIV3.ServerVariableObject;
+}) {
+    const { value, name, variable } = props;
+    return (
+        <select
```
SamyPesse, line 45: Using a select, it means the url is much harder to copy, no?

SamyPesse (commented):
Quick first pass, I think we should abstract the concept of enum you are introducing here (enum is also a confusing name btw), to also be used for other client states that the block needs.

packages/react-openapi/src/OpenAPIServerURLVariable.tsx
BrettJephson: Yes, I've made this only appear if you deliberately choose to 'edit' the url, so it is still easy to copy until you edit it.

packages/react-openapi/src/OpenAPIServerURLForm.tsx
BrettJephson: This is updated to use a context. 

I have one problem that I'm not sure how best to resolve though - using a client component (with useContext) means I can't update the url in the Request code block as that is a server component. And I don't want to bring the codeblock code onto the client.

Looking at whether I can use `nuq/server` to cache the url in query param atm.

packages/react-openapi/package.json
```diff
@@ -6,7 +6,8 @@
             "types": "./dist/index.d.ts",
             "development": "./src/index.ts",
             "default": "./dist/index.js"
-        }
+        },
+        "./client": "./src/client.ts"
```
BrettJephson, line 10: Exporting the client component separately here because I found that If I add a client component (OpenAPIContextProvider) to the `/src/index` file it has the unfortunate side effect of turning all the server components exported in that file into client components.

SamyPesse: How does it work if you put `'use server'` at the top of the other file?

packages/react-openapi/src/OpenAPIContextProvider.tsx
```diff
@@ -0,0 +1,72 @@
...
+import * as React from 'react';
+import { OpenAPIOperationData } from './fetchOpenAPIOperation';
+import { getServersURL } from './utils';
+
+type OpenAPIContextProps = {
+    isPending?: boolean;
+    serverUrl?: string;
+    state?: Record<string, string> | null;
+    onUpdate: (params: Record<string, string> | null) => void;
+};
```
SamyPesse, line 12: We should document these props, I have no idea what they mean (`isPending`, etc).

packages/react-openapi/src/OpenAPIContextProvider.tsx
```diff
@@ -0,0 +1,72 @@
...
+const OpenAPIContext = React.createContext<OpenAPIContextProps | null>(null);
+
+export function useOpenAPIContext() {
+    return React.useContext(OpenAPIContext);
+}
+
+export function OpenAPIContextProvider(props: {
+    children: React.ReactNode;
+    data: OpenAPIOperationData;
```
SamyPesse, line 21: Passing the `data` here goes against the whole server component approach. This means we'll include the OpenAPI operation data for every element in HTML, resulting in very big HTML output.

packages/gitbook/src/components/DocumentView/OpenAPI/OpenAPI.tsx
```diff
@@ -45,21 +47,28 @@ async function OpenAPIBody(props: BlockProps<DocumentBlockSwagger>) {
...
-                    chevronRight: <Icon icon="chevron-right" />,
-                },
-                CodeBlock: PlainCodeBlock,
-                defaultInteractiveOpened: context.mode === 'print',
-                id: block.meta?.id,
-                blockKey: block.key,
-            }}
-            className="openapi-block"
-        />
+        <OpenAPIContext block={block} data={data}>
```
SamyPesse, line 53: Are we not supposed to have a `onUpdate` here?

SamyPesse (commented):
Can you add a LOT more comments to explain the relationship and the flow of things? Right now the PR is quite complex and and hard to diggest; it's hard to validate how it'll impact server/client components without understanding the flow of data you ate trying to put in place.

packages/react-openapi/package.json
BrettJephson: Will that do anything? I thought `use server` was the default and `use client` is the only directive that really does anything, at the top of a file at least.

packages/react-openapi/src/OpenAPIContextProvider.tsx
BrettJephson: I've refactored this a bit but could you explain the comment a bit more? 

The component is a client component that provides a React context, so I'm not sure how it affects the HTML output. But happy to admit it if I'm wrong as still getting a handle on some of the details of server components.

packages/gitbook/src/components/DocumentView/OpenAPI/OpenAPI.tsx
BrettJephson: This component is a wrapper around the actual context provider so I could keep it as a client component without polluting the rest of the tree which are server components. I've renamed it and commented as admittedly wasn't very clear.

packages/react-openapi/package.json
BrettJephson: Tried this and it doesn't work. Throws an error as can't export a client side component from a server component.

PR #2453, opened 2024-09-05
Add Python support in code samples

vibhanshub:
Tested the generated samples in Replit, and they work
Screen Shot 2024-09-05 at 4 28 59 PM
Screen Shot 2024-09-05 at 4 29 12 PM
Screen Shot 2024-09-05 at 4 29 25 PM


packages/react-openapi/src/code-samples.ts
```diff
@@ -65,6 +65,24 @@ export const codeSampleGenerators: CodeSampleGenerator[] = [
             return lines.map((line, index) => (index > 0 ? indent(line, 2) : line)).join(separator);
         },
     },
+    {
+        id: 'python',
+        label: 'Python',
+        syntax: 'python',
+        generate: ({ method, url, headers, body }) => {
+            let code = 'import requests\n';
+            code += `response = requests.${method.toLowerCase()}("${url}",\n`;
```
SamyPesse, line 74: Should we add a line break before? Usually, it looks like Python has line breaks between imports and code body.

packages/react-openapi/src/code-samples.ts
```diff
@@ -65,6 +65,24 @@ export const codeSampleGenerators: CodeSampleGenerator[] = [
...
+            if (headers) {
+                code += indent(`headers = ${JSON.stringify(headers, null, 2)},\n`, 11);
```
SamyPesse, line 76: The indentation looks weird in the screenshot; it looks like the code is not properly formatted.

Maybe we should try to make it look like https://codebeautify.org/python-formatter-beautifier

SamyPesse: Also can you look at enabling line wrapping for the code-block that we use for display?

SamyPesse (approved):

packages/react-openapi/src/code-samples.ts
vibhanshub: @SamyPesse Maybe I am doing it wrong, but adding line wrapping by modifying the css in the dev console makes the JS code look ugly
Screen Shot 2024-09-06 at 1 49 10 PM

I am not good with CSS, so maybe I am missing something

vibhanshub: But if you think this looks fine, I will figure out where to implement it in code

vibhanshub: Here is what the new Python code looks like. Does this look okay to you? Tried to make it as close to Python prettifiers as possible
Screen Shot 2024-09-06 at 1 52 32 PM

SamyPesse (approved):
Looks good!

packages/react-openapi/src/code-samples.ts
SamyPesse: Yeah looks good 👍 

I was not saying to do it with css for the wrapping, I think there is a prop for it (that is equivalent to the editor option)

PR #2457, opened 2024-09-06
RND-3387: synchronize response tabs

BrettJephson:
Synchronizes the selected tabs of Response and Response Example in OpenAPI.

Resolves RND-3387


packages/react-openapi/package.json
```diff
@@ -15,8 +15,9 @@
         "classnames": "^2.5.1",
         "flatted": "^3.2.9",
         "openapi-types": "^12.1.3",
-        "yaml": "1.10.2",
-        "swagger2openapi": "^7.0.8"
+        "recoil": "^0.7.7",
```
spastorelli, line 18: It looks like the CI that builds & deploy the Next.js app failed with this error:
```
Error: Can't build edge function /(space)/(content)/[[...pathname]]/page: Exceeds maximum edge function size: 4.43 MB / 4 MB
```
https://github.com/GitbookIO/gitbook/actions/runs/10737942599/job/29780494887?pr=2457#step:8:166

I wonder if it's happening with the addition of the `recoil` library as deps.

@SamyPesse have you ever encountered this issue? Do you think it's due to adding the new dep?
Any suggestion on how to get around or solve it?

spastorelli (approved):
Nice, looks good on my end. Just one comment on the CI job error.

packages/react-openapi/package.json
SamyPesse: There is a patch to fix this (https://github.com/GitbookIO/gitbook/blob/main/patches/%40vercel%252Fnext%404.3.6.patch). It looks like @BrettJephson needs to update its bun version (I'm using `1.1.26`) and re-run `bun install`

SamyPesse: I just pushed a commit for it

spastorelli: Ha good to know. Thanks

BrettJephson: The edge size does still seem to be an issue. I guess because I added `recoil` to `react-openapi`. I maybe need it as a peer dependency.

PR #2461, opened 2024-09-11
Display pattern if available in OpenAPI params

vibhanshub:


packages/gitbook/src/components/DocumentView/OpenAPI/style.css
```diff
@@ -178,6 +178,10 @@
     @apply prose-sm mt-2 text-dark/10 dark:text-light/10;
 }
 
+.openapi-schema-pattern {
```
SamyPesse, line 181: Let's group with `.openapi-schema-example`:

```suggestion
.openapi-schema-pattern, .openapi-schema-example {
```

Otherwise `@apply prose-sm` generates a good amount of CSS and I'm not sure how the css optimizer deals with it.

PR #2478, opened 2024-09-23
RND-2147: automatically scroll TOC

BrettJephson:
Bringing back the work @spastorelli did on automatically scrolling the TOC to current page (in https://github.com/GitbookIO/gitbook/pull/2425).

That work was rolled back due to a bug. Main fix is preventing multiple scroll triggers with a ref. 

[Note/out-of-scope]: The difference in rendering of the TOC in localhost v prod is kind of concerning and should be investigated further. I'd expect a dev build to render more often but the toc items re-render a lot more in prod which is why scrolling issues were only picked up in prod. 

I've been testing against the visual regression preview where I could replicate the issue, e.g. https://211cca89.gitbook-open.pages.dev/gitbook.gitbook.io/test-gitbook-open

### DIFF
--- a/packages/gitbook/src/components/TableOfContents/TOCScroller.tsx
+++ b/packages/gitbook/src/components/TableOfContents/TOCScroller.tsx
@@ -1,9 +1,10 @@
 'use client';
 
-import { useParams } from 'next/navigation';
 import React from 'react';
 
 import { ClassValue, tcls } from '@/lib/tailwind';
+
+import { useHash } from '../hooks';
 
 const TOCScrollContainerContext = React.createContext<React.RefObject<HTMLDivElement> | null>(null);
 
@@ -34,7 +35,7 @@ export function useScrollToActiveTOCItem(tocItem: {
 }) {
     const { isActive, linkRef } = tocItem;
 
-    const params = useParams();
+    const hash = useHash();
     const scrollContainerRef = React.useContext(TOCScrollContainerContext);
 
     React.useLayoutEffect(() => {
@@ -57,8 +58,6 @@ export function useScrollToActiveTOCItem(tocItem: {
                 });
             }
         }
-        // We've included `param` from `useParams` hook to listen to respond to changes to hash or
-        // for updates to route from next and previous browser buttons.
-        // See https://github.com/vercel/next.js/discussions/49465#discussioncomment-5845312
-    }, [isActive, linkRef, scrollContainerRef, params]);
+        // We've included `hash` from `useHash` hook as a dependency so we trigger the effect in response to changes to the url hash
+    }, [isActive, hash, linkRef, scrollContainerRef]);
 }
--- a/packages/gitbook/src/components/TableOfContents/TOCScroller.tsx
+++ b/packages/gitbook/src/components/TableOfContents/TOCScroller.tsx
@@ -65,11 +65,14 @@ export function useScrollToActiveTOCItem(props: {
             return;
         }
         const tocItem = linkRef.current;
-        tocItem?.scrollIntoView({
-            behavior: 'instant', // using instant as smooth can interrupt or get interrupted by other `scrollIntoView` changes
-            block: 'center',
+        const tocContainer = scrollContainerRef?.current;
+        if (!tocItem || !tocContainer) {
+            return;
+        }
+        tocContainer?.scrollTo({
+            top: tocItem.offsetTop - TOC_ITEM_OFFSET,
         });
         isScrolled.current = true;
         // We've included `hash` from `useHash` hook as a dependency so we trigger the effect in response to changes to the url hash
-    }, [hash, isActive, isOutOfView, linkRef]);
+    }, [hash, isActive, isOutOfView, linkRef, scrollContainerRef]);
 }
--- a/packages/gitbook/src/components/TableOfContents/TOCScroller.tsx
+++ b/packages/gitbook/src/components/TableOfContents/TOCScroller.tsx
@@ -1,0 +1,64 @@
+'use client';
+
+import { useParams } from 'next/navigation';
+import React from 'react';
+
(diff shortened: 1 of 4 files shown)


spastorelli (approved):
Looks good on my end. Thanks for picking this up.

packages/gitbook/src/components/TableOfContents/TOCScroller.tsx
```diff
@@ -0,0 +1,78 @@
...
+        }
+        if (!isOutOfView() || isScrolled.current) {
+            return;
+        }
+        const tocItem = linkRef.current;
+        const tocContainer = scrollContainerRef?.current;
+        if (!tocItem || !tocContainer) {
+            return;
+        }
+        tocContainer?.scrollTo({
```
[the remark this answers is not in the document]
gregberge: Why not using [`scroll-margin`](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-margin)?

packages/gitbook/src/components/TableOfContents/TOCScroller.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+import React from 'react';
+
+import { ClassValue, tcls } from '@/lib/tailwind';
+
+import { useHash } from '../hooks';
+
+const TOCScrollContainerContext = React.createContext<React.RefObject<HTMLDivElement> | null>(null);
+
+export function TOCScrollContainerProvider(props: {
```
gregberge, line 11:
```suggestion
export function TOCScrollContainer(props: {
```
It provides context but also display an element, so it's a container, not a provider.

packages/gitbook/src/components/TableOfContents/TOCScroller.tsx
```diff
@@ -0,0 +1,78 @@
...
+    children: React.ReactNode;
+    className?: ClassValue;
+}) {
+    const { children, className } = props;
+    const scrollContainerRef = React.createRef<HTMLDivElement>();
+
+    return (
+        <TOCScrollContainerContext.Provider value={scrollContainerRef}>
+            <div ref={scrollContainerRef} className={tcls(className)}>
```
gregberge, line 20:
```suggestion
            <div ref={scrollContainerRef} className={className}>
```

packages/gitbook/src/components/TableOfContents/TOCScroller.tsx
```diff
@@ -0,0 +1,78 @@
+'use client';
+
+import React from 'react';
+
+import { ClassValue, tcls } from '@/lib/tailwind';
+
+import { useHash } from '../hooks';
+
+const TOCScrollContainerContext = React.createContext<React.RefObject<HTMLDivElement> | null>(null);
```
gregberge, line 9:
```suggestion
const TOCScrollContainerRefContext = React.createContext<React.RefObject<HTMLDivElement> | null>(null);
```

packages/gitbook/src/components/TableOfContents/TOCScroller.tsx
```diff
@@ -0,0 +1,78 @@
...
+ * Scrolls the table of contents container to the page item when it becomes active
+ */
+export function useScrollToActiveTOCItem(props: {
+    isActive: boolean;
+    linkRef: React.RefObject<HTMLAnchorElement>;
+}) {
+    const { isActive, linkRef } = props;
+
+    const hash = useHash();
+    const scrollContainerRef = React.useContext(TOCScrollContainerContext);
```
gregberge, line 39: We should throw if `null` to ensure it's always used in a container.

packages/gitbook/src/components/TableOfContents/TOCScroller.tsx
```diff
@@ -0,0 +1,78 @@
...
+        const tocContainer = scrollContainerRef?.current;
+        if (!tocItem || !tocContainer) {
+            return;
+        }
+        tocContainer?.scrollTo({
+            top: tocItem.offsetTop - TOC_ITEM_OFFSET,
+        });
+        isScrolled.current = true;
+        // We've included `hash` from `useHash` hook as a dependency so we trigger the effect in response to changes to the url hash
+    }, [hash, isActive, isOutOfView, linkRef, scrollContainerRef]);
```
gregberge, line 77: Why including `hash`, the `isActive` should be enough, can it be active without hash?

packages/gitbook/src/components/TableOfContents/TOCScroller.tsx
```diff
@@ -0,0 +1,78 @@
...
+            return;
+        }
+        if (!isOutOfView() || isScrolled.current) {
+            return;
+        }
+        const tocItem = linkRef.current;
+        const tocContainer = scrollContainerRef?.current;
+        if (!tocItem || !tocContainer) {
+            return;
+        }
```
gregberge, line 71: You could actually refactor the code, you are testing these conditions twice. Here and in the `isOutOfView`. By just creating a function that accepts the both elements you could avoid declaring a callback and just pass the elements to check this.

gregberge (changes requested):

packages/gitbook/src/components/TableOfContents/TOCScroller.tsx
BrettJephson: This project uses `tcls` and `twMerge` and this change results in a typescript error, ClassValue can't be assigned to `string | undefined`.

packages/gitbook/src/components/TableOfContents/TOCScroller.tsx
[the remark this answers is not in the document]
BrettJephson: I don't think it works well with `scrollTo`. At least it didn't when I tried it.

gregberge (approved):

PR #2486, opened 2024-09-25
Fix: mobile navigation scroll

BrettJephson:
Fix scrolling to top when navigating to a new page (on mobile)

### DIFF
--- a/packages/gitbook/src/components/Header/HeaderMobileMenu.tsx
+++ b/packages/gitbook/src/components/Header/HeaderMobileMenu.tsx
@@ -39,6 +39,7 @@ export function HeaderMobileMenu(props: Partial<React.ButtonHTMLAttributes<HTMLB
     // Close the navigation when navigating to a page
     useEffect(() => {
         document.body.classList.remove(globalClassName);
+        window.scrollTo(0, 0);
     }, [pathname]);
 
     useEffect(() => {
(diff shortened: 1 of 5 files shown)


packages/gitbook/src/components/Header/HeaderMobileMenu.tsx
```diff
@@ -39,6 +39,7 @@ export function HeaderMobileMenu(props: Partial<React.ButtonHTMLAttributes<HTMLB
     // Close the navigation when navigating to a page
     useEffect(() => {
         document.body.classList.remove(globalClassName);
+        window.scrollTo(0, 0);
```
SamyPesse, line 42: Should we put this somewhere else?

It's confusing at the moment as it applies on all devices but because of the name of the component, it seems tied to mobile.

SamyPesse (changes requested):
I don't think this approach works as it means navigating to anchor is no longer possible (try clicking a link with a `#hash`), it'll always scroll to the top.

packages/gitbook/src/components/Header/HeaderMobileMenu.tsx
BrettJephson: That is confusing. I've combined it with `useScrollToHash` as a `useScrollPage` hook that's runs when pathname or hash changes.

SamyPesse (approved):

PR #2491, opened 2024-09-25
Add transitions to Sections bg and line

viktorrenkema:
This PR adds a line animation transitions for the current active chapter in the Sections/ToC, respecting `prefers-reduced-motion`.

[Related Slack thread](https://gitbook.slack.com/archives/C032HV6N7/p1727246503552519?thread_ts=1727190633.551719&cid=C032HV6N7).

Before:

https://github.com/user-attachments/assets/737e58ae-7ea5-43de-9807-f7d9f261e8df

After:

https://github.com/user-attachments/assets/a21d7ff4-72ce-4899-8beb-9bcf849887e5


SamyPesse:
@viktorrenkema can you run `bun run changeset` to write a changelog entry?

While I think it's a great addition, I'm slightly concerned on the animated background, I find the background to be quite heavy/distracting.

@zenoachtig what do you think? Here is a preview link: https://a859aaea.gitbook-open.pages.dev/docs.gitbook.com/content-editor/editor/navigation

zenoachtig:
I agree with @SamyPesse's concern, i think the gray background is indeed a lot subtler and as a bonus it matches the left sidebar better 😄

The square corners on this element annoy me a bit, but I don't think rounding the two right corners is a good idea either. Probably nothing worth doing about.

viktorrenkema:
> I agree with @SamyPesse's concern, i think the gray background is indeed a lot subtler and as a bonus it matches the left sidebar better 😄
> 
> The square corners on this element annoy me a bit, but I don't think rounding the two right corners is a good idea either. Probably nothing worth doing about.

Yeah the sharp corners bug me too, but it also matches the 'was this helpful' picker below etc. we could prolly pimp this component in general, it can be prettier. but not worthwhile for now, and this pr provides the backbone for the interactive part and we can always tweak the exact design of it if we were to redesign the whole component.

@SamyPesse wdyt, keep the suggested gray background, or remove the background part altogether?

BrettJephson:
Just having a look at the preview branch. I think this looks great and had a couple of minor suggestions:

1. When you have hovered over an item and then move away the highlight returns suddenly to the current item. Compare to how Remote do it with a fade out and I think the latter looks a bit slicker.
2. We could maybe do something with focus as well, I notice Remote moves the highlight to the focused element when you change focus even though you can still move the selection with cursor too.

zenoachtig (approved):

packages/gitbook/src/components/PageAside/ScrollSectionsList.tsx
```diff
@@ -28,37 +37,46 @@ export function ScrollSectionsList(props: { sections: DocumentSection[] }) {
...
 
     return (
-        <ul className={tcls('border-l', 'border-dark/2', 'dark:border-light/1', 'space-y-1')}>
+        <ul className={tcls('border-l', 'border-dark/2', 'dark:border-light/1', 'pl-1')}>
             {sections.map((section) => (
-                <li key={section.id} className={tcls('flex', 'flex-row')}>
+                <motion.li
+                    key={section.id}
+                    className={tcls('flex', 'flex-row', 'relative', 'h-fit')}
+                    onMouseEnter={() => setHoveredId(section.id)}
```
BrettJephson, line 45: Do we still use the `hoveredId` ?

BrettJephson (approved):
Looks good!

packages/gitbook/src/components/PageAside/ScrollSectionsList.tsx
viktorrenkema: Ah no forgot it, thnx!

PR #2493, opened 2024-09-26
generate upto URL with 3 segments to support the new site url format

taranvohra:
With the new site URLs that doesn't include `/v/` for variant we still want to be able to match for scenarios like:
`company.gitbook.io/siteSlug/shareKey/variantSlug` (up to 3 segments)

### DIFF
--- a/packages/gitbook/src/lib/middleware.test.ts
+++ b/packages/gitbook/src/lib/middleware.test.ts
@@ -4,7 +4,9 @@ import { getURLLookupAlternatives, normalizeURL } from './middleware';
 
 describe('getURLLookupAlternatives', () => {
     it('should return all URLs up to the root', () => {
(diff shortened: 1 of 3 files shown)


packages/gitbook/src/lib/middleware.test.ts
```diff
@@ -4,7 +4,9 @@ import { getURLLookupAlternatives, normalizeURL } from './middleware';
 
 describe('getURLLookupAlternatives', () => {
     it('should return all URLs up to the root', () => {
-        expect(getURLLookupAlternatives(new URL('https://docs.mycompany.com/a/b/c'))).toEqual({
+        const res = getURLLookupAlternatives(new URL('https://docs.mycompany.com/a/b/c'));
+        console.log(res);
```
SamyPesse, line 8:
```suggestion
```

SamyPesse (approved):
Can you add a changeset to explain that it's about supporting variant with the `/v/` ?

PR #2497, opened 2024-09-27
URL alternative resolution criteria

taranvohra:
This PR changes the acceptance criteria of an alternative URL lookup. With the upcoming release of site URLs without `/v/` we need to make sure that the lookup result we choose amongst a bunch of alternatives is the right one. 

The criteria now is as follows:-
- The alternative is the `primary` one. This is the easiest one to decide as it is the longest and most inclusive path so should always be considered.
- The site content resolution resolved with a siteSpace that is different from the defaultSiteSpace. Again this also implies that we had the `variant` path in our alternative URL so consider this one too because it is the deepest one. Site URLs always end at the `variant` path after which we have the page paths. 
  - `company.gitbook.io/siteSlug/variantSlug/page`
  - `company.gitbook.io/siteSlug/shareKey/variantSlug/page`
  - `docs.company.com/variantSlug/page`
  - `docs.company.com/variantSlug/shareKey/page`


packages/gitbook/src/lib/api.ts
```diff
@@ -127,7 +127,15 @@ export function withAPI<T>(client: GitBookAPI, fn: () => Promise<T>): Promise<T>
...
 
 export type PublishedContentWithCache =
-    | ((PublishedContentLookup | PublishedSiteContentLookup) & {
+    | ((
+          | PublishedContentLookup
+          | (PublishedSiteContentLookup & {
+                /**
+                 * ID of the default site space variant in the URL lookup.
+                 */
+                defaultSiteSpace: string;
```
SamyPesse, line 136: Do we need this type? Why is not coming from the api?

SamyPesse (approved):

packages/gitbook/src/lib/api.ts
taranvohra: We'll have to bump the API client which I don't want to do now as its already being done in the cleanup PR that also addresses typing issues. I'll add a `TODO` here anyway

PR #2499, opened 2024-09-27
Switch to aria-current="page" for the TOC items

SamyPesse:


packages/gitbook/src/components/TableOfContents/ToggleableLinkItem.tsx
```diff
@@ -96,7 +96,7 @@ export function ToggleableLinkItem(props: {
             <Link
                 ref={linkRef}
                 href={href}
-                aria-selected={isActive}
+                {...(isActive ? { 'aria-current': 'page' } : {})}
```
gregberge, line 99: `aria-current={isActive ? 'page' : undefined}` is cleaner.

PR #2505, opened 2024-09-30
Remove /v/ from e2e tests

taranvohra:


packages/gitbook/e2e/pages.spec.ts
```diff
@@ -156,13 +156,13 @@ const testCases: TestsCase[] = [
...
                     await page.waitForURL(
-                        'https://gitbook-open-e2e-sites.gitbook.io/api-multi-versions/v/2.0/reference/api-reference/pets?fallback=true',
+                        'https://gitbook-open-e2e-sites.gitbook.io/api-multi-versions/2.0/reference/api-reference/pets?fallback=true',
                     );
                 },
             },
             {
                 name: 'Keep navigation path/route when switching variant (Share link)',
-                url: 'api-multi-versions-share-links/bRfQbzwsK8rbN1GRxx7K/reference/api-reference/pets',
+                url: 'api-multi-versions-share-links/8tNo6MeXg7CkFMzSSz81/reference/api-reference/pets',
```
taranvohra, line 165: Changed the `shareKey` because the tests would keep timing out because of the old cache. I've checked the URL separately and it works so there is nothing wrong in the resolution logic

PR #2511, opened 2024-10-03
Update Checkbox component

zenoachtig:
The Checkbox component was broken in some instances, particularly inside tables. It also used a low contrasting checkmark and light background. This PR gives Checkbox a brighter spot of colour, and introduces a new way to always use the right text colour on top of the primary colour using the `contrast-primary` colour classes.

- Make Checkbox component slightly bigger
- Change sizing of Checkbox so it's shown correctly in tables
- Make Checkbox use user-supplied primary color (primary-500)
- Add extra contrasting bg color and border using `contrast-more` (bg only in light mode, dark is contrasting enough)
- Add `contrast-primary` variables to provide white or black text for each shade of `color-primary`

### Current

Screenshot 2024-10-03 at 17 56 32

### Proposed

Screenshot 2024-10-03 at 17 59 09

### Works with all theme colours

Screenshot 2024-10-03 at 17 42 03

Screenshot 2024-10-03 at 17 44 35

Screenshot 2024-10-03 at 18 06 18

Screenshot 2024-10-03 at 17 40 50


package.json
```diff
@@ -25,6 +25,7 @@
     ],
     "devDependencies": {
         "@changesets/cli": "^2.27.7",
+        "postcss-color-contrast": "^1.1.0",
```
zenoachtig, line 28: Actually this should probably be a normal dependency 😬

Cassienicolelair703:
#Dad#Seeyousoon#831to#209

PR #2512, opened 2024-10-08
Add automatic color contrast in site header, restyle search button

zenoachtig:
Using the same color contrast function introduced with https://github.com/GitbookIO/gitbook/pull/2511, we can automatically choose the right `linkColor` to match the site nav's background color. This only applies to the Bold theme, where the user’s primary color is used as the site nav's background. In other themes, the `linkColor` is predetermined or explicitly set by the user.

This PR also reworks `SearchButton` to use the `linkColor` when used in the site top nav, instead of derivatives of the `header-background` color.

Default blue
**Default blue, stays the same** (Left current, right proposed)

Orange
**Orange, swaps color to be more contrasting** (Left current, right proposed)

Pure black, currently broken
**Pure black, currently broken** (Left current, right proposed)

Pink, key hint uses wrong color
**Pink, key hint uses wrong color** (Left current, right proposed)

Additionally, `SearchButton` now features more contrasting styling for users who opt into it using the `prefers-contrast: more` media query.
High contrast styling
**High contrast styling** (Left current, right proposed)

Finally, on Mac we simplify the `⌘ + K` key hint into `⌘ K` to match the operating system and other websites using this pattern. On Windows we still use `Ctrl + K` in full to match that OS’s convention.


.changeset/bright-cars-explain.md
```diff
@@ -0,0 +1,5 @@
+---
+'gitbook': patch
+---
+
+Add automatic color contrast in site header, restyle search button
```
gregberge, line 5: We don't need a changeset I think.

gregberge (approved):
It looks great!

.changeset/bright-cars-explain.md
SamyPesse: We should have changeset for everything. It acts as a changelog for our public content output.

SamyPesse (approved):
It looks good! Let's get this merged

.changeset/bright-cars-explain.md
gregberge: OK I didn't know. So we should deploy only if changeset is present (is it the case?).

PR #2514, opened 2024-10-11
add support for sites in multi id mode

taranvohra:


packages/gitbook/src/middleware.ts
```diff
@@ -406,11 +425,14 @@ async function lookupSpaceInMultiMode(request: NextRequest, url: URL): Promise<L
  * When serving multi spaces with the ID passed in the path.
  *
  * The format of the path is:
- *   - /~space/:id/:path
- *   - /~space/:id/~changes/:changeId/:path
- *   - /~space/:id/~revisions/:revisionId/:path
+ *   - /~space|~site/:id/:path
+ *   - /~space|~site/:id/~changes/:changeId/:path
+ *   - /~space|~site/:id/~revisions/:revisionId/:path
```
SamyPesse, line 430: Can this ones actually work with `/~site` ?

SamyPesse (approved):
Should we just add a small e2e test for it?

packages/gitbook/src/middleware.ts
taranvohra: Yes. The extracted revisionId or changeId is passed as headers in the lookup result just like normally

PR #2517, opened 2024-10-11
Update ogimage with new design

zenoachtig:
Our current OpenGraph / social / link preview images are very basic. Let’s make them nicer.

## Current
V4wFq98rvrFDby1IZXsd-5

### Default themes (90% of sites)
By default, we use the site’s primary colour to generate a gradient background. We use the site’s default theme mode (`light` or `dark`) to invert the colour scheme if desired.

V4wFq98rvrFDby1IZXsd-7

fTeGXT2KcWiIhuDouroB-2

a97K2vWVzHrxp7MPZuYe

e1plEWuSincfqEcmAwiR

J43iuEEkSDknOZLf0PcB

### Bold and custom themes (10% of sites)

When a site has the `Bold` or `Custom` header theme, we need to change the background of the image to make sure it always works with the site’s logo and colour. Instead of the primary colour on white or black, we use the site’s `header.backgroundColor` as the basis for the background, and overlay the site’s `header.linkColor` on top. To make sure body text is legible, we employ the same `colorContrast()` check as we perform in the header to pick form white or black text.

W0jQW9cv6VefOtRTEn1h

TEbrzcJSYpRqdsZT7UOG
  
dYmNacqUHCvbTHu1yRlB

-LtH6rb0F--BRmm0NEKa

L8Dpy94S9cjPUlbPd0Dt-2

## Caveats and next steps
- **Fonts**: 
  - In order to load a custom font into the `satori` renderer, we need to supply an `ArrayBuffer` of font data by fetching the font file and passing it to `ImageResponse`. 
  - Ideally we’d use the already loaded and cached fonts from `next/font/google` for this, but this doesn’t expose the cached font files, and thus can’t be passed along. 
  - We now colocate two weights of the Inter font in this repo to make it available as a static asset.
  - Ideally we would use the site’s chosen font in the ogimage, but for the above reason this is difficult to do unless we add all fonts to the repository (definitely doable, but perhaps not worth the effort).
- **Emoji and icons**: It’s difficult (but not impossible) to add custom icon fonts and emojis to the renderer, but it’s probably too much effort to be worth it for now. Emojis are supported by default in the renderer but with a different font (Twemoji instead of Emojione)


packages/gitbook/src/app/(space)/(core)/~gitbook/ogimage/[pageId]/route.tsx
```diff
@@ -1,53 +1,210 @@
...
+    const bold = customization.header.preset == CustomizationHeaderPreset.Bold;
+    const custom = customization.header.preset == CustomizationHeaderPreset.Custom;
+
+    // We have no access to CSS variables, so we'll have to hardcode some values
+    const baseColors = {
+        light: '#fff',
+        dark: '#111827',
+    };
+
+    const colors = {
```
Comment (zenoachtig) [inline packages/gitbook/src/app/(space)/(core)/~gitbook/ogimage/[pageId]/route.tsx:71]: This logic is very messy, but it works. I’m intending to refactor this into more readable code before we merge.

zenoachtig:
Something seems on the ogimage route seems to crash in the branch deployment, will take a look and fix after the weekend!

packages/gitbook/src/app/(space)/(core)/~gitbook/ogimage/[pageId]/route.tsx
```diff
@@ -1,53 +1,210 @@
...
 
+// TODO: Support all fonts available in GitBook
+// Right now this is impossible since next/font/google does not expose the cached font file
+// Another option would be to use the Satori prop `loadAdditionalAsset` [example](https://github.com/vercel/satori/blob/main/playground/pages/index.tsx),
+// but this prop isn't (yet) exposed through `ImageResponse`.
+const interRegular = fetch(
+    new URL('../../../../../../fonts/Inter/Inter-Regular.ttf', import.meta.url),
+).then((res) => res.arrayBuffer());
+const interBold = fetch(
+    new URL('../../../../../../fonts/Inter/Inter-Bold.ttf', import.meta.url),
```
Comment (SamyPesse) [inline packages/gitbook/src/app/(space)/(core)/~gitbook/ogimage/[pageId]/route.tsx:24]: Does it actually work? I'm surprised, it should be possible to import them instead.

Also should this fetch be moved in the `GET` function instead

Comment (SamyPesse) [reply packages/gitbook/src/app/(space)/(core)/~gitbook/ogimage/[pageId]/route.tsx:24]: Ah yes https://vercel.com/guides/using-custom-font ok 😅

packages/gitbook/src/app/(space)/(core)/~gitbook/ogimage/[pageId]/route.tsx
```diff
@@ -1,53 +1,210 @@
...
...YRXRGcRdwAAAGBKgju6E2ABAAAAMDQBFgAAAABDE2ABAAAAMDQBFgAAAABDE2ABAAAAMDQBFgAAAABDuy0AAADAqG7KmEbdr2tcOqbT/f9nmgh0fj2GOOaIHQ4AAAAiWVu7L33+1kyg/tzphf/fss2Wz9v6/Od+5rlj3mN/Hv6uPV6XLds8lbK5v7y07ZbPBQAAAPglTMh4N9e7km8ZptrOH8sgrIEFAAAALHUql2dfwS4EWAAAAAAMTYAFAAAAcJlZZwcTYAEAAADwlGGCOwEWAAAAAEMTYAEAAABruJ2O7gRYAAAAAAxNgAUAAADAY6diDSwAAAAAWEaABQAAAMDQBFgAAADAGhkXcbdw/cEEWAAAAACX3RQOJcACAAAAuMwMrIMJsAAAAAAYmgALAAAAgKEJsAAAAIA1LOJOdwIsAAAAAIYmwAIAAADgKcPMPBNgAQAAADA0ARYAAAAAQxNgAQAAAFx2KhZyP5QACwAAAIChCbAAAAAAeIpF3AEAAABgCQEWAAAAAEMTYAEAAAAwNAEWAAAAwGUZv4FwqGMWYAEAAADwFIu4AwAAAMASAiwAAAAAhibAAgAAAGBoAiwAAAAAhibAAgAAAGBoAiwAAAAAnuJbCAEAAABgCQEWAAAAwGWnMtBspIwEWAAAAAAMTYAFAAAAwNAEWAAAAAAMTYAFAAAAwNAEWAAAAACXWcD9YAIsAAAAAB4b6psXBVgAAAAADE2ABQAAAMDQBFgAAAAADE2ABQAAAMDQBFgAAAAAPMUi7gAAAACwhAALAAAAgKEJsAAAAAAYmgALAAAAgKEJsAAAAAAYmgALAAAAgKEJsAAAAAAYmgALAAAAgKEJsAAAAAB4yqkMQoAFAAAAwNAEWAAAAAAMTYAFAAAAwNAEWAAAAAAMTYAFAAAAwNAEWAAAAAA8ZZhvIbwpAAAAAGS0Nhca7fkAAAAAJHZTBEoAAAAADG6YAMsaWAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAQHD/AagXIjjIr64XAAAAAElFTkSuQmCC'
```
Comment (SamyPesse) [inline packages/gitbook/src/app/(space)/(core)/~gitbook/ogimage/[pageId]/route.tsx:155]: Should we move these as variables?

or event better, either:
- Store them as images and import them (I don't know if next supports this out of the box)
- Put them in the public folder and reference by url (if it works?)

SamyPesse (approved):
Looks great!

packages/gitbook/src/app/(space)/(core)/~gitbook/ogimage/[pageId]/route.tsx
Comment (zenoachtig) [reply packages/gitbook/src/app/(space)/(core)/~gitbook/ogimage/[pageId]/route.tsx:155]: I first had them as images in the public folder and that worked, but we (purposely?) have `/public/~gitbook/static/` in our `.gitignore` so it seemed cleaner to keep that as-is and also save us the round trip of fetching the image. But I agree it’s probably cleaner to have them in there, I’d just have to change that gitignore to store them in the right place.

Comment (zenoachtig) [reply packages/gitbook/src/app/(space)/(core)/~gitbook/ogimage/[pageId]/route.tsx]: Update: proposing to move these images into `/src/images` and import them as you suggested.

Since these images should only be used as part of the ogimage generation process, it seems incorrect to host them in `/public/~gitbook/static/images` and so they should be part of `/src` instead.

PR #2519, opened 2024-10-14
Refactor `getContentPointer` usage into `getSiteContentPointer`

taranvohra:
This PR splits `app/(space)` into `app/(site)` and `app/(space)/~gitbook/pdf`
- `app/(site)` will ALWAYS use `getSiteContentPointer`
- `app/(space)/~gitbook/pdf` supports both pointers (special case)

Next PR:
- `getSpaceCustomization` should return the default customization object instead of making an API call
- `listSpaceIntegrationScripts` should return `[]` instead of making an API call


packages/gitbook/package.json
```diff
@@ -16,7 +16,7 @@
         "clean": "rm -rf ./.next && rm -rf ./public/~gitbook/static"
     },
     "dependencies": {
-        "@gitbook/api": "0.60.0",
+        "@gitbook/api": "0.64.1",
```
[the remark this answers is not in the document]
SamyPesse: Yes I'll make a PR once this one is merged

packages/gitbook/src/app/(site)/(core)/robots.txt/route.ts
```diff
@@ -3,22 +3,18 @@ import { NextRequest } from 'next/server';
...
 export async function GET(req: NextRequest) {
-    const pointer = getContentPointer();
-    const space = await getSpace(
-        pointer.spaceId,
-        'siteId' in pointer ? pointer.siteShareKey : undefined,
-    );
-    const site = 'siteId' in pointer ? await getSite(pointer.organizationId, pointer.siteId) : null;
+    const pointer = getSiteContentPointer();
+    const space = await getSpace(pointer.spaceId, pointer.siteShareKey);
+    const site = await getSite(pointer.organizationId, pointer.siteId);
```
SamyPesse, line 17: It looks like we can do both in parallel ?

packages/gitbook/src/app/(site)/layout.tsx
```diff
@@ -0,0 +1,16 @@
+import { SpaceRootLayout } from '@/components/SpaceRootLayout';
```
SamyPesse, line 1: Should we name it `CustomizationRootLayout` instead?

It has nothing to do with a space really.

packages/gitbook/src/app/(space)/~gitbook/pdf/pointer.ts
```diff
@@ -0,0 +1,15 @@
+import { SiteContentPointer, SpaceContentPointer } from '@/lib/api';
+import { getSiteContentPointer, getSpacePointer } from '@/lib/pointer';
+
+/**
+ * PDF generation can be done at the site level (e.g. docs.foo.com/~gitbook/pdf) or
+ * at the space level (e.g. open.gitbook.com/~space/:spaceId/~gitbook/pdf). This function
```
SamyPesse, line 6: Let's just add a small explanation that this is for in-app PDF export of private content

packages/gitbook/src/components/Contexts/ClientContexts.tsx
SamyPesse, line 1: Should we have all of this in a `RootLayout` fodler with `CustomizationRootLayout` instead of being in a generic `Contexts` folder?

packages/gitbook/src/components/SpaceLayout/SpaceLayout.tsx
```diff
@@ -17,15 +16,15 @@ import { CONTAINER_STYLE } from '@/components/layout';
...
 import { TableOfContents } from '@/components/TableOfContents';
-import { ContentPointer, ContentTarget, SiteContentPointer } from '@/lib/api';
+import { ContentTarget, SiteContentPointer } from '@/lib/api';
 import { ContentRefContext } from '@/lib/references';
 import { tcls } from '@/lib/tailwind';
 
 /**
  * Render the entire content of the space (header, table of contents, footer, and page content).
  */
 export function SpaceLayout(props: {
```
SamyPesse, line 26: This is why I think the other one should be named `CustomizationRootLayout`, otherwise it's confusing with this component. This one has a valid name since it renders a space.

packages/gitbook/src/components/SpaceRootLayout/SpaceRootLayout.tsx
```diff
@@ -11,30 +11,28 @@ import assertNever from 'assert-never';
...
 import '@gitbook/icons/style.css';
-import { getContentPointer } from './fetch';
 
 /**
  * Layout shared between the content and the PDF renderer.
  * It takes care of setting the theme and the language.
  */
-export default async function SpaceRootLayout(props: { children: React.ReactNode }) {
-    const { children } = props;
+export async function SpaceRootLayout(props: {
```
SamyPesse, line 30:
```suggestion
export async function CustomizationRootLayout(props: {
```

packages/gitbook/src/components/SpaceRootLayout/SpaceRootLayout.tsx
```diff
@@ -11,30 +11,28 @@ import assertNever from 'assert-never';
...
-import { emojiFontClassName } from '@/components/primitives';
 import { fonts, ibmPlexMono } from '@/fonts';
 import { getSpaceLanguage } from '@/intl/server';
-import { getSiteLayoutData, getSpaceLayoutData } from '@/lib/api';
 import { getStaticFileURL } from '@/lib/assets';
 import { hexToRgb, shadesOfColor } from '@/lib/colors';
 import { tcls } from '@/lib/tailwind';
 
-import { ClientContexts } from './ClientContexts';
+import { ClientContexts } from '../Contexts';
```
SamyPesse, line 20: Let's move it in this folder

packages/gitbook/src/lib/api.ts
```diff
@@ -126,18 +125,25 @@ export function withAPI<T>(client: GitBookAPI, fn: () => Promise<T>): Promise<T>
     return apiSyncStorage.run(client, fn);
 }
 
+// FIXME: Fix @gitbook/api types for PublishedSiteContentLookup
```
SamyPesse, line 128: Still true?

SamyPesse (approved):
It looks good, we should monitor it when we deploy and test all routes.

packages/gitbook/src/lib/api.ts
taranvohra: Yes, it'll be a small change in our API type to be able to extract this type for `SpaceContentLookup` because currently it's an union of `{  redirect: '...' } | { site: '...' }`, I want the latter one to be it's own type

PR #2522, opened 2024-10-14
Mock data for getSpaceCustomization and getSpaceIntegrationScripts

taranvohra:
Making API calls to these endpoints are useless now with sites. We only need to the default data to ensure rendering works end to end

### DIFF
--- a/packages/gitbook/src/lib/api.ts
+++ b/packages/gitbook/src/lib/api.ts
@@ -814,23 +814,6 @@ export async function getCurrentSiteCustomization(args: {
 /**
  * Get the customization settings for a space from the API.
  */
-export const getSpaceCustomizationFromAPI = cache({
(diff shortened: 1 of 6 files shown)


packages/gitbook/src/lib/api.ts
```diff
@@ -826,29 +811,12 @@ export async function getCurrentSiteCustomization(args: {
...
-        return cacheResponse(response, {
-            revalidateBefore: 60 * 60,
-        });
-    },
-});
-
 /**
  * Get the customization settings for a space from the API.
  */
 export async function getSpaceCustomization(spaceId: string): Promise<CustomizationSettings> {
```
SamyPesse, line 817: Should we move this to the PDF export folder instead of having it globally?

SamyPesse (approved):

PR #2529, opened 2024-10-16
Show placeholder ad when site ads status is pending or in-review

spastorelli:
### Preview

https://da511e05.gitbook-open.pages.dev/spastorelli.gitbook.io/steeves-personal-organization?ads_preview=placeholder


packages/gitbook/src/components/Ads/renderAd.tsx
```diff
@@ -78,3 +73,50 @@ async function fetchAd({
 
     return null;
 }
+
+function getPlaceholderAd(): { ad: AdItem; ip: string } {
+    const { ip } = getUserAgentAndIp();
+    const imageHref = absoluteHref('~gitbook/static/images/ad-rainbow.svg', true);
```
spastorelli, line 79: Not sure what's best in this case for the call to `absoluteHref()`: should I pass `withHost: true` or not. Initially went with **not** passing it. And lastly tested it with `withHost: true` as I could see that it's what we used for the fallback in SpaceIcon so went with it.

I also saw that there is `getStaticFileURL` so not sure what's best here.

SamyPesse (approved):
Perfect!

PR #2530, opened 2024-10-17
RND-5227: Site section tabs

BrettJephson:


packages/gitbook/src/app/(site)/fetch.ts
```diff
@@ -129,20 +139,44 @@ async function resolvePage(
...
 /**
  * Fetch the structure of an organization site.
- * This includes the site and its spaces.
+ * This includes the site and its sections and spaces.
  */
 async function fetchSiteStructure(args: {
     organizationId: string;
     siteId: string;
     siteShareKey: string | undefined;
+    siteSpaceId: string | undefined;
```
taranvohra, line 148:
```suggestion

```
We don't need this

packages/gitbook/src/app/(site)/fetch.ts
```diff
@@ -68,6 +77,7 @@ export async function fetchPageData(params: PagePathParams | PageIdParams) {
             organizationId: content.organizationId,
             siteId: content.siteId,
             siteShareKey: content.siteShareKey,
+            siteSpaceId: content.siteSpaceId,
```
taranvohra, line 80: Same thing, we shouldn't be passing `siteSpaceId`

packages/gitbook/src/app/(site)/fetch.ts
```diff
@@ -33,19 +33,28 @@ export async function fetchContentData() {
...
                 siteShareKey: content.siteShareKey,
+                siteSpaceId: content.siteSpaceId,
             }),
         ]);
 
     const site = siteStructure.site;
-    const spaces = siteStructure.spaces;
+
+    const sections = siteStructure.sections;
+    const section = siteStructure.sections?.find((section) => section.id === content.siteSectionId);
```
taranvohra, line 43: If it's a sections site then `section` should ALWAYS be defined else we try. I think you can put a condition that 
```js
if(sections) {
   assertIsDefined(section);
}
```

packages/gitbook/src/app/(site)/fetch.ts
```diff
@@ -33,19 +33,28 @@ export async function fetchContentData() {
...
-    const spaceRelativeToParent = spaces.find((space) => space.id === content.spaceId);
+    const spaceRelativeToParent = spaces?.find((space) => space.id === content.spaceId);
 
     return {
         content,
         contentTarget,
         space: spaceRelativeToParent ?? space,
         pages,
+        sections,
+        section,
```
taranvohra, line 57: On that same note the return typing should also indicate that `section` would always exist if site has sections. We could maybe refactor this to be

```js
return {
  ...,
 sections: sections ? { list: sections, section } : undefined
}
```

packages/gitbook/src/components/SiteSectionTabs/SiteSectionTab.tsx
taranvohra, line 1: Lets also add few comments to these components

taranvohra (approved):
Looks good overall, minor suggestions to improve the typing DX

packages/gitbook/src/app/(site)/fetch.ts
BrettJephson: Forgot to remove it. Thanks!

packages/gitbook/src/app/(site)/fetch.ts
```diff
@@ -54,6 +64,19 @@ export async function fetchContentData() {
     };
 }
 
+function getSiteSectionsList(siteSectionId?: string, sections?: SiteSection[] | null) {
+    if (!sections) {
+        return null;
+    }
+    const section = sections?.find((section) => section.id === siteSectionId);
```
taranvohra, line 71:
```suggestion
    const section = sections.find((section) => section.id === siteSectionId);
```

packages/gitbook/src/app/(site)/fetch.ts
```diff
@@ -54,6 +64,19 @@ export async function fetchContentData() {
     };
 }
 
+function getSiteSectionsList(siteSectionId?: string, sections?: SiteSection[] | null) {
```
taranvohra, line 67:
```suggestion
function parseSiteSectionsList(siteSectionId?: string, sections?: SiteSection[] | null) {
```

taranvohra, line 67: This signature is a bit confusing. Why not `(sections: SiteSection[], siteSectionId: string)` and then do `const siteSections = content.siteSectionId && siteStructure.sections ? parseSiteSections(...) : null`

taranvohra (approved):

packages/gitbook/src/app/(site)/fetch.ts
BrettJephson: Good point. I was thinking I'd encapsulate it in the function but it definitely makes the function signature confusing.

PR #2531, opened 2024-10-17
Add support for reusable content

emmerich:
GitBook recently added reusable content to its editor. This PR renders that reusable content correctly in GBO.


packages/gitbook/src/lib/references.tsx
```diff
@@ -35,6 +36,8 @@ export interface ResolvedContentRef {
     active: boolean;
     /** File, if the reference is a file */
     file?: RevisionFile;
+    /** ID of an attached document. */
+    documentId?: string;
```
SamyPesse, line 40: Is it not better to do:

```suggestion
    reusableContent?: RevisionReusableContent;
```

otherwise it's unclear which case will define a document

SamyPesse (approved):

PR #2534, opened 2024-10-18
Improve perf perception and fix flickering in Ask answer

SamyPesse:
This PR aims at improving the perception of loading times when the page contains a lot of blocks and mainly images:

- Don't show a skeleton for all the first blocks we consider in the viewport
- Improve image component to suspend rendering less and have next render it directly as much as possible

It also improves the "Ask" feature by fixing RND-4369

### DIFF
--- a/packages/gitbook/src/components/DocumentView/Blocks.tsx
+++ b/packages/gitbook/src/components/DocumentView/Blocks.tsx
@@ -55,21 +55,29 @@ export function UnwrappedBlocks<TBlock extends DocumentBlock>(props: UnwrappedBl
     return (
         <>
             {nodes.map((node) => {
-                isOffscreen = isOffscreen || isBlockOffscreen({ document: props.document, block: node, ancestorBlocks: props.ancestorBlocks });
-                
-                return <Block
-                    key={node.key}
-                    block={node}
-                    style={[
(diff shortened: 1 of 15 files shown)


packages/gitbook/src/components/DocumentView/Blocks.tsx
```diff
@@ -49,22 +50,35 @@ type UnwrappedBlocksProps<TBlock extends DocumentBlock> = DocumentContextProps &
...
-                />
-            ))}
+            {nodes.map((node) => {
+                isOffscreen =
+                    isOffscreen ||
+                    isBlockOffscreen({
+                        document: props.document,
+                        block: node,
+                        ancestorBlocks: props.ancestorBlocks,
+                    });
```
taranvohra, line 64: I get that you're avoiding to recompute if true but are all blocks unwrapped at the same time? can it not change?

taranvohra (approved):

packages/gitbook/src/components/DocumentView/Blocks.tsx
SamyPesse: The blocks are in order, so if a block is offscreen, the next one will be as well.

PR #2536, opened 2024-10-18
Fix CLS and others

gregberge:
- **Inverse layout to fix CLS**
- **Remove useless async**
- **Type buttons**

### DIFF
--- a/packages/gitbook/src/app/(site)/(content)/[[...pathname]]/page.tsx
+++ b/packages/gitbook/src/app/(site)/(content)/[[...pathname]]/page.tsx
@@ -78,7 +78,7 @@ export default async function Page(props: {
                 <PageCover as="full" page={page} cover={page.cover} context={contentRefContext} />
             ) : null}
             {/* We use a flex row reverse to render the aside first because the page is streamed. */}
-            <div className="flex flex-row-reverse">
+            <div className="flex flex-row-reverse justify-end">
                 {page.layout.outline ? (
                     <PageAside
                         space={space}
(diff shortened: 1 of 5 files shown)


packages/gitbook/src/app/(site)/(content)/[[...pathname]]/page.tsx
```diff
@@ -77,20 +77,7 @@ export default async function Page(props: {
...
-                    customization={customization}
-                    context={contentRefContext}
-                    page={page}
-                    document={document}
-                    withPageFeedback={
-                        // Display the page feedback in the page footer if the aside is not visible
-                        withPageFeedback && !page.layout.outline
-                    }
-                />
+            <div className="flex flex-row-reverse">
```
Comment (SamyPesse) [inline packages/gitbook/src/app/(site)/(content)/[[...pathname]]/page.tsx:80]: Can you add a comment to explain why? otherwise someone will simplify this and remove it

SamyPesse (approved):

packages/icons/src/Icon.tsx
```diff
@@ -52,10 +52,20 @@ export const Icon = React.forwardRef(function Icon(
...
     const url = getIconAssetURL(context, iconStyle, icon);
+    const [loaded, setLoaded] = React.useState(false);
+
+    React.useEffect(() => {
+        const img = new Image();
+        img.src = url;
+        img.onload = () => {
+            setLoaded(true);
+        };
+    }, [url]);
```
SamyPesse, line 63: Why do we do this? can you add a quick comment 

It looks like it's for visual tests. Can it impact perfs on prod? I guess the browser only loads it once?

SamyPesse (approved):

packages/icons/src/Icon.tsx
gregberge: Yeah it's for visual test, I think there is no impact on performance. Because it's in cache. But it still a double render of icon, I will see if I can do better.

SamyPesse: Before taking the screenshot, can we wait for all requests to be done?

gregberge: It's not recommended but yeah we could try that.

PR #2543, opened 2024-10-22
Update header styling of sections, variant selector, and button links

zenoachtig:
- Change position of variant selector depending on context
  - By default: shown next to logo
  - With sections: shown in table of contents
  - With sidebar-only view: shown in table of contents
  - Without sidebar: shown next to logo
- Update styling of variant selector, remove `kind` prop since we’re always showing the same styling now.
- Update section tab styling and animation
- Make header buttons smaller with a new `medium` button size

## Preview
Screenshot 2024-10-22 at 14 25 20

Screenshot 2024-10-22 at 14 27 14

### DIFF
--- a/packages/gitbook/src/components/SpaceLayout/SpaceLayout.tsx
+++ b/packages/gitbook/src/components/SpaceLayout/SpaceLayout.tsx
@@ -61,6 +61,7 @@ export function SpaceLayout(props: {
     };
 
     const withSections = Boolean(sections && sections.list.length > 0);
+    const withVariants = Boolean(site && spaces.length > 1);
     const headerOffset = { sectionsHeader: withSections, topHeader: withTopHeader };
 
     return (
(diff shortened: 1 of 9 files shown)


packages/gitbook/src/components/SpaceLayout/SpaceLayout.tsx
```diff
@@ -116,6 +105,11 @@ export function SpaceLayout(props: {
                                 />
                             )
                         }
+                        innerHeader={
+                            sections || (!withTopHeader && site && spaces.length > 1) ? (
```
taranvohra, line 109: can you extract it into a readable variable. Otherwise it's hard to read just by glancing over it

packages/gitbook/src/components/TableOfContents/TableOfContents.tsx
```diff
@@ -35,8 +35,10 @@ export function TableOfContents(props: {
     ancestors: Array<RevisionPageDocument | RevisionPageGroup>;
     header?: React.ReactNode;
     headerOffset: { sectionsHeader: boolean; topHeader: boolean };
+    innerHeader?: React.ReactNode;
```
taranvohra, line 38: If we can put a comment on what's the difference between `header` and `innerHeader` :)

taranvohra (approved):
Great!!

PR #2545, opened 2024-10-23
Use transparent bg for secondary button

gregberge:


SamyPesse (commented):
Can you add some visual tests for all of this in https://github.com/GitbookIO/gitbook/blob/main/packages/gitbook/e2e/pages.spec.ts#L436?

gregberge:
> Can you add some visual tests for all of this in https://github.com/GitbookIO/gitbook/blob/main/packages/gitbook/e2e/pages.spec.ts#L436?

I thought I added some.

taranvohra (approved):
@gregberge can you also remove the `@ts-ignore` from those E2E tests now that we are on the latest ver of the API

PR #2546, opened 2024-10-23
Remove deprecated synced block from GitBook Open

emmerich:
Synced blocks are no longer supported in GitBook, they've been replaced by reusable content. This PR removes the old code.


packages/gitbook/src/components/DocumentView/Block.tsx
```diff
@@ -107,8 +106,6 @@ export function Block<T extends DocumentBlock>(props: BlockProps<T>) {
                 throw new Error('Blocks should be directly rendered by parent');
             case 'integration':
                 return <IntegrationBlock {...props} block={block} />;
-            case 'synced-block':
```
scazan, line 110: have we indexed this and potentially need to display it in search results?

clairechabas (approved):

PR #2547, opened 2024-10-23
Add siteSpaceId to Page View Event

scazan:
This adds the siteSpaceID into the tracking event for better reporting while removing the space tracking since it is no longer relevant (spaceIds are extracted from the siteSpaceID instead)


scazan:
This will be handled in a separate PR.

PR #2552, opened 2024-10-24
Unify section tab styles with page item styles

taranvohra:
### Before
image

### After
image


zenoachtig (approved):
Much better! Thanks 🙏

PR #2555, opened 2024-10-24
Bump @scalar/api-client-react version

amritk:
Now that we are on the current version, will be a lot more updates!


taranvohra (approved):
`1.0.65` has an issue where people can't select anything on the docs

PR #2562, opened 2024-10-30
Nested list support

gregberge:
**Before**
CleanShot 2024-10-30 at 15 27 19@2x

**After**
CleanShot 2024-10-30 at 15 27 44@2x

**Heading in list before**
CleanShot 2024-10-30 at 15 46 51@2x

**Heading in list after**
CleanShot 2024-10-30 at 15 48 03@2x


packages/gitbook/src/components/DocumentView/ListItem.tsx
```diff
@@ -1,114 +1,231 @@
...
+        // Use roman numbers
+        case 2: {
+            return `${toRoman(value).toLowerCase()}.`;
+        }
+        default:
+            return '•';
+    }
+}
+
+function toRoman(input: number): string {
```
BrettJephson, line 206: There's probably an npm module for this.

BrettJephson (approved):
Looks good to me!

packages/gitbook/src/components/DocumentView/ListItem.tsx
gregberge: Yep but it's fine like this.

PR #2568, opened 2024-11-06
Fix an issue where the active site section indicator appeared above any dropdowns.

emmerich:


packages/gitbook/src/components/Header/Dropdown.tsx
```diff
@@ -44,7 +44,6 @@ export function Dropdown<E extends HTMLElement>(props: {
                     'absolute',
                     'top-full',
                     'left-0',
-                    'z-20',
```
emmerich, line 47: It wasn't needed - this is an absolute element so its z-index is not relevant to the application (creates a new stack). zindex must be set higher.

PR #2577, opened 2024-11-21
Resolve site level redirects 

taranvohra:
This PR adds support for resolving site redirects configured on a site. The order is as follows
- First site content is always resolved to its canonical URL 
- Then we lookup the page using the `pathname`
  a.  If we find the page, good enough
  b.  If we don't find the page we check if there is a redirect at content (space) level 
  c. If we still don't find the page we check if there is a redirect at site level

Between (b) & (c), I suggest we do (b) first because content level redirects existed before.


packages/gitbook/src/app/(site)/fetch.ts
```diff
@@ -88,20 +104,26 @@ async function resolvePage(
...
+
+        // If a page still can't be found, we try with the API, in case we have a redirect at site level.
+        const resolvedSiteRedirect = await getSiteRedirectBySource({
+            organizationId,
+            siteId,
+            source: rawPathname.startsWith('/') ? rawPathname : `/${rawPathname}`,
+            siteShareKey: input.shareKey,
+        });
+        if (resolvedSiteRedirect) {
+            return redirect(resolvedSiteRedirect.target);
```
taranvohra, line 125: Using this [redirect](https://nextjs.org/docs/app/building-your-application/routing/redirecting#redirect-function) function returns a `307` which is fine for now but we'll need to think if we let users configure custom status codes for their redirects

packages/gitbook/src/app/(site)/fetch.ts
```diff
@@ -88,20 +104,26 @@ async function resolvePage(
...
+        const resolved = await getRevisionPageByPath(spaceId, revisionId, rawPathname);
         if (resolved) {
             return resolvePageId(pages, resolved.id);
         }
+
+        // If a page still can't be found, we try with the API, in case we have a redirect at site level.
+        const resolvedSiteRedirect = await getSiteRedirectBySource({
+            organizationId,
+            siteId,
+            source: rawPathname.startsWith('/') ? rawPathname : `/${rawPathname}`,
```
SamyPesse, line 121: The API should be improved to handle this and normalize input

packages/gitbook/src/app/(site)/fetch.ts
SamyPesse: We should keep everything as temporary redirect for now.

SamyPesse (approved):

packages/gitbook/src/app/(site)/fetch.ts
taranvohra: Sure 👍

packages/gitbook/src/app/(site)/fetch.ts
gregberge: It's not a 307, it's a redirect done client-side. Same issue as 404.  https://github.com/vercel/next.js/blob/5b7833e3da2b22e45e5b13c7001d4d728d093133/packages/next/src/client/components/redirect-boundary.tsx#L28-L37

PR #2580, opened 2024-11-22
Get the best possible target space for content refs

taranvohra:
Fixes RND-5602

### DIFF
--- a/packages/gitbook/e2e/pages.spec.ts
+++ b/packages/gitbook/e2e/pages.spec.ts
@@ -492,6 +492,42 @@ const testCases: TestsCase[] = [
         ],
     },
     {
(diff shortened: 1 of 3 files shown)


SamyPesse:
Does this solution work in the context of a content-ref to a page in another space (that is published in 2 sites)?

Are we computing the url of the page from the space url or are we using the published url on the page itself from the API?

taranvohra:
> Does this solution work in the context of a content-ref to a page in another space (that is published in 2 sites)?
> 
> Are we computing the url of the page from the space url or are we using the published url on the page itself from the API?

I only changed the part of determining the target space to use. Everything else is unchanged on how we further compute the url

PR #2582, opened 2024-11-22
Serve the correct status code for not found and redirects 👌

gregberge:
By reading the code, we are already fetching things in `generateMetadata` that is awaited before the first streaming chunk is emitted. So actually the streaming is not active when we load the first page.

Giving that, the only issue to solve was to enable streaming when we switch to a page dynamically (from the sidebar). By wrapping page content into `React.Suspense` it's now working.

So this PR gives exactly the same result as before but it provides the good status code and redirect code.

Also refactor the code to mutualize things better.

--- 

This does not work because Next.js is waiting for `generateMetadata` if there is no `loading.tsx` present. The problem is that streaming and instant loading are activated by the same thing: `loading.tsx`.

I am now trying to use the parallel routing to put a `loading.tsx` only for nav routing.

--- 

Solve RND-4299


packages/gitbook/src/app/(site)/(content)/[[...pathname]]/page.tsx
```diff
@@ -168,27 +169,38 @@ export async function generateMetadata({
 /**
  * Fetches the page data matching the requested pathname and fallback to root page when page is not found.
  */
-async function getPageDataWithFallback(args: {
-    pagePathParams: PagePathParams;
-    searchParams: { fallback?: string };
-    redirectOnFallback?: boolean;
```
Comment (gregberge) [inline packages/gitbook/src/app/(site)/(content)/[[...pathname]]/page.tsx:174]: Not sure why it as `true` for `Page` and false for `generatePageMetadata`, I don't see a reason for that. We can bring it back of course if it makes sense.

Comment (SamyPesse) [reply packages/gitbook/src/app/(site)/(content)/[[...pathname]]/page.tsx:174]: I remember there was a reason, but it should have been documented with a comment...

Let's add it back instead of taking risks, or you need to try the different cases.

maybe @taranvohra @spastorelli remember?

packages/gitbook/src/app/(site)/(content)/(.)[[...pathname]]/page.tsx
Comment (taranvohra) [inline packages/gitbook/src/app/(site)/(content)/(.)[[...pathname]]/page.tsx:1]: What does the file path mean now?

Comment (gregberge) [reply packages/gitbook/src/app/(site)/(content)/(.)[[...pathname]]/page.tsx:1]: https://nextjs.org/docs/14/app/building-your-application/routing/parallel-routes

Comment (gregberge) [reply packages/gitbook/src/app/(site)/(content)/(.)[[...pathname]]/page.tsx:1]: File based routing 🤪

Comment (SamyPesse) [reply packages/gitbook/src/app/(site)/(content)/(.)[[...pathname]]/page.tsx:1]: Can you comment on the file to explain why we have it?

packages/gitbook/src/app/(site)/(content)/(.)[[...pathname]]/loading.tsx
Comment (SamyPesse) [inline packages/gitbook/src/app/(site)/(content)/(.)[[...pathname]]/loading.tsx:1]: The main difference between the `loading.tsx` and the Suspense wrapping is that now it'll only suspend when fetching the document and inside of the page.

If I access a page cached in the CDN where the revision is not cached in the data cache at the edge, clicking on a next page will not show anything until the Next server app fetches the revision.

Basically order before:

1. Show `loading.tsx` / STREAMING STARTS
2. Fetch revision (often cached)
3. Fetch document
4. Fetch images, etc

Now the order is:

1. Fetch revision
2. Show Suspense boundary / STREAMING STARTS
3. Fetch document
4. Fetch images, etc

The main reason is `generateMetadata` where the data needs to be fetched to get the page title for the new page.

Comment (SamyPesse) [reply packages/gitbook/src/app/(site)/(content)/(.)[[...pathname]]/loading.tsx:1]: It's a tradeoff, we can accept

SamyPesse (commented):
I want to run more tests before approving (of different caching scenarios).

But the direction looks like a good tradeoff.

packages/gitbook/src/app/(site)/(content)/(.)[[...pathname]]/loading.tsx
Comment (gregberge) [reply packages/gitbook/src/app/(site)/(content)/(.)[[...pathname]]/loading.tsx:1]: Steaming starts but no chunk is emitted, so from a user point of view, it does not change anything.

Comment (gregberge) [reply packages/gitbook/src/app/(site)/(content)/(.)[[...pathname]]/loading.tsx:1]: In `generateMetadata` the same function was called. So it was blocking the first chunk anyway.

Comment (SamyPesse) [reply packages/gitbook/src/app/(site)/(content)/(.)[[...pathname]]/loading.tsx:1]: My point is that before your change, `loading.tsx` before `generateMetadata` is called

packages/gitbook/src/app/(site)/(content)/[[...pathname]]/page.tsx
```diff
@@ -128,26 +144,28 @@ export async function generateViewport({ params }: { params: PagePathParams }):
...
+}
 
-    if (!page) {
-        notFound();
+export async function generateMetadata(props: PageProps): Promise<Metadata> {
+    const headerList = await headers();
+    const fetchMode = headerList.get('sec-fetch-mode');
+
+    // We only generate metadata in navigation mode. Else we let the browser handle it.
+    if (fetchMode !== 'navigate') {
```
Comment (SamyPesse) [inline packages/gitbook/src/app/(site)/(content)/[[...pathname]]/page.tsx:159]: Are we sure bots for SEO will pass this header with this value?

Otherwise we'll lose all metadata

Comment (SamyPesse) [reply packages/gitbook/src/app/(site)/(content)/[[...pathname]]/page.tsx:159]: Can you reverse the condition to be only equal to the case where we don't want it?

SamyPesse (changes requested):

packages/gitbook/src/app/(site)/(content)/[[...pathname]]/page.tsx
```diff
@@ -183,6 +180,16 @@ export async function generateMetadata(props: PageProps): Promise<Metadata> {
...
 }
 
+/**
+ * Check if the navigation is in-app, meaning the user clicks on a link.
+ */
+async function checkIsInAppNavigation() {
+    const headerList = await headers();
+    const fetchMode = headerList.get('sec-fetch-mode');
+
+    return fetchMode === 'cors';
```
Comment (SamyPesse) [inline packages/gitbook/src/app/(site)/(content)/[[...pathname]]/page.tsx:190]: There is no other internal next headers sent that we could check to avoid using a web standard one for this?

Comment (gregberge) [reply packages/gitbook/src/app/(site)/(content)/[[...pathname]]/page.tsx:190]: Nop there's nothing.

PR #2583, opened 2024-11-24
Update shiki

gregberge:


packages/gitbook/src/components/DocumentView/CodeBlock/CodeBlock.tsx
```diff
@@ -15,8 +15,7 @@ import './theme.css';
  */
 export async function CodeBlock(props: BlockProps<DocumentBlockCode>) {
     const { block, document, style, context } = props;
-    const withHighlighting = context.shouldHighlightCode(context.content?.spaceId);
```
SamyPesse, line 18: Why remove this?
The goal was to prevent some spaces from crashing where highlighting is currently inefficient.

SamyPesse: If we remove it, we should remove all of it (the context, the function to create it, the list of spaces)

gregberge: Because it does not crash:  https://db308920.gitbook-open.pages.dev/gaowanlu.gitbook.io/note/c++-ji-chu/c++/di-2-zhang-bian-liang-he-ji-ben-lei-xing

PR #2586, opened 2024-11-26
Fix canonical URL ending with "/"

gregberge:


packages/gitbook/src/lib/links.ts
```diff
@@ -82,7 +82,9 @@ export function baseUrl(): string {
  */
 export function absoluteHref(href: string, withHost: boolean = false): string {
     const base = withHost ? baseUrl() : basePath();
-    return `${base}${href.startsWith('/') ? href.slice(1) : href}`;
+    const absoluteHref = `${base}${href.startsWith('/') ? href.slice(1) : href}`;
+    // Trim trailing slashes
+    return absoluteHref.replace(/\/$/, '');
```
SamyPesse, line 87: Screenshot 2024-11-26 at 17 19 26

We pass hrefs to `absoluteHref` (meaning path + query params); but this logic is checking the whole thing as a path.

It's not very clean and error prone. We should improve this code to handle the fact that it's not just a path

gregberge: Sorry I made a safer approach for that, only doing this in canonical.

packages/gitbook/src/app/(site)/(content)/[[...pathname]]/page.tsx
```diff
@@ -165,7 +165,7 @@ export async function generateMetadata(props: PageProps): Promise<Metadata> {
         title: getTitle(data),
         description: page.description ?? '',
         alternates: {
-            canonical: absoluteHref(getPagePath(pages, page), true),
+            canonical: absoluteHref(getPagePath(pages, page), true).replace(/\/+$/g, ''),
```
Comment (SamyPesse) [inline packages/gitbook/src/app/(site)/(content)/[[...pathname]]/page.tsx:168]: In fact I think the fix should be in `absoluteHref`, but it's just the case where `getPagePath` returns an empty string, then we can append it to `{host}/`.

SamyPesse (approved):
Feel free to merge like this to unblock, even if it looks lost and we'll probably soon forget why it's here

gregberge:
> Feel free to merge like this to unblock, even if it looks lost and we'll probably soon forget why it's here

I will comment it.

packages/gitbook/src/app/(site)/(content)/[[...pathname]]/page.tsx
Comment (gregberge) [reply packages/gitbook/src/app/(site)/(content)/[[...pathname]]/page.tsx]: Yeah, sorry I didn't follow the whole thing. But by analyzing deeply it only happens on the root URL because `basePath()` and `baseUrl()` ends with a "/". So I fixed it here as you suggested.

PR #2590, opened 2024-11-27
Fix hint block decoration color

zenoachtig:
Links inside hint blocks correctly change their color to match the hint block, but the `text-decoration` property is not changed with them. Let’s fix that by making them inherit their text color properly.

## Preview
Screenshot 2024-11-27 at 20 59 41


BrettJephson (approved):
Looks good. 

Do we not want to keep the slightly lighter (or semi-transparent) effect on the text decoration?

zenoachtig:
> Looks good.
> 
> Do we not want to keep the slightly lighter (or semi-transparent) effect on the text decoration?

Yeah upon second look you’re right, I’ve added the semi-transparent decoration style to the hint blocks after all. The default one inherits from the default link style, which includes it.

PR #2593, opened 2024-11-28
Add an icon to site sections

BrettJephson:
This PR adds site section icons. 

Screenshot 2024-11-29 at 14 51 21


packages/gitbook/src/components/SiteSectionTabs/SectionIcon.tsx
```diff
@@ -0,0 +1,15 @@
...
+import { type ClassValue, tcls } from '@/lib/tailwind';
+
+/**
+ * Icon shown beside a section in the site section tabs.
+ */
+export function SectionIcon(props: { section: SiteSection; className?: ClassValue }) {
+    const { section, className } = props;
+
+    return section.icon ? (
+        <Icon icon={section.icon as IconName} className={tcls('size-[1em]', className)} />
```
taranvohra, line 13: I feel we don't need to check it again if we do assertion already in the parent. This component should maybe only accept `icon` prop of the section

taranvohra (approved):

PR #2596, opened 2024-12-03
Support dropdown without link

gregberge:
CleanShot 2024-12-03 at 16 32 52@2x

### DIFF
--- a/packages/gitbook/src/components/Header/HeaderLink.tsx
+++ b/packages/gitbook/src/components/Header/HeaderLink.tsx
@@ -1,14 +1,19 @@
 import {
     CustomizationContentLink,
-    CustomizationHeaderLink,
     CustomizationSettings,
     CustomizationHeaderPreset,
     SiteCustomizationSettings,
+    CustomizationHeaderLink,
 } from '@gitbook/api';
 import assertNever from 'assert-never';
 
 import { ContentRefContext, resolveContentRef } from '@/lib/references';
 import { tcls } from '@/lib/tailwind';
+
+// @TODO replace by api.CustomizationHeaderItem when available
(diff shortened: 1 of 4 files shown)


packages/gitbook/src/components/Header/HeaderLink.tsx
```diff
@@ -52,20 +65,20 @@ export async function HeaderLink(props: {
...
                 return (
                     <Button
-                        href={target.href}
+                        href={target?.href}
                         variant={variant}
                         size="medium"
                         className={tcls(
                             {
-                                'button-primary':
+                                ['button-primary']:
```
scazan, line 73: just curious, why the change with the brackets?

packages/gitbook/src/components/Header/HeaderLink.tsx
```diff
@@ -52,20 +65,20 @@ export async function HeaderLink(props: {
                 })();
                 return (
                     <Button
-                        href={target.href}
+                        href={target?.href}
```
scazan, line 68: Can this result in a "dead button" since we don't pass an onClick? should we check for it?

scazan (approved):
left a comment but looks good.

packages/gitbook/src/components/Header/HeaderLink.tsx
gregberge: It was just a change and another change.

packages/gitbook/src/components/Header/HeaderLink.tsx
gregberge: I simplified the whole logic @scazan

scazan: looks great 👍

PR #2599, opened 2024-12-05
Fix cookie banner button being non responsive

valentin0h:
We missed removing `onClick` in https://github.com/GitbookIO/gitbook/pull/2596

### DIFF
--- a/packages/gitbook/src/components/primitives/Button.tsx
+++ b/packages/gitbook/src/components/primitives/Button.tsx
@@ -15,7 +15,6 @@ type ButtonProps = {
 
 export function Button({
     href,
-    onClick,
     children,
     variant = 'primary',
     size = 'default',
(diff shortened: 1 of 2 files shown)


packages/gitbook/src/components/primitives/Button.tsx
```diff
@@ -15,7 +15,6 @@ type ButtonProps = {
 
 export function Button({
     href,
-    onClick,
```
taranvohra, line 18: Can you add an E2E test for cookie banner accept?

valentin0h: I can, but it'll slow down the "hotfix".

gregberge (approved):

taranvohra (approved):

PR #2605, opened 2024-12-06
Add tint color

zenoachtig:
This PR adds support for the new tint colour customisation setting. It is meant as an additional way to theme your docs by subtly colouring various elements of your theme, such as the background or header. The `primary-color` remains in use for primary navigation elements such as links and buttons.

This first PR aims to swap the properties used in colour calculations _without_ introducing big visual changes. In future PRs will propose new ways of using tint in various places. There are no changes for sites that have not opted into the Tint setting.

- Create `tint-color` and `contrast-tint` CSS variables
- Use `tint-color` to set base colour palette (`light-*` and `dark-*` variables)
- Use `tint-color` for the header colour of Bold. When Custom is used, the set "Header background" and "Link color” settings take precedence over tint colour.
- Deprecate `plain-background` Tailwind variant in favour of `tint` and `no-tint` variants.
- Deprecate `primary-base` variable in favour of `tint-color`

# Preview
- **Without tint**: same as current `Background: plain`
- **Primary tint**: same as current `Background: match`
- **Custom tint**: new functionality

## Default
No changes with tint off. If tint is enabled, we colour the bg slightly with the tint color (primary color or custom color)

Default, without tint
Default, primary tint

## Bold
No changes with tint off. If tint is enabled, we colour the bg slightly and header strongly with the tint color. The links are always set to a contrast of the tint color.

Bold, without tint
Bold, primary tint
Bold, custom tint

## Contrast
No changes with tint off. If tint is enabled, we colour only the bg slightly. Header stays contrasting.

Contrast, without tint
Contrast, primary tint

## Custom
No changes with tint off – we use the custom-set `headerBackgroundColor` and `headerLinkColor`. If tint is enabled, we colour only the bg slightly. The custom header colours stay in place. We aim to eventually clean up these options, favouring tint over custom-set header colours.

Custom, without tint
Custom, primary tint
Custom, custom tint

---

_Resolves RND-5670_


packages/gitbook/src/lib/api.ts
```diff
@@ -1204,19 +1204,19 @@ async function getAll<T, E>(
...
     const headersList = headers();
     const extend = headersList.get('x-gitbook-customization');
     if (extend) {
         try {
-            const parsed = rison.decode_object<DeepPartial<T>>(extend);
+            const parsedSettings = rison.decode_object<SiteCustomizationSettings>(extend);
 
-            // Merge objects and some properties deep
-            return mergeDeepPlainObject(raw, parsed, ['styling', 'themes']);
+            return parsedSettings;
```
valentin0h, line 1219: Expects the full settings object in `x-gitbook-customization`

valentin0h (approved):

PR #2606, opened 2024-12-06
Adapt GBO cache logic to support users sessions info

spastorelli:


packages/gitbook/src/lib/api.ts
```diff
@@ -87,21 +87,41 @@ const cacheTtl_1day = {
     ttl: 24 * 60 * 60,
 };
 
-const apiSyncStorage = new AsyncLocalStorage<GitBookAPI>();
+const apiSyncStorage = new AsyncLocalStorage<GitBookAPIWithContextKey>();
 
 export const DEFAULT_API_ENDPOINT = process.env.GITBOOK_API_URL ?? 'https://api.gitbook.com';
 
+export class GitBookAPIWithContextKey extends GitBookAPI {
```
SamyPesse, line 94: Do we need a class for this compared to just extending the type (`GitBookAPI & { comtextKey: string; }`)?

packages/gitbook/src/lib/api.ts
```diff
@@ -87,21 +87,41 @@ const cacheTtl_1day = {
...
+    get contextKey() {
+        return this.#contextKey;
+    }
+}
+
 /**
  * Create a new API client with a token.
  */
-export function apiWithToken(apiToken: string): GitBookAPI {
+export function apiWithToken(apiToken: string, contextKey?: string): GitBookAPIWithContextKey {
```
SamyPesse, line 117:
```suggestion
export function apiWithToken(apiToken: string, contextKey: string | undefined): GitBookAPIWithContextKey {
```

Optional argument are easy to miss, we should avoid them

packages/gitbook/src/lib/cache/cache.ts
```diff
@@ -52,6 +52,9 @@ export interface CacheDefinition<Args extends any[], Result> {
     /** Tag to associate to the entry */
     tag?: (...args: Args) => string;
 
+    /** Context key matching the hash of the claims included in the visitor JWT token */
+    getContextKey?: () => string | undefined;
```
SamyPesse, line 56: This comment and even the function name is too specific to a usage that `cache.ts` should not know about.

packages/gitbook/src/lib/api.ts
spastorelli: Unless I'm missing something, yes I think. Because typescript would complain that returning `{ ...gitbookAPIClient, contextKey }` would not match `GitBookAPI & { comtextKey: string; }` type.

The other approach I considered was to do `Object.assign(gitbookAPIClient, { contextKey });` but I wasn't sure about it as it breaks the encapulation of `contextKey` prop by making it writeable and not read only prop anymore.

Can't think of any other way to get around it.

packages/gitbook/src/lib/cache/cache.ts
spastorelli: Renamed to `getKeySuffix` but happy to change it if you don't think that works. Other name I was thinking was `getDiscriminatorKey`. Otherwise if you have any other suggestions.

packages/gitbook/src/lib/api.ts
SamyPesse: Hum 🤔  it just feels ugly to create a huge class just for this, with such a name.

Maybe we solve it with:

```ts
type GitBookAPIContext = {
   client: GitBookAPI;
   contextKey: string | undefined;
}
```

and we just need to replace `api()` by `api().client`

It's probably cleaner

packages/gitbook/src/middleware.ts
```diff
@@ -470,10 +477,12 @@ async function lookupSiteOrSpaceInMultiIdMode(
         throw new Error('Collection is not supported in multi-id mode');
     }
 
-    const gitbookAPI = new GitBookAPI({
+    const contextKey = decoded.claims ? hash(decoded.claims) : undefined;
```
SamyPesse, line 480: Are these claims going to contain some useless props (`iat`, etc)?

We don't want to bust the cache on useless things.

Also this condition will always be true for VA; causing the cache to be busted for all VA website.

```suggestion
    const contextKey = decoded.claims && Object.keys(decoded.claims).length > 0 ? hash(decoded.claims) : undefined;
```

SamyPesse: Also generally I don't understand why we are computing our own hash here; while at line 158 we use the one from the published content resolution?

spastorelli: > Also generally I don't understand why we are computing our own hash here; while at line 158 we use the one from the published content resolution?

This part is for `lookupSiteOrSpaceInMultiIdMode` (e.g preview) which I don't think it calls `getPublishedContentByURL` (the one that returns the contextID) which is way I'm doing it manually based on the "preview" token provided in the url. 

> Are these claims going to contain some useless props (iat, etc)?
>
> We don't want to bust the cache on useless things.
> 
> Also this condition will always be true for VA; causing the cache to be busted for all VA website.

This logic applies only `lookupSiteOrSpaceInMultiIdMode` which I think doesn't handle VA.

SamyPesse: Ha yes it makes sense 👍 

So yeah we need a hash to ensure we can use the preview for adaptive content

spastorelli: Changed it to pull the context from the URL intead of parsing it from the token claims. I'll make in parallel the changes to the getSitePublishingPreviewById API endpoint to return the preview URL with the context hash when it was provided claims.

spastorelli: It's for preview this mode no? Also I know but other than pulling it from the token and computing a hash in GBO I don't see any other option.

SamyPesse: I think computing the hash from the token's claims is good. My original comment was missing the context of the why.

packages/gitbook/src/lib/api.ts
```diff
@@ -87,14 +87,19 @@ const cacheTtl_1day = {
     ttl: 24 * 60 * 60,
 };
 
-const apiSyncStorage = new AsyncLocalStorage<GitBookAPI>();
+export type GitBookAPIContext = {
+    client: GitBookAPI;
+    contextKey: string | undefined;
```
SamyPesse, line 92: Add a comment

packages/gitbook/src/middleware.ts
```diff
@@ -153,13 +157,17 @@ export async function middleware(request: NextRequest) {
     // Resolution might have changed the API endpoint
     apiEndpoint = resolved.apiEndpoint ?? apiEndpoint;
 
+    const contextKey = 'site' in resolved ? resolved.contextId : undefined;
```
SamyPesse, line 160: We should align the nameing, change it here to be `contextId`

packages/gitbook/src/middleware.ts
```diff
@@ -470,22 +481,29 @@ async function lookupSiteOrSpaceInMultiIdMode(
         throw new Error('Collection is not supported in multi-id mode');
     }
 
+    const decodedClaims =
```
SamyPesse, line 484: Add a comment to explain when it's used

packages/gitbook/src/middleware.ts
```diff
@@ -470,22 +481,29 @@ async function lookupSiteOrSpaceInMultiIdMode(
...
+        decoded.claims && Object.keys(decoded.claims).length
+            ? sanitizeJWTTokenClaims(decoded.claims)
```
SamyPesse, line 486: To be removed after chat

SamyPesse (approved):

PR #2610, opened 2024-12-09
Pin React versions

gregberge:


valentin0h (approved):
context https://github.com/DefinitelyTyped/DefinitelyTyped/issues/71363

PR #2611, opened 2024-12-09
Allow cells in tables to use more available space

viktorrenkema:
We were wrapping the inner content of a cell with a div and applying text-balance, which doesn't work nicely when a span contains nested elements such as `code`, which is often the case with table cells. 

Before:
CleanShot 2024-12-09 at 11 35 15@2x

After:
CleanShot 2024-12-09 at 11 35 29@2x


packages/gitbook/src/components/DocumentView/Table/RecordRow.tsx
```diff
@@ -33,9 +33,7 @@ export async function RecordRow(
                             'dark:border-light/2',
                         )}
                     >
-                        <div className={tcls('pr-4 ', 'text-balance')}>
                             <RecordColumnValue key={column} {...props} column={column} />
```
clairechabas, line 36: Why remove the `text-balance` too? Seems like the issue was the padding. The balanced wrapping looks like a good addition for a table cell

viktorrenkema: The text balance is actually the problem as mentioned in the pr description. The text-balance is too aggressive, not just when we have nested elements like <code> but in general and I suspect its because of the multiple wrappers that wrap the content after the text-balance div.

Here's what multiple rows with varying setups (pure strings vs nested <code> items) look like with `text-balance`:
CleanShot 2024-12-09 at 11 51 36@2x

When I remove text-balance:
CleanShot 2024-12-09 at 12 13 55@2x

I've tried to figure out what's causing this poor wrapping, and while flattening the tree's layout of children with `display: contents` helps for pure strings, it doesn't when we have inline children:
CleanShot 2024-12-09 at 12 16 40@2x

So i think its safer and the net gain is larger to just not use any text-balance at all, it seems to backfire more than that it helps.

PR #2620, opened 2024-12-11
Fix sublinks display on mobile

gregberge:
CleanShot 2024-12-11 at 14 20 31@2x
CleanShot 2024-12-11 at 14 20 43@2x

### DIFF
--- a/packages/gitbook/src/components/Header/Dropdown.tsx
+++ b/packages/gitbook/src/components/Header/Dropdown.tsx
@@ -39,7 +39,7 @@ export function Dropdown<E extends HTMLElement>(props: {
                 aria-labelledby={dropdownId}
                 className={tcls(
                     'w-52',
-                    'max-h-56',
+                    'max-h-80',
                     'flex',
                     'absolute',
                     'top-full',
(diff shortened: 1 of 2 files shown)


scazan (approved):

packages/gitbook/src/components/Header/Dropdown.tsx
```diff
@@ -111,31 +111,38 @@ export function DropdownMenu(props: { children: React.ReactNode }) {
  * Menu item in a dropdown.
  */
 export function DropdownMenuItem(props: {
-    href: string;
+    href: string | null;
     active?: boolean;
+    className?: ClassValue;
```
clairechabas, line 116: I'm not familiar with the GBO dev env yet : why isn't that a string?

packages/gitbook/src/components/Header/Dropdown.tsx
```diff
@@ -111,31 +111,38 @@ export function DropdownMenu(props: { children: React.ReactNode }) {
...
 }) {
-    const { children, active = false, href } = props;
+    const { children, active = false, href, className } = props;
+
+    if (href) {
+        return (
+            <Link
+                href={href}
+                prefetch={false}
+                className={tcls(
```
clairechabas, line 126: Curious to know why we went for `tcls` instead of `clsx` on GBO?

clairechabas (approved):
much better 👍

packages/gitbook/src/components/Header/Dropdown.tsx
gregberge: It's a mistake, but it's like that from the beginning. At some point we should change it to improve perfs.

packages/gitbook/src/components/Header/Dropdown.tsx
gregberge: We manipulate `clsx` values, but IMO we should not. All components should only accept a `className?: string` and if you need to merge thing you can do it when you pass the prop: `<Any className={clsx(...)} />` like we do in the app.

PR #2622, opened 2024-12-11
Refactor Table to use flex layouts

viktorrenkema:
This PR changes the way we render Tables. It now implements a strategy closer to the GitBook in-app editor, where we use flexbox for layout instead of semantic `<table>` `<tr>` etc, and use `role` on divs for these instead.

The change was motivated by limitations of layout with html table elements. Our columns can be sized in 3 different ways:
     1. Set to auto-size by default, these columns share the available width
     2. Explicitly set by the user by dragging column separator (we then turn off auto-size)
     3. Auto-size is turned off without setting a width, we then default to a fixed width of 100px

This lead to a bunch of layout headaches, like:
- if you have the `<table>` take up `width: 100%` of the available space, columns would always increase in size when possible, even though if we/users explicitly set a custom `width` on a column through dragging the column separator
- we can use `width: fit-content`, which also allows for tables in smaller width than the page (altho [this is currently broken](https://linear.app/gitbook-x/issue/RND-5813/tables-smaller-in-width-than-the-available-space-increase-anyway), but then any columns with no set width should enforce auto-sizing. but this is not possible without measuring the DOM (inefficient), as the parent is set to fit-content, not 100%

Ultimately using flexbox gives us a lot more flexibility and potential for adding more features, and we're closer to the editor.

This PR also fixes a number of bugs:
- https://linear.app/gitbook-x/issue/RND-5813/tables-smaller-in-width-than-the-available-space-increase-anyway
- https://linear.app/gitbook-x/issue/RND-3264/table-columns-sizing-doesnt-match-published-content-vs-editor
- https://linear.app/gitbook-x/issue/RND-5784/full-width-tables-take-more-space-than-needed


packages/gitbook/src/components/DocumentView/Table/ViewGrid.tsx
```diff
@@ -7,159 +8,128 @@ import { TableViewProps } from './Table';
...
+                'grid',
+                'w-full',
+                'overflow-x-auto',
+                'overflow-y-hidden',
+                'mx-auto',
+            )}
         >
-            {/* ProgressScroller: */}
-            <div
-                className={`${styles.progressOpacitySharp} ${tcls(
```
viktorrenkema, line 44: I'm returning this one, but with the new layout model I need to take a closer look + it is a nice to have and not a crucial element to have.

packages/gitbook/src/components/DocumentView/Table/table.module.css
```diff
@@ -1,3 +1,54 @@
+/* Detect whether a scrollbar exists on the table */
+@keyframes detect-scroll {
+    from,
+    to {
+        --can-scroll: ;
+    }
+}
+
+/* Apply styles to the Table if scrollbar exists */
```
viktorrenkema, line 9: Fun trick I learned as I don't think there is any other css selector or query that can be used to select elements with a scrollbar: https://csscade.com/can-you-detect-overflow-with-css/

packages/gitbook/src/components/DocumentView/Table/RecordColumnValue.tsx
```diff
@@ -2,7 +2,7 @@ import { ContentRef, DocumentBlockTable } from '@gitbook/api';
 import { Icon } from '@gitbook/icons';
```
viktorrenkema, line 2: As this is a bigger change, I made sure to test a bunch of different docs with differing table setups. For us internally, here's also a quick before/after comparison of some customer docs shared with us thru support when they raised this issue: https://linear.app/gitbook-x/issue/RND-3264/table-columns-sizing-doesnt-match-published-content-vs-editor#comment-b5f07348

In the cases i could check, this PR fixed it for all.

packages/gitbook/src/components/DocumentView/Table/RecordRow.tsx
```diff
@@ -1,42 +1,53 @@
...
+                    columnWidths: view.columnWidths,
+                    autoSizedColumns,
+                    fixedColumns,
+                });
                 return (
-                    <td
+                    <div
                         key={column}
+                        role="cell"
                         className={tcls(
```
SamyPesse, line 32: Tables have a huge impact on the DOM / HTML size. I'm wondering if we should not use a `table.module.css` for this with:

```
.row {
  @apply flex-1 align-middle ...
}

```

It'll make the class name much smaller

SamyPesse: We should do it for code block lines as well at some point

packages/gitbook/src/components/DocumentView/Table/ViewGrid.tsx
SamyPesse: We don't need it and it impact DOM sizes, performances (for ex for docs.nexthink.com/platform/user-guide/nexthink-query-language-nql/nql-data-model); it looks like your PR makes it faster this way.

SamyPesse (approved):
Looks good, but I'm not a CSS expert

packages/gitbook/src/components/DocumentView/Table/RecordRow.tsx
viktorrenkema: Ah ok nice good to know that table is one place where we can improve performance on. So you mean something like this: https://github.com/GitbookIO/gitbook/pull/2622/commits/2ff10758d13303be7ddf6a9100ae9e2dafb32ce2

I didnt realize applying tw classes in a css class is more efficient/performant than a bunch of individual classNames passed to the component. Is this a common known thing perhaps with a term, so I can read up on it?

PR #2624, opened 2024-12-12
Improve ask/search Enter key press UX

clairechabas:
This PR improves search/ask `Enter` key press handling:
- Always display the `Ask <question>` result item at the top of the results list.
- Auto-focuses the first item in the results list if there's a query and result items so that pressing the `Enter` triggers an AI search unless the user types again or uses arrows/tabs to navigate in results.
- Improves result item highlighting UI from @zenoachtig ✨


gregberge (approved):

packages/gitbook/src/components/Search/SearchModal.tsx
```diff
@@ -182,6 +183,14 @@ function SearchModalBody(
         } else if (event.key === 'Enter') {
             event.preventDefault();
             resultsRef.current?.select();
+
+            const queryIsQuestion = isQuestion(state.query);
+
+            onChangeQuery({
```
zenoachtig, line 189: - We should highlight what action we’re about to perform when pressing Enter, right now I don’t get any feedback on what I’m about to do. We can easily do this by moving the cursor such that either the first result (Ask) or the second result (first search result) is highlighted. I can then move up/down myself to deviate from the default.
- I think we need to always show the Ask AI prompt, right now we still conditionally show it. Even if we don’t think it’s a question, users should have the ability to ask it I think.

gregberge (approved):

packages/gitbook/src/intl/translations/en.ts
```diff
@@ -13,6 +13,7 @@ export const en = {
     search_scope_space: '${1}',
     search_scope_all: 'All content',
     search_ask: 'Ask "${1}"',
+    search_ask_description: 'Let GitBook AI answer your question',
```
emmerich, line 16:
```suggestion
    search_ask_description: 'Ask AI for an answer to your question',
```

I don't think we should mention GitBook here. As these are visitor-facing sites, they carry the branding of the user and the GitBook touch should be as light as possible (keeping it just to the "powered by GitBook" badge).

I'm not sure of the best wording for it, let's discuss internally.

emmerich (changes requested):

packages/gitbook/src/intl/translations/en.ts
clairechabas: That phrasing is a less easy to read. Before changing it since it means updating all translation files I'll ask @stephen-ashby for his copy input

stephen-ashby: Could just be

> Find the answer with AI

Or

> Use AI to answer your question

or even

> Let AI answer your question

clairechabas: How about :

> Ask AI to answer your question

clairechabas: Actually I think 

> Find the answer with AI

is best. 

@emmerich ok with you or another suggestion?

emmerich (commented):
We're just missing the translations of the new copy and we're good to merge.

clairechabas:
@emmerich yes I'm on it along with the other changes you suggested above

packages/gitbook/src/intl/translations/en.ts
clairechabas: Ok going with "Find the answer with AI"

emmerich (approved):

PR #2633, opened 2024-12-16
Fix header logo margin

zenoachtig:
# Before

Screenshot 2024-12-16 at 12 13 29

# After

Screenshot 2024-12-16 at 12 13 32


emmerich (approved):
Please include a changeset with `bun changeset` 🙏

PR #2635, opened 2024-12-16
Pass down ancestors instead of computing them

SamyPesse:
This PR fixes a potential invalid breadcrumb being displayed for a page.


packages/gitbook/src/components/PageBody/PageHeader.tsx
zenoachtig, line 1: As cleanup we can remove the `flattenPages` function at the bottom of this file, and two unused imports (`RevisionPageGroup`, `RevisionPageType`).

zenoachtig (approved):
Thanks for improving 🙏

PR #2637, opened 2024-12-16
Fix closing search modal on Escape key press

clairechabas:
The current escape key press handler is only applied on the search text input. If the focus leaves the input (when user presses arrow or tab keys to navigate in the search results), then pressing the `Escape` key is handled anymore.

This PR moves the closing logic from the `input` to the `body`.


gregberge (approved):
The best solution would be to use our Dialog / Modal system from Spine for this dialog.

But good fix for now 👌

PR #2640, opened 2024-12-17
Fix search input value removing chars while typing

clairechabas:
The search input value randomly and inconsistently sometimes "removes characters" while user is typing a query. 

What seem to be happening is that their seem to some kind of race condition between the global search state which updates while the input value keeps changing. It could be that the input value is changes faster than the global state can update, leading to loosing the last character here and there during typing. This looks like a re-rendering race issue : `state.query` which is used to set the value on the input is updated too late compared to the actual field value which gets overwritten.

The approach in this PR is to switch to using a local state for the input value and debouncing the global state update.

### DIFF
--- a/packages/gitbook/src/components/Search/SearchModal.tsx
+++ b/packages/gitbook/src/components/Search/SearchModal.tsx
@@ -192,13 +192,31 @@ function SearchModalBody(
         }
     };
 
-    const onChange = (event: React.ChangeEvent<HTMLInputElement>) => {
-        onChangeQuery({
-            ask: false, // When typing, we go back to the default search mode
-            query: event.target.value,
-            global: state.global,
-        });
-    };
+    // Mange input value locally to avoid flickering during typing when global state updates
+    const [inputValue, setInputValue] = React.useState(state?.query || '');
+    const timeoutIdRef = React.useRef<number | null>(null);
+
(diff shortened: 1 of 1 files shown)


gregberge (dismissed):
We should understand why and fix the underlying issue. I am not sure this fix is good.

gregberge (approved):

emmerich (changes requested):
Ha we think alike 👍  it's the same approach I started to take when I was looking quickly at the bug. I think it's along the right lines but it needs some deeper thought.

There is shared state between the "Search" state and the "Ask" state. You can see this because the "Ask" state is an atom which is referenced in some of the generic search UI. This PR now stores that state in two places (a local React state in SearchModal and an atom in the parent updated by `onChangeQuery`). Having this state represented in two places is going to cause inconsistencies between the two.

I think the state needs to be brought up a level and managed in the parent. I also think that we are trying to render a streamed server component which is causing the React render loop to go a bit whacky. 

I'd like to tackle this by going one step higher in the component tree and also looking at how we fetch data from the server action - happy to pair on this too.

clairechabas:
Closing this one, after pairing with @gregberge the fix was made here: https://github.com/GitbookIO/gitbook/pull/2643

PR #2641, opened 2024-12-17
Prepare @gitbook/proxy to support serving under sub-directory

SamyPesse:
This PR implements a package `@gitbook/proxy` to easily build a proxy to serve a site under a sub-directory on a marketing website.

It also updates the middleware to support this.


packages/gitbook/src/middleware.ts
```diff
@@ -332,27 +347,36 @@ function getInputURL(request: NextRequest): { url: URL; mode: URLLookupMode } {
         mode = 'multi-id';
     }
 
+    // When passing a x-gitbook-site-url header, this URL is used instead of the request URL
+    // to determine the site to serve.
+    const xGitbookSite = request.headers.get('x-gitbook-site-url');
+    if (xGitbookSite) {
+        mode = 'proxy';
+    }
```
taranvohra, line 355: I think we should be setting the `url` using `xGitbookSite` here instead of doing it in `lookupSiteInProxy` because that lookup function should only be concerned with getting a `mode` and `url` and determine how to best serve the site.

packages/gitbook/src/middleware.ts
```diff
@@ -386,13 +410,31 @@ async function lookupSpaceInSingleMode(url: URL): Promise<LookupResult> {
...
+async function lookupSiteInProxy(request: NextRequest, url: URL): Promise<LookupResult> {
+    const rawSiteUrl = request.headers.get('x-gitbook-site-url');
+    if (!rawSiteUrl) {
+        throw new Error(
+            `Missing x-gitbook-site-url header. It should be passed when using GITBOOK_MODE=proxy.`,
+        );
+    }
+
+    const siteUrl = new URL(rawSiteUrl);
+    siteUrl.pathname = joinPath(siteUrl.pathname, url.pathname);
```
taranvohra, line 426: Like I said above, I think it makes more sense to compute this `siteUrl` in `getInputURL` for `proxy` mode and this function just accepts it. Right now `x-gitbook-site-url` is being utilized in more than 1 place (getInputURL and this) and I don't think that should be the case

taranvohra (approved):

packages/gitbook/src/middleware.ts
SamyPesse: This is what I originally thought, but actually it'll cause trouble as we also use this URL as a representation of where the user currently is, which is not the case.

PR #2653, opened 2024-12-20
Reintro vertical table lines + hover state

viktorrenkema:
- Adds back the vertical lines for tables, got feedback from Snyk that they preferred without. We'll reintroduce the ability to hide these lines as a config on table level instead
- Add hover state over hovered table rows to also improve readability of large tables

Checked with @zenoachtig 

After

https://github.com/user-attachments/assets/325f540b-ac95-4935-8e7b-e2025c42e30e


packages/gitbook/src/components/DocumentView/Table/table.module.css
```diff
@@ -27,33 +27,33 @@
...
+    @apply text-sm font-medium py-2 px-4 text-tint-900;
 }
 
 :global(.dark) .columnHeader {
     @apply text-white;
 }
 
 .row {
-    @apply flex border-tint-700/2;
+    @apply flex border-tint-700/2 hover:bg-tint-700/2 transition-colors;
```
viktorrenkema, line 38: Ideally I'd apply `hover:bg-tint-700/1` `dark:hover:bg-tint-700/2` so the dark mode hover color is different than the light-mode, but tailwind doesn't agree. Open to suggestions, otherwise ill look into it later and fix after.

emmerich (approved):
Don't forget a changeset!

PR #2669, opened 2025-01-04
Fixes #2321. Fix code format overriding inline link styles

### LINKED ISSUE #2321: Links in code font are not highlighted on the generated website
## Bug description

Links in code font are not highlighted on the generated page, so they are indistinguishable from regular text in code font.
However, the links are highlighted in the editor.

Editor:
image

Generated page:
image

## How to reproduce

1. Add a link to the text in code font.
2. Publish the page.

eztaK-red:
Fixes #2321.

Does this by removing the gray font color of inline code blocks, but these were inconsistent with the gitbook editor anyway (plus I don't know how to allow the text to have its own custom color without overriding the inline link font color)

Also removes the inline-flex class name since it was responsible for overriding the underlined decoration of links. Results remain seemingly unchanged after removing it.

Before:
image
(Link in code font are indistinguishable from normal text)

After:
image
(Link in code font is correctly colored and underlined)


gregberge:
reopened in #2673 to avoid issues of forked PR

PR #2671, opened 2025-01-06
Upgrade Next.js to v15

gregberge:
## Todo

- [x] Make all headers and params async
- [x] Patch `decode-named-character-reference` because edge environment uses "browser" field
- [x] Remove Recoil because [not supported in React v19](https://github.com/facebookexperimental/Recoil/issues/2318)
- [x] Update Sentry setup


emmerich (approved):

packages/gitbook/src/app/(site)/(content)/[[...pathname]]/page.tsx
```diff
@@ -20,10 +20,13 @@ export const runtime = 'edge';
  * Fetch and render a page.
  */
 export default async function Page(props: {
```
Comment (emmerich) [inline packages/gitbook/src/app/(site)/(content)/[[...pathname]]/page.tsx:22]: Isn't there a Nextjs type for these props?

Comment (gregberge) [reply packages/gitbook/src/app/(site)/(content)/[[...pathname]]/page.tsx:22]: nop

PR #2675, opened 2025-01-06
Ignore invalid API calls to `getSiteRedirectFromSource` API

gregberge:
Fix GITBOOK-OPEN-1W2E

### DIFF
--- a/packages/gitbook/src/lib/api.ts
+++ b/packages/gitbook/src/lib/api.ts
@@ -677,6 +677,15 @@ export const getSiteRedirectBySource = cache({
             );
             return cacheResponse(response, cacheTtl_1day);
         } catch (error) {
+            // 422 is returned when the source is invalid
+            // we don't want to throw but just return null
+            if ((error as GitBookAPIError).code === 422) {
+                return {
+                    data: null,
+                    ...cacheTtl_1day,
+                };
+            }
+
             if ((error as GitBookAPIError).code === 404) {
                 return {
                     data: null,
--- a/packages/gitbook/src/lib/api.ts
+++ b/packages/gitbook/src/lib/api.ts
@@ -646,6 +646,13 @@ export const getDocument = cache({
 });
 
 /**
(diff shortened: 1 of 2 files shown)


packages/gitbook/src/lib/api.ts
```diff
@@ -677,6 +677,15 @@ export const getSiteRedirectBySource = cache({
             );
             return cacheResponse(response, cacheTtl_1day);
         } catch (error) {
+            // 422 is returned when the source is invalid
+            // we don't want to throw but just return null
+            if ((error as GitBookAPIError).code === 422) {
```
emmerich, line 682: Why is the source invalid?

gregberge: Because it contains invalid character the RegExp is very restrictive.

gregberge: @emmerich I also added a pre-validation to avoid reaching the API if it's useless. We have lot of API calls on that and sometimes it results in a timeout. Let's optimize it.

packages/gitbook/src/lib/api.ts
```diff
@@ -662,6 +669,13 @@ export const getSiteRedirectBySource = cache({
         },
         options: CacheFunctionOptions,
     ) => {
+        // Validate the source to avoid unnecessary API calls.
+        if (!validateSiteRedirectSource(args.source)) {
```
emmerich, line 673: Let's move this check one level higher, as we are already doing some basic validation of the pathname before calling this fn. It'll make exit conditions clearer: https://github.com/GitbookIO/gitbook/blob/99b6c92377f8bcb9f90ef6a255399463317c8b98/packages/gitbook/src/app/(site)/fetch.ts#L108

gregberge: I am not sure of this call to have the same restrictions. That's why I did it as closer as the API call.
CleanShot 2025-01-07 at 10 02 45@2x

emmerich (approved):

PR #2678, opened 2025-01-07
Fix error from visitor id not set

gregberge:
Fix GITBOOK-OPEN-1W86

### DIFF
--- a/packages/gitbook/src/components/Insights/InsightsProvider.tsx
+++ b/packages/gitbook/src/components/Insights/InsightsProvider.tsx
@@ -87,15 +87,6 @@ export function InsightsProvider(props: InsightsProviderProps) {
     }>({});
 
     /**
-     * Get the visitor ID and store it in a ref.
-     */
-    React.useEffect(() => {
-        getVisitorId().then((visitorId) => {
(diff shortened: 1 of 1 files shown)


packages/gitbook/src/components/Insights/InsightsProvider.tsx
```diff
@@ -194,13 +185,19 @@ export function InsightsProvider(props: InsightsProviderProps) {
         },
     );
 
-    // When the page is unloaded, flush all events
+    /**
+     * Get the visitor ID and store it in a ref.
+     */
     React.useEffect(() => {
-        window.addEventListener('beforeunload', flushEventsSync);
+        getVisitorId().then((visitorId) => {
```
emmerich, line 192: could it be that the visitor ID is undefined/empty string? It would also cause the same bug

emmerich (approved):

packages/gitbook/src/components/Insights/InsightsProvider.tsx
gregberge: We will see in Sentry after I merged that. Anyway the logic was wrong here, now it's good.

PR #2684, opened 2025-01-07
Fix search loading state

clairechabas:
Currently the loading state is based on the `results` array being `undefined` which includes the query object meaning the array will be defined way before it's populated with the actual result items, causing the loading state to never show.


emmerich (approved):

packages/gitbook/src/components/Search/SearchResults.tsx
```diff
@@ -175,7 +175,7 @@ export const SearchResults = React.forwardRef(function SearchResults(
         [moveBy, select],
     );
 
-    if (!results) {
+    if (!results || results.filter((result) => result.type !== 'question').length === 0) {
```
emmerich, line 178:
```suggestion
    if (!results?.some((result) => result.type !== 'question')) {
```

.some exit earlier than .filter

emmerich (commented):
Don't forget the changeset :)

packages/gitbook/src/components/Search/SearchResults.tsx
clairechabas: Updated that logic and added the changeset

PR #2692, opened 2025-01-08
Improve scroll listener

gregberge:
Fix GITBOOK-OPEN-1WGY


packages/gitbook/src/components/hooks/useScrollListener.ts
```diff
@@ -0,0 +1,48 @@
...
+ *    console.log('scroll', event);
+ *  }, ref);
+ *
+ *  return <div ref={ref} />;
+ * }
+ */
+export function useScrollListener(
+    listener: (event: Event) => void,
+    elementRef: RefObject<HTMLElement | Window> | null,
+    options?: {
```
emmerich, line 19: Is it needed?

emmerich (approved):

packages/gitbook/src/components/hooks/useScrollListener.ts
```diff
@@ -0,0 +1,48 @@
+import { type RefObject, useEffect, useLayoutEffect, useRef } from 'react';
```
emmerich, line 1: should we add 'use client' ?

packages/gitbook/src/components/hooks/useScrollListener.ts
gregberge: Copied from GitBook project.

gregberge: Removed it.

PR #2694, opened 2025-01-08
Make cookies access safer

gregberge:
Fix GITBOOK-OPEN-1WJE

Fix GITBOOK-OPEN-1WFJ

### DIFF
--- a/packages/gitbook/src/lib/cookies.ts
+++ b/packages/gitbook/src/lib/cookies.ts
@@ -1,0 +1,27 @@
+import cookies from 'js-cookie';
+
+export function getAll(): {
+    [key: string]: string;
+} {
+    try {
+        return cookies.get();
+    } catch (error) {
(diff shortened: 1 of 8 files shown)


packages/gitbook/src/lib/cookies.ts
```diff
@@ -0,0 +1,27 @@
+import cookies from 'js-cookie';
+
+export function getAll(): {
+    [key: string]: string;
+} {
+    try {
+        return cookies.get();
```
emmerich, line 7: Aren't you gonna have to do the same thing here that you do with localStorage where simplify accessing the var will give an error? Or does `js-cookie` handle all that for you?

gregberge: Not from what I see from Sentry, only the security error for now. So I guess it handles it for us.

emmerich (approved):

PR #2697, opened 2025-01-08
Fix Sentry instrumentation (again)

gregberge:


packages/gitbook/sentry.server.config.ts
```diff
@@ -0,0 +1,18 @@
+import { init } from '@sentry/nextjs';
```
emmerich, line 1: Is `.server` needed? 🤔 presumably we're either on the edge or on the client. Otherwise you could just import from `.edge.config.ts`

emmerich (approved):

packages/gitbook/sentry.server.config.ts
gregberge: IDK I just want to be safe. Maybe internally Sentry uses it for something, I really don't know.

PR #2711, opened 2025-01-09
Handle security issue with cookies on Safari

gregberge:
Fix GITBOOK-OPEN-1WQZ


emmerich (approved):
are there any other places we're using js-cookie directly? Let's move them all to the lib

gregberge:
> are there any other places we're using js-cookie directly? Let's move them all to the lib

No there is no normally!

PR #2713, opened 2025-01-09
Fix UI search with no ask

gregberge:
Also mitigate the bug from cache

Fix GITBOOK-OPEN-1WTD

### DIFF
--- a/packages/gitbook/src/components/Search/SearchResults.tsx
+++ b/packages/gitbook/src/components/Search/SearchResults.tsx
@@ -116,6 +116,14 @@ export const SearchResults = React.forwardRef(function SearchResults(
                     return;
                 }
 
+                if (!results) {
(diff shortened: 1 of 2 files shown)


clairechabas (approved):

packages/gitbook/src/components/Search/SearchResults.tsx
```diff
@@ -1,3 +1,4 @@
+import { captureException } from '@sentry/nextjs';
```
emmerich, line 1: isn't it the client package you need?

emmerich (approved):

packages/gitbook/src/components/Search/SearchResults.tsx
gregberge: This is both, client and server package. It's the same one.

PR #2714, opened 2025-01-09
Vertical orientation for sections list on sites without header

zenoachtig:
This PR adds a vertical display component for sections, useful for sites that have the header turned off (and might become an opt-in choice for sites with sections). When there are more than 5 sections we display a scrollview (and scroll to the active section if needed).

# Preview

https://github.com/user-attachments/assets/c15f056a-7b46-4b86-a851-b87aa0275795

https://github.com/user-attachments/assets/092cfb5a-3d6a-4cc1-a6ac-fda9d271e237


gregberge (approved):
Would be great to avoid using `tcls` for perf if it's useless and to put property on each lines. 283 lines for a simple component like that is 🤯.

We will have to refactor a bit the project at some point.

Anyway OK for that 👌

packages/gitbook/src/components/SpaceLayout/SpaceLayout.tsx
```diff
@@ -152,6 +153,12 @@ export async function SpaceLayout(props: {
                                         </React.Suspense>
                                     </div>
                                 )}
+                                {!withTopHeader && withSections && sections && (
+                                    <SiteSectionList
+                                        className={tcls('hidden', 'lg:block')}
+                                        {...sections}
```
valentin0h, line 159: I'd keep the same prop structure as the current component, in a proper prop drilling fashion,`sections={sections}`

and in SiteSectionList

```typescript
export function SiteSectionList(props: {
    sections: SectionsList;
    className: ClassValue;
}) {
```

valentin0h (approved):
Nice 👍 

A few things that we should address in other PRs.

- No header AND no ToC combo. Common for landing pages I'd think.
- No header mobile version. We still render a header.

zenoachtig:
> Would be great to avoid using `tcls` for perf if it's useless and to put property on each lines. 283 lines for a simple component like that is 🤯.
> 
> We will have to refactor a bit the project at some point.
> 
> Anyway OK for that 👌

I thought `tcls` was necessary for us in projects, but [reading up on it](https://github.com/dcastil/tailwind-merge/blob/v2.6.0/docs/when-and-how-to-use-it.md) I suppose it isn't. I'll refactor it to use less lines, but I still like grouping classes together functionally for better DX.

PR #2721, opened 2025-01-10
Force route handler to be dynamic to avoid errors

gregberge:
Fix GITBOOK-OPEN-1WYY
Fix GITBOOK-OPEN-1WYX


emmerich (approved):
Approving: as soon as we use next/headers we are dynamic anyways.

PR #2728, opened 2025-01-13
Fix server actions stability

gregberge:
Add `NEXT_SERVER_ACTIONS_ENCRYPTION_KEY` to ensure server-actions are
stable across deployments.

Without that, when we deploy server-actions can randomly failed because
we have cache on HTML/JS bundles.

Related: https://github.com/vercel/next.js/discussions/73951, https://nextjs.org/docs/app/building-your-application/data-fetching/server-actions-and-mutations#overwriting-encryption-keys-advanced

### DIFF
--- a/.changeset/perfect-laws-pay.md
+++ b/.changeset/perfect-laws-pay.md
@@ -1,0 +1,5 @@
+---
+'gitbook': patch
+---
+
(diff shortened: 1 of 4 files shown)


emmerich (approved):
Note the key needed to be a specific format: https://github.com/vercel/next.js/issues/61020

PR #2736, opened 2025-01-13
Enable anti-aliasing for published docs

scazan:
This enables anti-aliasing on text in docs for a sharper look.

After <- before:
Screenshot 2025-01-13 at 17 13 54

Screenshot 2025-01-13 at 17 13 09


clairechabas (approved):
Text looks better 👍 Let's double check that the icons are still displaying correctly though. Don't know how that would affect them but better safe than sorry :)

zenoachtig (approved):

PR #2739, opened 2025-01-14
Remove useless animation on section tabs

gregberge:
Page is reloaded (for technical reasons), so the animation is not accurate here.


valentin0h (approved):

viktorrenkema (approved):
We could consider bringing this back with an API that's made to work well with page navigation, but yeah rn this doesnt add value if it just makes it feel broken.

PR #2745, opened 2025-01-14
Allow only valid values for theme query string

gregberge:
Fix GITBOOK-OPEN-1X8D

### DIFF
--- a/packages/gitbook/src/middleware.ts
+++ b/packages/gitbook/src/middleware.ts
@@ -264,7 +264,7 @@ export async function middleware(request: NextRequest) {
     }
 
     const theme = url.searchParams.get('theme');
-    if (theme) {
+    if (theme === 'dark' || theme === 'light') {
         headers.set('x-gitbook-theme', theme);
     }
 
--- a/packages/gitbook/src/middleware.ts
+++ b/packages/gitbook/src/middleware.ts
@@ -5,6 +5,7 @@ import jwt from 'jsonwebtoken';
 import type { ResponseCookie } from 'next/dist/compiled/@edge-runtime/cookies';
 import { NextResponse, NextRequest } from 'next/server';
 import hash from 'object-hash';
+import rison from 'rison';
 
 import {
     PublishedContentWithCache,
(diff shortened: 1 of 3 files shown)


taranvohra (changes requested):
Use `api.CustomizationThemeMode`

packages/gitbook/src/middleware.ts
```diff
@@ -259,12 +261,12 @@ export async function middleware(request: NextRequest) {
     }
 
     const customization = url.searchParams.get('customization');
-    if (customization) {
+    if (customization && validateSerializedCustomization(customization)) {
```
taranvohra, line 264: What is the reason we need to validate this? We send it from our gitbook app and if it's broken then the problem is there not here. We are just suppressing it here

gregberge: https://gitbook-x.sentry.io/issues/6216219929/events/a6aa7a54b061407d8c1e7ce7aa57ed7e/

We can't pass any value in the header.

gregberge: So everything coming from query string should be validated.

gregberge: See https://dyor.hacken.ai/~gitbook/icon?size=small&theme=1ACUSTART%27%22*%2F%0D%0A+%3C%3FphpACUEND

taranvohra: Right but how is this `theme` being passed? Again, we generate these URLs and if the values are garbled then we need to fix the problem at source.

gregberge: If it's public it has to be validated even if we are generating the URL anything can modify the query string or put one.

taranvohra: Okay so users are passing garbage values themselves

taranvohra (approved):

packages/gitbook/src/middleware.ts
gregberge: Exactly, especially for themes. Internet and browsers are wild!

PR #2747, opened 2025-01-14
Antialias text but not code/pre elements

clairechabas:
Apply antialiasing on any text except inline/block code elements to avoid contrast issues.

### DIFF
--- a/packages/gitbook/src/components/RootLayout/globals.css
+++ b/packages/gitbook/src/components/RootLayout/globals.css
@@ -23,9 +23,7 @@
         @apply leading-relaxed;
     }
     body {
-        @apply text-dark dark:text-light;
-        -webkit-font-smoothing: antialiased;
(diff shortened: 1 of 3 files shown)


packages/gitbook/src/components/RootLayout/globals.css
```diff
@@ -24,6 +24,8 @@
     }
     body {
         @apply text-dark dark:text-light;
+        -webkit-font-smoothing: antialiased;
+        -moz-osx-font-smoothing: grayscale;
```
scazan, line 28: why not use @apply to lean into tailwind ?

clairechabas: Oups my bad just an old default habit when I'm in a css file ^^ Updated

scazan (approved):

PR #2751, opened 2025-01-15
Fix error when accessing some not found pages

gregberge:
## Context

Some URLs are not matched by the middleware but they are matched from our main route.

In this PR we handle that gracefully, if a request is not from the middleware we will display a proper 404 instead of a 500 error.

CleanShot 2025-01-15 at 10 22 07@2x

## Sentry

Fix GITBOOK-OPEN-1XDB
Fix GITBOOK-OPEN-1XDC
Fix GITBOOK-OPEN-1X07
Fix GITBOOK-OPEN-1XCJ
Fix GITBOOK-OPEN-1X86


packages/gitbook/src/app/(site)/layout.tsx
```diff
@@ -1,18 +1,30 @@
...
+    const fromMiddleware = await checkIsFromMiddleware();
+    if (!fromMiddleware) {
+        return (
+            <html lang="en">
+                <body className="font-[sans-serif]">
+                    <main>{props.children}</main>
+                </body>
+            </html>
+        );
+    }
```
SamyPesse, line 20: While it works, I think having to handle it in every single route/layout that we have is expensive.

Another solution could have been:

- When the middleware does the `NextResponse.rewrite`, we put a prefix to the path like `middleware`
- We wrap all the routes in a folder `src/app/middleware`
- We put a not found / layout outside for the case that are not handled

gregberge: Done, the default 404 for Vercel is good enough.

SamyPesse (approved):

PR #2755, opened 2025-01-17
Restore borders on embed blocks with Captions

viktorrenkema:
Fixes oversight in commit 994eb. 
The Caption block wraps a couple other blocks to enable captions. In a prior PR the border on the Caption block was removed as we wanted it removed on images and drawings, as these can function as diagrams with white backgrounds, and a border adds more noise than utility then.

However, this also removed the border on the Embed block, which is an external link block. 

This PR:
- restores the bordering for Embed blocks
- also improves consistency by having the Files block adhere to the same bordering styles from Caption, it was using its own but these were a bit different

Before (url embed block & files block)
CleanShot 2025-01-17 at 10 26 25@2x

After:

CleanShot 2025-01-17 at 10 25 41@2x

### DIFF
--- a/packages/gitbook/src/components/DocumentView/Caption.tsx
+++ b/packages/gitbook/src/components/DocumentView/Caption.tsx
@@ -23,16 +23,16 @@ export function Caption(
         fit?: boolean;
         wrapperStyle?: ClassValue;
         block: DocumentBlockImage | DocumentBlockDrawing | DocumentBlockEmbed | DocumentBlockFile;
+        withBorder?: boolean;
     } & DocumentContextProps,
 ) {
-    const needsBorder = props.block.type === 'embed' || props.block.type === 'file';
-
     const {
         children,
         document,
         block,
         context,
         fit = false,
+        withBorder = false,
         wrapperStyle = [
             'relative',
             'overflow-hidden',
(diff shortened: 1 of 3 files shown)


packages/gitbook/src/components/DocumentView/File.tsx
```diff
@@ -21,7 +21,7 @@ export async function File(props: BlockProps<DocumentBlockFile>) {
     const contentType = getSimplifiedContentType(file.contentType);
 
     return (
-        <Caption {...props} wrapperStyle={[]}>
+        <Caption {...props}>
             <Link
                 href={file.downloadURL}
                 download={file.name}
```
viktorrenkema, line 27: @SamyPesse I removed the styles and empty wrapperStyle that was passed to Caption in the Files block. As also mentioned in the PR description: "improves consistency by having the Files block adhere to the same bordering styles from Caption, it was using its own but these were a bit different"

Let me know if you added explicitly different bordering styles on the Files block on purpose, but if not, I think its nice to streamline these styles while i'm at it.

packages/gitbook/src/components/DocumentView/Caption.tsx
```diff
@@ -25,6 +25,8 @@ export function Caption(
         block: DocumentBlockImage | DocumentBlockDrawing | DocumentBlockEmbed | DocumentBlockFile;
     } & DocumentContextProps,
 ) {
+    const needsBorder = props.block.type === 'embed' || props.block.type === 'file';
```
emmerich, line 28: Why not pass it as a prop like `withBorder`?

viktorrenkema: That's better yeah, [changed](https://github.com/GitbookIO/gitbook/pull/2755/commits/2dad85ffbf13853165027f948f3c9f6766a549f8).

emmerich (approved):

PR #2757, opened 2025-01-20
Fix Image zoom e2e test

taranvohra:


gregberge (commented):
CleanShot 2025-01-20 at 10 57 02@2x
This can't be null, the error guard is useless

gregberge (approved):

PR #2765, opened 2025-01-21
Reduce gap between subsequent header buttons

zenoachtig:
Screenshot 2025-01-21 at 17 28 56
Screenshot 2025-01-21 at 17 29 04

### DIFF
--- a/packages/gitbook/src/components/Header/HeaderLinks.tsx
+++ b/packages/gitbook/src/components/Header/HeaderLinks.tsx
@@ -13,7 +13,7 @@ export async function HeaderLinks({ children }: HeaderLinksProps) {
         <div
             className={tcls(
                 styles.containerHeaderlinks,
-                'grow shrink flex justify-end items-center gap-x-4 lg:gap-x-6 min-w-9 z-20 [&>.button+.button]:-ml-3',
+                'grow shrink flex justify-end items-center gap-x-4 lg:gap-x-6 min-w-9 z-20 lg:[&>.button+.button]:-ml-2',
             )}
         >
             {children}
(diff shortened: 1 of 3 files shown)


packages/gitbook/src/components/primitives/Button.tsx
```diff
@@ -62,6 +62,7 @@ export function Button({
     const sizeClasses = sizes[size] || sizes.default;
 
     const domClassName = tcls(
+        'button',
```
valentin0h, line 65: Not sure what this is, but it applies to all buttons

valentin0h (approved):

packages/gitbook/src/components/primitives/Button.tsx
zenoachtig: It's needed to be able to select buttons from its parent (`lg:[&>.button+.button]:-ml-2`), this serves as an identifying class. Indeed not _officially_ needed on all buttons, but it seemed cleaner than to add it as an override class to the `<Button>` instance in `HeaderLink.tsx`… I think it's fine tbh

packages/gitbook/src/components/Header/HeaderLinks.tsx
```diff
@@ -13,7 +13,7 @@ export async function HeaderLinks({ children }: HeaderLinksProps) {
         <div
             className={tcls(
                 styles.containerHeaderlinks,
-                'grow shrink flex justify-end items-center gap-x-4 lg:gap-x-6 min-w-9 z-20',
+                'grow shrink flex justify-end items-center gap-x-4 lg:gap-x-6 min-w-9 z-20 lg:[&>.button+.button]:-ml-2',
```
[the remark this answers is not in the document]
zenoachtig: Yep, this [next sibling combinator](https://developer.mozilla.org/en-US/docs/Web/CSS/Next-sibling_combinator) selects any .button that is _preceded by_ a .button. That means it selects the 2nd, 3rd, nth button in a row, but never the first.

I'd like to keep users' freedom of ordering their links & buttons in any way they want intact, which this does 😊

PR #2766, opened 2025-01-21
Increase search bar size on sites with few header links

zenoachtig:


zenoachtig:
Closed in favour of #2936

PR #2767, opened 2025-01-22
Generate sitemap for all sections and spaces

gregberge:
Currently the sitemap does not include all site links. With this new approach we include all the links in the site by having a root site map and one site map per section and site space.

I also simplified the site space markup:

- Remove useless attributes
- Use a more precise lastmod (ISO format is supported)

**Example**

- https://pr2767.gitbook-open.pages.dev/docs.gitbook.com/sitemap.xml

### DIFF
--- a/packages/gitbook/src/app/middleware/(site)/(core)/sitemap-pages.xml/route.ts
+++ b/packages/gitbook/src/app/middleware/(site)/(core)/sitemap-pages.xml/route.ts
@@ -1,0 +1,100 @@
+import { RevisionPage, RevisionPageDocument, RevisionPageGroup } from '@gitbook/api';
+import jsontoxml from 'jsontoxml';
+
+import { getRevisionPages, getSpace } from '@/lib/api';
+import { getAbsoluteHref, getBasePath } from '@/lib/links';
+import { getPagePath } from '@/lib/pages';
+import { getSiteContentPointer } from '@/lib/pointer';
+import { isPageIndexable } from '@/lib/seo';
+
+export const runtime = 'edge';
+
+/**
+ * Generate a sitemap.xml for the current space.
+ */
+export async function GET() {
+    const pointer = await getSiteContentPointer();
+
+    const revisionId = await (async () => {
+        if (pointer.revisionId) {
+            return pointer.revisionId;
+        }
+        const space = await getSpace(pointer.spaceId, pointer.siteShareKey);
+        return space.revision;
+    })();
+
+    const rootPages = await getRevisionPages(pointer.spaceId, revisionId, { metadata: false });
+
+    const pages = flattenPages(rootPages, (page) => !page.hidden && isPageIndexable([], page));
+
+    const urls = await Promise.all(
+        pages.map(async ({ page, depth }) => {
+            // Decay priority with depth
+            const priority = Math.pow(2, -0.25 * depth);
+            // Normalize to keep 2 decimals
+            const normalizedPriority = Math.floor(100 * priority) / 100;
+            const lastModified = page.updatedAt || page.createdAt;
+
+            const url: { loc: string; priority: number; lastmod?: string } = {
+                priority: normalizedPriority,
+                loc: await getAbsoluteHref(getPagePath(rootPages, page), true),
+            };
+
+            if (lastModified) {
+                url.lastmod = new Date(lastModified).toISOString();
+            }
+
+            return { url };
+        }),
+    );
+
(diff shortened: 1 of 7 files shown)


packages/gitbook/src/app/middleware/(site)/(core)/~gitbook/sitemaps/[[...pathname]]/route.ts
```diff
@@ -0,0 +1,112 @@
...
+export async function GET(req: NextRequest, { params }: { params: Promise<Params> }) {
+    const { pathname } = await params;
+    const lastPath = pathname[pathname.length - 1];
+    if (lastPath !== 'sitemap.xml') {
+        return new Response('Not found', { status: 404 });
+    }
+    const paths = pathname.slice(0, -1);
+    const basePath = await getBasePath();
+    const origin = `https://${basePath}`;
+    const path = `/${paths.join('/')}`;
```
Comment (gregberge) [inline packages/gitbook/src/app/middleware/(site)/(core)/~gitbook/sitemaps/[[...pathname]]/route.ts:30]: @SamyPesse I have doubt on this logic to build the lookup URL. It works locally but not sure it's the correct approach.

Comment (gregberge) [reply packages/gitbook/src/app/middleware/(site)/(core)/~gitbook/sitemaps/[[...pathname]]/route.ts:30]: Another concern I have is to restrict the ability to generate random sitemaps (one for each page for example).

packages/gitbook/src/lib/lookup.ts
gregberge, line 1: I am just moving some code here, nothing changed.

packages/gitbook/src/app/middleware/(site)/(core)/sitemap-pages.xml/route.ts
```diff
@@ -0,0 +1,100 @@
...
+export async function GET() {
+    const pointer = await getSiteContentPointer();
+
+    const revisionId = await (async () => {
+        if (pointer.revisionId) {
+            return pointer.revisionId;
+        }
+        const space = await getSpace(pointer.spaceId, pointer.siteShareKey);
+        return space.revision;
+    })();
```
SamyPesse, line 24: I'm not sure it's worth optimizing for this; the only case that matters in reality will be the case where `pointer.revisionId` is not defined

```suggestion
    const revisionId = pointer.revisionId ?? (await getSpace(pointer.spaceId, pointer.siteShareKey).revision);
```

packages/gitbook/src/app/middleware/(site)/(core)/sitemap.xml/route.ts
```diff
@@ -15,43 +16,30 @@ export const runtime = 'edge';
...
+    const { structure: siteStructure } = await getPublishedContentSite({
+        organizationId: pointer.organizationId,
+        siteId: pointer.siteId,
+        siteShareKey: pointer.siteShareKey,
+    });
 
-            const lastModified = page.updatedAt || page.createdAt;
+    // This sitemap is only available at root (/sitemap.xml).
+    if (!checkIsRootPointer(pointer, siteStructure)) {
+        return new Response('Not found', { status: 404 });
```
SamyPesse, line 28: Should we put a better error message:

```suggestion
        return new Response('Root sitemap is only served from the root of the site', { status: 404 });
```

packages/gitbook/src/app/middleware/(site)/(core)/sitemap.xml/route.ts
```diff
@@ -68,30 +56,66 @@ export async function GET(req: NextRequest) {
...
+ * Get Sitemap URLs from site spaces.
+ */
+async function getUrlsFromSiteSpaces(siteSpaces: SiteSpace[]): Promise<string[]> {
+    const urls = await Promise.all(
+        siteSpaces.map(async (siteSpace) => {
+            if (!siteSpace.urls.published) {
+                return null;
+            }
+            const url = new URL(siteSpace.urls.published);
+            return getAbsoluteHref(joinPath(url.pathname, 'sitemap-pages.xml'), true);
```
SamyPesse, line 117: I'm not sure this logic will work in all cases:

- `getAbsoluteHref` will append based on the current request URL
- if a site is served under a slug in a org domain (ex: orgdomain.com/siteslug`)
- a part of the URL will be lost
```suggestion
            const url = new URL(siteSpace.urls.published);
            url.pathname = joinPath(url.pathname, 'sitemap-pages.xml')
            return url.toString();
```

SamyPesse (approved):

PR #2769, opened 2025-01-23
Align the revisions/changes path for multi-id with the normal one

SamyPesse:
We use `/~/changes/:id` for the normal mode, but we were using `/~changes/:id`.

### DIFF
--- a/packages/gitbook/src/middleware.ts
+++ b/packages/gitbook/src/middleware.ts
@@ -494,7 +494,7 @@ async function lookupSiteOrSpaceInMultiIdMode(
 
     const eatPathId = (...prefixes: string[]): string | undefined => {
         const match = prefixes.every((prefix, index) => pathSegments[index] === prefix);
-        if (!match || pathSegments.length < (prefixes.length + 1)) {
+        if (!match || pathSegments.length < prefixes.length + 1) {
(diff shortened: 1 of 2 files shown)


packages/gitbook/src/middleware.ts
```diff
@@ -482,8 +482,8 @@ async function lookupSiteInMultiMode(request: NextRequest, url: URL): Promise<Lo
  *
  * The format of the path is:
  *   - /~space|~site/:id/:path
- *   - /~space|~site/:id/~changes/:changeId/:path
- *   - /~space|~site/:id/~revisions/:revisionId/:path
+ *   - /~space|~site/:id/~/changes/:changeId/:path
+ *   - /~space|~site/:id/~/revisions/:revisionId/:path
```
taranvohra, line 486: Should we add backwards compatibility for old URLs to redirect to the proper one in case some users are already using the old format?

taranvohra (approved):

packages/gitbook/src/middleware.ts
SamyPesse: No, there are not used anywhere (multi-id mode). Only the new preview was starting to use them.

PR #2772, opened 2025-01-23
Add breadcrumbs to search results

zenoachtig:
When the search displays many similarly named pages, the user has no idea which search result leads where on the site or which result they actually need.

CleanShot 2025-01-16 at 13 24 44@2x

By adding breadcrumbs, we make it easier to understand what content we're matching. For now we are adding the section and variant names as breadcrumbs. Ideally we'd support adding the ancestor page titles in a future patch.

# Preview
Screenshot 2025-03-28 at 10 18 39
Screenshot 2025-03-28 at 10 18 29

Screenshot 2025-03-28 at 10 18 51
Screenshot 2025-03-28 at 10 18 54


packages/gitbook/src/components/Search/server-actions.tsx
```diff
@@ -317,12 +320,15 @@ async function transformSectionsAndPage(args: {
         })) ?? [],
     );
 
+    const pageData = await fetchPageData({ pathname: [item.path] });
```
zenoachtig, line 323: This function correctly gets the page data when the page is in the same space as we're currently searching in, but doesn't work when we need to search across sections/variants. Any suggestions to make it work are welcome.

SamyPesse: We should not use this function as it's tied to the current context, and generally, it'll make things slower by fetching a lot of data to render search results.

We need to think about a different approach, potentially by changing our API, because otherwise, this PR could badly impact performances.

valentin0h (approved):

packages/gitbook/src/components/Search/server-actions.tsx
```diff
@@ -258,15 +262,21 @@ async function searchSiteContent(
...
                 return Promise.all(
                     spaceItem.pages.map((pageItem) =>
                         transformSitePageResult(context, {
                             pageItem,
                             spaceItem,
-                            space: siteSpace?.space,
+                            space:
+                                !siteSection || siteSection.siteSpaces.length > 1
+                                    ? siteSpace?.space
+                                    : undefined,
```
valentin0h, line 277: We shouldn't mix breadcrumb logic here

packages/gitbook/src/components/Search/server-actions.tsx
```diff
@@ -48,6 +49,7 @@ export interface ComputedPageResult {
 
     /** When part of a multi-spaces search, the title of the space */
     spaceTitle?: string;
+    section?: SiteSection;
```
SamyPesse, line 52: We should not pass all of these data to the client side.

Instead we should merge both `spaceTitle` and `section` as a `breadcrumbs` property here

Something like:

```
breadcrumbs?: Array<{ icon?: IconName; label: string; >
```

packages/gitbook/src/components/Search/server-actions.tsx
```diff
@@ -258,7 +260,9 @@ async function searchSiteContent(
     return (
         await Promise.all(
             searchResults.map(async (spaceItem) => {
+                const sections = getSiteStructureSections(structure, { ignoreGroups: true });
                 const siteSpace = findSiteSpaceById(structure, spaceItem.id);
+                const siteSection = sections.find((section) => section.id === siteSpace?.section);
```
SamyPesse, line 265: Do we want to take into consideration the section group as well in the breadcrumb?

packages/gitbook/src/components/Search/server-actions.tsx
```diff
@@ -267,6 +271,7 @@ async function searchSiteContent(
                             spaceItem,
                             space: siteSpace?.space,
                             spaceURL: siteSpace?.urls.published,
+                            siteSection: siteSection ?? undefined,
```
SamyPesse, line 274: With my sugegstion, you could pass a `Breadcrumb[]` directly here

PR #2780, opened 2025-01-29
Use Scalar OpenAPI parser

gregberge:
Lot of problems come from the parsing of the OpenAPI spec and the resolution of refs. By replacing our custom parser with the Scalar one we avoid these issues.


packages/react-openapi/src/fetchOpenAPIOperation.ts
```diff
@@ -79,56 +80,51 @@ export async function fetchOpenAPIOperation(
...
-    const commonParameters = await resolveOpenAPIPath<OpenAPIV3.ParameterObject[]>(
-        input.url,
-        ['paths', input.path, 'parameters'],
-        fetcher,
-    );
-    if (commonParameters) {
-        operation = {
-            ...operation,
-            parameters: [...commonParameters, ...(operation.parameters ?? [])],
-        };
```
gregberge, line 104: I may have to put back this logic.

packages/react-openapi/src/fetchOpenAPIOperation.ts
```diff
@@ -79,56 +80,60 @@ export async function fetchOpenAPIOperation(
...
     }
 
-    // Resolve servers
-    const servers = await resolveOpenAPIPath<OpenAPIV3.ServerObject[]>(
-        input.url,
-        ['servers'],
-        fetcher,
-    );
+    const servers: OpenAPIV3.ServerObject[] = 'servers' in schema ? (schema.servers ?? []) : [];
+    const security: OpenAPIV3.SecurityRequirementObject[] = operation.security ?? [];
```
nolannbiron, line 121: Can we handle top-level security if it is not provided in the operation ?

I was doing this : 
` // If the operation doesn't have security, try to resolve it from the spec
    if (!operation.security) {
        operation.security = await resolveOpenAPIPath<OpenAPIV3.SecurityRequirementObject[]>(
            input.url,
            ['security'],
            fetcher,
        );
    }`

packages/react-openapi/src/stringifyOpenAPI.ts
```diff
@@ -1,20 +1,6 @@
-import { SYMBOL_MARKDOWN_PARSED, SYMBOL_REF_RESOLVED } from './resolveOpenAPIPath';
-
 /**
  * Stringify an OpenAPI object. Same API as JSON.stringify.
  */
 export function stringifyOpenAPI(body: unknown, transformer?: null, indent?: number): string {
-    return JSON.stringify(
```
nolannbiron, line 7: Does Scalar handle the `__$markdownParsed`?

We had an issue about this :
[RND-6046 Request & Response sample wrapping description field in p tag and showing a __$markdownParsed property when it's not present in the spec file](https://linear.app/gitbook-x/issue/RND-6046/request-and-response-sample-wrapping-description-field-in-p-tag-and)

packages/react-openapi/src/fetchOpenAPIOperation.ts
```diff
@@ -79,56 +80,60 @@ export async function fetchOpenAPIOperation(
...
-        ['paths', input.path, input.method],
-        fetcher,
-    );
+    const specData = await fetcher.fetch(input.url);
+
+    const { valid } = await validate(specData);
+
+    // Spec is invalid, we stop here.
+    if (!valid) {
+        return null;
```
emmerich, line 89: Log an error maybe? It's out of our control, we will render nothing + hard to track why

packages/react-openapi/src/fetchOpenAPIOperation.ts
```diff
@@ -79,56 +80,60 @@ export async function fetchOpenAPIOperation(
...
+    }
+
+    const { schema } = await dereference(specData);
+
+    // No schema, we stop here.
+    if (!schema) {
+        return null;
```
emmerich, line 96: Same, log an error?

packages/react-openapi/src/fetchOpenAPIOperation.ts
```diff
@@ -142,6 +147,75 @@ export async function fetchOpenAPIOperation(
...
+            promises[obj.description] = parseMarkdown(obj.description);
+        }
+        return obj;
+    });
+    await Promise.all(
+        Object.entries(promises).map(async ([key, promise]) => {
+            results[key] = await promise;
+        }),
+    );
+    return traverse(spec, (obj) => {
```
emmerich, line 172: you could do it with a single traverse

packages/react-openapi/src/stringifyOpenAPI.ts
gregberge: I will test this one.

emmerich: `__$markdownParsed` was something on our side, we used it to tag a description field as having already been parsed to prevent reparsing the same thing over and over. I already tested with this branch and it's working.

packages/react-openapi/src/fetchOpenAPIOperation.ts
gregberge: I should throw in this case to be honest. Because we should not allow it.

packages/react-openapi/src/fetchOpenAPIOperation.ts
gregberge: How can I do it, the traverse is not async?

emmerich: Just a coding problem, but it's fine I think. We'll see if the double traverse causing issues with larger OpenAPI specs

packages/react-openapi/src/OpenAPICodeSample.tsx
```diff
@@ -50,8 +51,11 @@ export function OpenAPICodeSample(props: {
         }
     });
 
-    const requestBody = noReference(data.operation.requestBody);
-    const requestBodyContent = requestBody ? Object.entries(requestBody.content)[0] : undefined;
+    const requestBody = noReference((data.operation as OpenAPIV3.OperationObject).requestBody);
```
emmerich, line 54: The `as` is a bit risky no? What else could this be?

packages/react-openapi/src/fetchOpenAPIOperation.ts
```diff
@@ -79,56 +36,59 @@ export async function fetchOpenAPIOperation(
...
+        throw new Error(`Invalid OpenAPI spec: ${input.url}`);
+    }
+
+    let { schema } = await dereference(specData);
+
+    // No schema, we stop here.
+    if (!schema) {
+        throw new Error(`Schema undefined following the dereference operation: ${input.url}`);
+    }
+    schema = schema as OpenAPIV3_1.Document;
```
emmerich, line 54: Another `as` - why is it needed?

gregberge: I will change it

gregberge: I isolated it in a function, so it's clearer.

packages/react-openapi/src/fetchOpenAPIOperation.ts
gregberge: Note that I traverse only the operation.

emmerich (approved):
It's a big change. I'm approving it, it's a step in the right direction. But a different parser means potentially a change in behaviour for some docs sites - the main thing I'm thinking is that our parser and our blocks have been written together. Now that the parser is doing more, the blocks might not support it - best case not rendering the new properties, worst case causing errors.

I'm going to get the list of some of the top sites using API blocks and we can test them properly on the feature branch before we merge.

emmerich (changes requested):
Some errors to investigate, will chat internally

emmerich (approved):
We'll merge tomorrow morning 👍

PR #2786, opened 2025-01-30
Highlight code client-side

gregberge:
This PR let us highlight the code client-side if the block is offscreen. It represents a huge performance improvement because highlighting code is very expensive. Also it unlocks the ability to create interactive blocks with code by using them client-side.

### DIFF
--- a/packages/gitbook/src/components/DocumentView/Caption.tsx
+++ b/packages/gitbook/src/components/DocumentView/Caption.tsx
@@ -6,7 +6,7 @@ import {
     JSONDocument,
 } from '@gitbook/api';
 
-import { getNodeFragmentByName, isNodeEmpty } from '@/lib/document';
+import { getNodeFragmentByName, isNodeEmpty } from '@/lib/document-util';
 import { ClassValue, tcls } from '@/lib/tailwind';
 
 import { DocumentContextProps } from './DocumentView';
--- a/packages/gitbook/src/components/DocumentView/Caption.tsx
+++ b/packages/gitbook/src/components/DocumentView/Caption.tsx
@@ -6,7 +6,7 @@ import {
     JSONDocument,
 } from '@gitbook/api';
 
-import { getNodeFragmentByName, isNodeEmpty } from '@/lib/document-util';
+import { getNodeFragmentByName, isNodeEmpty } from '@/lib/document';
 import { ClassValue, tcls } from '@/lib/tailwind';
 
 import { DocumentContextProps } from './DocumentView';
--- a/packages/gitbook/src/components/DocumentView/CodeBlock/ClientCodeBlock.tsx
+++ b/packages/gitbook/src/components/DocumentView/CodeBlock/ClientCodeBlock.tsx
@@ -15,7 +15,8 @@ type ClientBlockProps = Pick<BlockProps<DocumentBlockCode>, 'block' | 'document'
 };
 
 /**
- * Render a code-block client-side.
+ * Render a code-block client-side by calling a server actions to highlight the code.
+ * It allows us to defer some load to avoid blocking the rendering of the whole page with block highlighting.
  */
 export function ClientCodeBlock(props: ClientBlockProps) {
     const { block, document, style, inlines } = props;
--- a/packages/gitbook/src/components/DocumentView/CodeBlock/ClientCodeBlock.tsx
+++ b/packages/gitbook/src/components/DocumentView/CodeBlock/ClientCodeBlock.tsx
@@ -7,6 +7,7 @@ import type { HighlightLine, RichInlineIndexed } from './highlight';
 import type { BlockProps } from '../Block';
 import './theme.css';
 import { ClientCodeBlockRenderer } from './CodeBlockRenderer';
+import { highlightAction } from './highlight-action';
 import { plainHighlight } from './plain-highlight';
 
 type ClientBlockProps = Pick<BlockProps<DocumentBlockCode>, 'block' | 'document' | 'style'> & {
@@ -20,7 +21,7 @@ export function ClientCodeBlock(props: ClientBlockProps) {
     const { block, document, style, inlines } = props;
     const [lines, setLines] = useState<HighlightLine[]>(() => plainHighlight(block));
     useEffect(() => {
-        import('./highlight').then(({ highlight }) => highlight(block, inlines).then(setLines));
+        highlightAction(block, inlines).then(setLines);
     }, [block, inlines]);
     return (
         <ClientCodeBlockRenderer block={block} document={document} style={style} lines={lines} />
--- a/packages/gitbook/src/components/DocumentView/CodeBlock/ClientCodeBlock.tsx
+++ b/packages/gitbook/src/components/DocumentView/CodeBlock/ClientCodeBlock.tsx
@@ -10,7 +10,7 @@ import { ClientCodeBlockRenderer } from './CodeBlockRenderer';
 import { highlightAction } from './highlight-action';
 import { plainHighlight } from './plain-highlight';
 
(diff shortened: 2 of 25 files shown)


packages/gitbook/src/components/DocumentView/CodeBlock/highlight.ts
```diff
@@ -330,7 +352,7 @@ const createHighlighter = createdBundledHighlighter<any, any>({
  * Load the highlighter, only once, and reuse it.
  * It makes sure to handle concurrent calls.
  */
-const loadHighlighter = singleton(async () => {
+const loadHighlighter = memoize(async () => {
```
gregberge, line 355: This part was a singleton and now it's memoized because I can't use server-side code in a client-component, even if we don't really use it.

packages/gitbook/src/components/DocumentView/CodeBlock/highlight.ts
```diff
@@ -340,19 +362,16 @@ const loadHighlighter = singleton(async () => {
     });
 });
 
-const loadLanguagesMutex = asyncMutexFunction();
-async function loadHighlighterLanguage(
+const loadHighlighterLanguage = memoize(async function loadHighlighterLanguage(
```
gregberge, line 365: Same here.

packages/gitbook/src/components/DocumentView/CodeBlock/highlight.ts
gregberge: I solved it by using a server action.

packages/gitbook/src/components/DocumentView/CodeBlock/highlight.ts
gregberge: I solved it by using a server action.

packages/gitbook/src/components/DocumentView/Caption.tsx
```diff
@@ -6,7 +6,7 @@ import {
     JSONDocument,
 } from '@gitbook/api';
 
-import { getNodeFragmentByName, isNodeEmpty } from '@/lib/document';
+import { getNodeFragmentByName, isNodeEmpty } from '@/lib/document-util';
```
SamyPesse, line 9: Why rename this file? it's already in `lib` so `util` is kind of obvious

packages/gitbook/src/components/DocumentView/CodeBlock/ClientCodeBlock.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { ClientCodeBlockRenderer } from './CodeBlockRenderer';
+import { highlightAction } from './highlight-action';
+import { plainHighlight } from './plain-highlight';
+
+type ClientBlockProps = Pick<BlockProps<DocumentBlockCode>, 'block' | 'document' | 'style'> & {
+    inlines: RichInlineIndexed[];
+};
+
+/**
+ * Render a code-block client-side.
```
SamyPesse, line 18: It'd be good to document a bit more the flow of this component

packages/gitbook/src/components/DocumentView/CodeBlock/CodeBlock.tsx
```diff
@@ -1,291 +1,52 @@
...
+            );
+        })();
 
-function CodeHighlightToken(props: {
-    token: HighlightToken;
-    document: JSONDocument;
-    context: DocumentContext;
-}) {
-    const { token, document, context } = props;
+        return { ...inline, body };
```
SamyPesse, line 35: I usually find it to be quite bad to merge interfaces this way. I'd recommend:

```suggestion
        return { inline, body };
```

It doesn't make the code more complicated and avoid a lot of spread with weird types everywhere.

packages/gitbook/src/components/DocumentView/CodeBlock/CodeBlockRenderer.tsx
```diff
@@ -0,0 +1,272 @@
...
+import { BlockProps } from '../Block';
+
+import './theme.css';
+
+type CodeBlockRendererProps = Pick<
+    BlockProps<DocumentBlockCode>,
+    'block' | 'document' | 'style'
+> & {
+    lines: HighlightLine[];
+};
```
SamyPesse, line 17: It's almost either to just write:

```suggestion
type CodeBlockRendererProps = {
    block: DocumentBlockCode;
    document: DocumentJSON;
    style: StyleValue;
    lines: HighlightLine[];
};
```

SamyPesse: 💥  by doing this, you can notice that you are passing `document` which means: every client code blocks gets the entire document JSON serialized by Next to be passed as a prop.

We can't do that, it'll multiply by x10/x100 the HTML size.

packages/gitbook/src/components/DocumentView/CodeBlock/CodeBlockRenderer.tsx
```diff
@@ -0,0 +1,272 @@
...
+        <>
+            {tokens.map((token, index) => (
+                <CodeHighlightToken key={index} token={token} document={document} />
+            ))}
+        </>
+    );
+}
+
+function CodeHighlightToken(props: { token: HighlightToken; document: JSONDocument }) {
+    const { token, document } = props;
```
SamyPesse, line 253: It looks like `document` is not even used

packages/gitbook/src/components/DocumentView/CodeBlock/highlight-action.ts
```diff
@@ -0,0 +1,9 @@
+'use server';
+
+import { DocumentBlockCode } from '@gitbook/api';
+
+import { highlight, RichInlineIndexed } from './highlight';
+
+export async function highlightAction(block: DocumentBlockCode, inlines: RichInlineIndexed[]) {
```
SamyPesse, line 7: Please comment to epxlain the need for it

packages/gitbook/src/components/DocumentView/CodeBlock/highlight.test.ts
```diff
@@ -1,9 +1,18 @@
+import type { DocumentBlockCode } from '@gitbook/api';
 import { it, expect } from 'bun:test';
 
-import { highlight } from './highlight';
+import { getInlines, highlight } from './highlight';
+
+async function highlightWithInlines(block: DocumentBlockCode) {
+    const inlines = getInlines(block).map((inline, index) => ({
+        ...inline,
+        body: null,
```
SamyPesse, line 9:
```suggestion
        inline,
        body: null,
```

packages/gitbook/src/lib/document-sections.ts
SamyPesse, line 1: Why is it not in `document-util` ?

SamyPesse (commented):
Looks good, but we should be careful about auditing the size of the props we pass to client components

packages/gitbook/src/components/DocumentView/Caption.tsx
gregberge: It was just during the refactoring, can rename it.

packages/gitbook/src/lib/document-sections.ts
gregberge: Because it's "server-only".

packages/gitbook/src/components/DocumentView/CodeBlock/CodeBlockRenderer.tsx
gregberge: When you use that notation, JSDOC comments are not preserved, it's always better to use `Pick`.

PR #2791, opened 2025-01-30
Setup package and deployment for v2

SamyPesse:
This branch contains the v2 of GitBook Open.


SamyPesse:
Few issues:

- Deployment URL for Cloudflare Worker is not working in CI because of https://github.com/cloudflare/wrangler-action/issues/343
- Cloudflare Worker version is not working because of https://github.com/opennextjs/opennextjs-cloudflare/issues/57

These issues are not blocking to iterate

gregberge (approved):
It does not work yet but at least we have a structure to iterate on.

PR #2794, opened 2025-01-30
Update Node version required to fix failing icons generate script

clairechabas:
We use [import.meta.resolve()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import.meta/resolve) for the icons' `generate` script. It's supported on Node from version 20.8 so we need to update the Node required version.


gregberge (approved):

SamyPesse:
Why are you using Node? it should use `bun`.

How are you running things?

clairechabas:
@SamyPesse the `icons` package `generate` script is using node for some reason instead of bun. Probably better to stick to bun like other scripts. Bun supports the feature in question so it will work too. I'm closing this PR in favor of [this one](https://github.com/GitbookIO/gitbook/pull/2798) then.

PR #2797, opened 2025-01-31
Highlight code client-side

gregberge:
Previously we were highlighting the code server-side. It's slow because the worker has to reload shiki every time. We now use it client-side, so once loaded highlight is nearly instant for all blocks.

- **Reapply "Highlight code client-side" (#2796)**
- **Highlight code client-side (instead of using an action)**
- **Always highlight client-side**
- **Simplify code**


packages/gitbook/src/components/DocumentView/CodeBlock/CodeBlockRenderer.tsx
```diff
@@ -0,0 +1,117 @@
...
+import './CodeBlockRenderer.css';
+
+type CodeBlockRendererProps = Pick<BlockProps<DocumentBlockCode>, 'block' | 'style'> & {
+    lines: HighlightLine[];
+};
+
+/**
+ * The logic of rendering a code block from lines.
+ */
+export function ClientCodeBlockRenderer(props: CodeBlockRendererProps) {
```
SamyPesse, line 22: Is this really a "client component" or the common one between server and client?

```suggestion
export function UniversalCodeBlockRenderer(props: CodeBlockRendererProps) {
```

packages/gitbook/src/lib/document-sections.ts
SamyPesse, line 1: Why did you needed to move this to a different file? where is it used on the client side?

Asking because it takes a document as input so it could indicate that we are leaking the document on the client side.

SamyPesse: I'm starting to think we should apply https://react.dev/reference/react/experimental_taintUniqueValue on the `JSOnDocument` and `Revision` to avoid any risk

packages/gitbook/src/components/DocumentView/CodeBlock/ClientCodeBlock.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Render a code-block client-side by calling a server actions to highlight the code.
+ * It allows us to defer some load to avoid blocking the rendering of the whole page with block highlighting.
+ */
+export function ClientCodeBlock(props: ClientBlockProps) {
+    const { block, style, inlines } = props;
+    const [lines, setLines] = useState<HighlightLine[]>(() => plainHighlight(block, []));
+    useEffect(() => {
+        import('./highlight').then(({ highlight }) => highlight(block, inlines).then(setLines));
```
SamyPesse, line 23: Right now we do this at startup time for all code blocks, it can potentially slow down the loading especially on mobile (as it takes CPU).

What do you think of moving this to use an IntersectionObserver and only do it when it gets close to the screen viewport?

I was thinking about this for the server action approach as well.

SamyPesse (approved):

packages/gitbook/src/components/DocumentView/CodeBlock/CodeBlockRenderer.tsx
gregberge: Yeah of course it was a typo.

packages/gitbook/src/lib/document-sections.ts
gregberge: It's necessary because `getNodeText` is called in `plain-highlighting` that is called client-side.

The `getDocumentSections` calls the API that is `server-only` (at some point).

packages/gitbook/src/components/DocumentView/CodeBlock/ClientCodeBlock.tsx
gregberge: > What do you think of moving this to use an IntersectionObserver and only do it when it gets close to the screen viewport?

Yes I was thinking about this. I will do it now.

packages/gitbook/src/components/hooks/useHasBeenInViewport.ts
```diff
@@ -0,0 +1,38 @@
...
+const HAS_INTERSECTION_OBSERVER = typeof IntersectionObserver !== 'undefined';
+
+/**
+ * Return true if the element has been in the viewport once.
+ */
+export function useHasBeenInViewport(
+    containerRef: React.RefObject<HTMLElement>,
+    options?: Pick<IntersectionObserverInit, 'root' | 'rootMargin' | 'threshold'>,
+) {
+    const [visible, setVisible] = useState(!HAS_INTERSECTION_OBSERVER);
```
SamyPesse, line 13: With SSR, it's going to be true on the server, and then false on the client 🤔 

I think you need `false` here and then set it to true in a hook.

gregberge: SSR ❤️

PR #2805, opened 2025-02-04
Page outline positioning

zenoachtig:
Fixes RND-6068, RND-5185, and an alignment issue with Ads.

# Outline stickiness and mode toggle
- Outline now stays in view as much as possible
- Mode toggle is displayed in outline sidebar, eliminating the need for a footer if it only contains a mode toggle
- Mode toggle is now available on pages without an outline view, like landing pages
Screenshot 2025-02-04 at 12 08 55
Screenshot 2025-02-04 at 12 09 26

# Ad alignment
- Ad blocks now always display at the bottom of the view, together with mode toggle (if enabled)
Screenshot 2025-02-04 at 12 11 04
Screenshot 2025-02-04 at 12 11 18

- Mode toggle (and ad) are always in view, even when the page outline starts underneath the page cover
Screenshot 2025-02-04 at 12 16 14
Screenshot 2025-02-04 at 12 17 34

# Footer tweaks
- Footer has been tweaked to rely less on the left and right columns. In the future we might make the sidebars sticky to prevent scrolling it up altogether.
Screenshot 2025-02-04 at 12 19 54
Screenshot 2025-02-04 at 12 19 49

PR #2807, opened 2025-02-04
Only push scrolling down when Footer is present

scazan:
This fixes vertical scrolling when there is no Footer below the content to see by conditionally adding the extra height. The screen was scrolling unnecessarily when the footer wasn't present. This makes sure that there is no scrolling when the footer is not present and the content does not go beyond the screen size.

Before:

https://github.com/user-attachments/assets/b00be21a-194b-4330-ad32-53eaeb0b6981

After:

https://github.com/user-attachments/assets/b7547106-1655-4087-b8a5-6615d8884ef1

### DIFF
--- a/.changeset/cold-cameras-work.md
+++ b/.changeset/cold-cameras-work.md
@@ -1,0 +1,5 @@
+---
+'gitbook': patch
+---
+
+Don't add extra page scroll when footer is not present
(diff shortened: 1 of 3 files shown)


clairechabas (approved):

zenoachtig:
Might want to check if this works well with https://github.com/GitbookIO/gitbook/pull/2805 still, both of which will hopefully merge today

zenoachtig:
Not sure if it's intentional or not but something changed with the 404 pages on small screens, they were displayed centered before and now they're aligned top: https://app.argos-ci.com/GitbookIO/gitbook/builds/2255/135184281

PR #2811, opened 2025-02-05
Handle unresolved references in OpenAPI response'

scazan:
This handles unresolved references passed to the body in an OpenAPI response by returning an empty object instead of a crash.

Renders any for the type instead:

Screenshot 2025-02-06 at 01 34 52

### DIFF
--- a/packages/react-openapi/src/OpenAPIResponse.tsx
+++ b/packages/react-openapi/src/OpenAPIResponse.tsx
@@ -7,18 +7,6 @@ import { OpenAPIClientContext } from './types';
 import { InteractiveSection } from './InteractiveSection';
 import { Markdown } from './Markdown';
 
-const handleReferences = (
-    input: OpenAPIV3.SchemaObject | OpenAPIV3.ReferenceObject | undefined,
-): OpenAPIV3.SchemaObject => {
-    const isReference = checkIsReference(input);
-
-    if (isReference || input === undefined) {
-        return {};
-    }
-
-    return input;
-};
-
 /**
  * Display an interactive response body.
  */
@@ -81,4 +69,17 @@ export function OpenAPIResponse(props: {
             ) : null}
         </>
     );
+}
+
+function handleReferences(
+    input: OpenAPIV3.SchemaObject | OpenAPIV3.ReferenceObject | undefined,
+): OpenAPIV3.SchemaObject {
+    const isReference = checkIsReference(input);
+
+    if (isReference || input === undefined) {
+        // If we find a reference that wasn't resolved or needed to be resolved externally, do not try to render it
+        return {};
+    }
+
+    return input;
 }
--- a/packages/react-openapi/src/OpenAPIResponse.tsx
+++ b/packages/react-openapi/src/OpenAPIResponse.tsx
@@ -59,7 +59,7 @@ export function OpenAPIResponse(props: {
                             label: contentType,
                             body: (
                                 <OpenAPIRootSchema
-                                    schema={handleUnresolvedReference(mediaType.schema) ?? {}}
+                                    schema={handleUnresolvedReference(mediaType.schema)}
                                     context={context}
                                 />
                             ),
--- a/packages/react-openapi/src/OpenAPIResponse.tsx
+++ b/packages/react-openapi/src/OpenAPIResponse.tsx
@@ -77,7 +77,8 @@ function handleUnresolvedReference(
     const isReference = checkIsReference(input);
 
     if (isReference || input === undefined) {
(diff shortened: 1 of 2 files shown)


packages/react-openapi/src/OpenAPIResponse.tsx
```diff
@@ -59,7 +59,7 @@ export function OpenAPIResponse(props: {
                             label: contentType,
                             body: (
                                 <OpenAPIRootSchema
-                                    schema={noReference(mediaType.schema) ?? {}}
+                                    schema={handleUnresolvedReference(mediaType.schema) ?? {}}
```
emmerich, line 62:
```suggestion
                                    schema={handleUnresolvedReference(mediaType.schema)}
```

packages/react-openapi/src/OpenAPIResponse.tsx
```diff
@@ -70,3 +70,16 @@ export function OpenAPIResponse(props: {
...
 }
+
+function handleUnresolvedReference(
+    input: OpenAPIV3.SchemaObject | OpenAPIV3.ReferenceObject | undefined,
+): OpenAPIV3.SchemaObject {
+    const isReference = checkIsReference(input);
+
+    if (isReference || input === undefined) {
+        // If we find a reference that wasn't resolved or needed to be resolved externally, do not try to render it
+        return {};
```
emmerich, line 81: I don't know if rendering `any` is the best solution. Is it possible to render the ref itself like `#/Schemas/User` or whatever?

packages/react-openapi/src/OpenAPIResponseExample.tsx
```diff
@@ -56,7 +56,7 @@ export function OpenAPIResponseExample(props: {
                 return null;
             }
 
-            const example: OpenAPIV3.ExampleObject | null = noReference(
+            const example: OpenAPIV3.ExampleObject | null = handleUnresolvedReference(
```
emmerich, line 59:
```suggestion
            const example = handleUnresolvedReference(
```

packages/react-openapi/src/OpenAPIResponseExample.tsx
```diff
@@ -115,3 +115,16 @@ export function OpenAPIResponseExample(props: {
...
 }
+
+function handleUnresolvedReference(
+    input: OpenAPIV3.ExampleObject | null,
+): OpenAPIV3.ExampleObject | null {
+    const isReference = checkIsReference(input?.value);
+
+    if (isReference) {
+        // If we find a reference that wasn't resolved or needed to be resolved externally, don't try to render it
+        return null;
```
emmerich, line 126: Again, is there something better that we can render like the ref itself?

scazan: We could render out the path here pretty easily. would kinda look something like this: 
Screenshot 2025-02-05 at 16 35 50

packages/react-openapi/src/OpenAPIResponse.tsx
scazan: yeah, we could definitely render the ref in some form. I'm modeling this off of what the editor renders at the moment (`any`). We could display something here that makes reference to the unresolved ref. maybe a link?

emmerich: I'd just render the ref as a string. It's at least somewhat helpful to the user and it's extremely clear to the author what went wrong.

packages/react-openapi/src/OpenAPIResponseExample.tsx
emmerich: Can we just render it as a URL?

packages/react-openapi/src/OpenAPIResponseExample.tsx
```diff
@@ -115,3 +115,16 @@ export function OpenAPIResponseExample(props: {
...
 }
+
+function handleUnresolvedReference(
+    input: OpenAPIV3.ExampleObject | null,
+): OpenAPIV3.ExampleObject | null {
+    const isReference = checkIsReference(input?.value);
+
+    if (isReference) {
+        // If we find a reference that wasn't resolved or needed to be resolved externally, render out the URL
+        return { value: `See: ${input.value.$ref}` };
```
emmerich, line 126:
```suggestion
        return { value: `${input.value.$ref}` };
```

emmerich (approved):

packages/react-openapi/src/OpenAPIResponseExample.tsx
scazan: I'll remove it now (will pull out the string template as well then)

PR #2812, opened 2025-02-06
Fix header offset

zenoachtig:
- Consistent use of the (renamed) `site-header:` and (new) `site-header-sections:` Tailwind variants in TOC, PageAside, and OpenAPI block
- Fixes API response block overlapping header if it has sections

Before:
Screenshot 2025-02-06 at 11 15 56

After:
Screenshot 2025-02-06 at 11 16 54

### DIFF
--- a/.changeset/khaki-bags-shout.md
+++ b/.changeset/khaki-bags-shout.md
@@ -1,0 +1,5 @@
+---
+'gitbook': patch
+---
+
+Improve header offset
--- a/packages/gitbook/src/components/DocumentView/OpenAPI/style.css
+++ b/packages/gitbook/src/components/DocumentView/OpenAPI/style.css
@@ -128,7 +128,7 @@
 }
 
 .openapi-column-preview-body {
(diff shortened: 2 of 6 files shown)


nolannbiron:
@zenoachtig Nice one! With small headers the block is pretty far [(example)](https://51111947.gitbook-open.pages.dev/docs.snyk.io/snyk-api/reference/containerimage), also the "On this page" TOC is going over the block, can we try to handle all those cases ?

zenoachtig:
> @zenoachtig Nice one! With small headers the block is pretty far [(example)](https://51111947.gitbook-open.pages.dev/docs.snyk.io/snyk-api/reference/containerimage), also the "On this page" TOC is going over the block, can we try to handle all those cases ?

Yeah whoops, my selector was incorrect (because the page always contains a `header > nav`, inside the `PageHeader`. Fix is up and will be ready in a sec 😊

nolannbiron (approved):

PR #2813, opened 2025-02-06
Support for OpenAPI references

nolannbiron:
Support for external references (link/path)

Before:
CleanShot 2025-02-06 at 15 02 33@2x

After:
CleanShot 2025-02-06 at 15 03 07@2x


packages/react-openapi/src/fetchOpenAPIOperation.ts
```diff
@@ -45,7 +47,16 @@ export async function fetchOpenAPIOperation(
...
 ): Promise<OpenAPIOperationData | null> {
     const refSchema = await fetcher.fetch(input.url);
-    const schema = await memoDereferenceSchema(refSchema, input.url);
+
+    // Get the prefix of the URL to resolve relative references
+    const prefix = input.url.split('/').slice(0, -1).join('/');
+
+    // Load the schema and fetch any external references
+    const { filesystem } = await load(refSchema, {
+        plugins: [fetchUrls(), readFiles(prefix)],
```
gregberge, line 56: Why do we need `readFiles`?

nolannbiron: Now that you mention it, I should handle the relative paths via fetchUrls and the prefix

packages/react-openapi/src/parser/fetchUrls.ts
```diff
@@ -0,0 +1,85 @@
...
+    // Exclude external URLs
+    const externalUrlPattern = /^(https?:\/\/|www\.|data:)/;
+
+    return !externalUrlPattern.test(path);
+}
+
+function getReferenceUrl(value: string, prefix?: string) {
+    if (isLocalFile(value)) {
+        return `${prefix}/${value}`;
+    }
```
gregberge, line 82: We can't support local file, so I don't think it's relevant to have that.

nolannbiron: Wrong naming it's not for local files, it's to handle all type of paths: 
`./xxx` 
`../xxx` 
`/xxx` 
`xxx/`

nolannbiron: I change the name

gregberge (approved):

PR #2815, opened 2025-02-06
Handle isArray schema type

nolannbiron:
Handle cases where the schema type is an array

Before :
CleanShot 2025-02-06 at 23 11 20@2x

After:
CleanShot 2025-02-06 at 23 20 01@2x

### DIFF
--- a/packages/gitbook/src/components/DocumentView/OpenAPI/style.css
+++ b/packages/gitbook/src/components/DocumentView/OpenAPI/style.css
@@ -35,7 +35,7 @@
 }
 
 .openapi-description.openapi-markdown {
-    @apply prose-sm text-[15px];
+    @apply prose-sm text-[0.938rem];
 }
 
 .openapi-description.openapi-markdown code {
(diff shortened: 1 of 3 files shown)


packages/gitbook/src/components/DocumentView/OpenAPI/style.css
```diff
@@ -35,7 +35,7 @@
 }
 
 .openapi-description.openapi-markdown {
-    @apply prose-sm text-[15px];
+    @apply prose-sm text-[0.938rem];
```
scazan, line 38: wondering if these very specific rem values can be simplified with something from our design system?

gregberge (approved):

packages/gitbook/src/components/DocumentView/OpenAPI/style.css
nolannbiron: We could extend tailwind config to support 13px and 15px, but naming might be tricky since the current scale uses xs, sm, base, and lg

We could add text-13 and text-15, but that would imply a direct 13 * 4 or 15 * 4 scale, following tailwind’s convention. Same for text-3.25 and text-3.75, it's not aligned with tailwind's convention

So arbitrary values are kinda the go-to in tailwind's docs

PR #2829, opened 2025-02-12
Improvements to inline images in headings

zenoachtig:
Fixes the following issues:
- Text with inline images in it unexpectedly gets centered
- Headings don't inherit the line height set on them
- Inline images set to `line` size are displayed a bit too big in relation to the line.

### DIFF
--- a/packages/gitbook/src/components/DocumentView/InlineImage.tsx
+++ b/packages/gitbook/src/components/DocumentView/InlineImage.tsx
@@ -48,7 +48,7 @@ export async function InlineImage(props: InlineProps<DocumentInlineImage>) {
                 }}
                 priority="lazy"
                 preload
-                style={[size === 'line' ? ['max-h-[1.1em]', 'h-[1.1em]', 'mb-[.1em]', 'w-auto'] : null]}
+                style={[size === 'line' ? ['max-h-[1lh]', 'h-[1lh]', 'w-auto'] : null]}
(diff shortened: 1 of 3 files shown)


packages/gitbook/src/components/DocumentView/InlineImage.tsx
```diff
@@ -48,7 +48,7 @@ export async function InlineImage(props: InlineProps<DocumentInlineImage>) {
                 }}
                 priority="lazy"
                 preload
-                style={[size === 'line' ? ['max-h-[1lh]', 'h-[1lh]', 'w-auto'] : null]}
+                style={[size === 'line' ? ['max-h-[1.1em]', 'h-[1.1em]', 'mb-[.1em]', 'w-auto'] : null]}
```
viktorrenkema, line 51: I'm a bit wary of changes on this, make sure to test it really well that it's not causing unexpected issues elsewhere as we've had that before with inline images.

If you look at this visual diff: https://app.argos-ci.com/GitbookIO/gitbook/builds/2392/137281906, the image is getting quite a bit smaller for an already small image. Are you sure this is needed, as it'll impact all published docs with inline images.

zenoachtig: Fair point – I've left out the size change for now, it's now only the line height and justification change.

PR #2831, opened 2025-02-13
Fix card layout for cards without cover

zenoachtig:
I accidentally broke card layouts without covers, here's a fix
Screenshot 2025-02-13 at 14 30 11
Screenshot 2025-02-13 at 14 30 18

### DIFF
--- a/packages/gitbook/src/components/DocumentView/Table/RecordCard.tsx
+++ b/packages/gitbook/src/components/DocumentView/Table/RecordCard.tsx
@@ -54,13 +54,15 @@ export async function RecordCard(
                 // On mobile, check if we can display the cover responsively or not:
                 // - If the file has a landscape aspect ratio, we display it normally
                 // - If the file is square or portrait, we display it left with 40% of the card width
-                coverIsLandscape
-                    ? 'grid-rows-[auto,1fr]'
-                    : [
-                          'grid-cols-[40%,_1fr]',
-                          'min-[432px]:grid-cols-none',
-                          'min-[432px]:grid-rows-[auto,1fr]',
-                      ],
+                cover
+                    ? coverIsLandscape
+                        ? 'grid-rows-[auto,1fr]'
+                        : [
+                              'grid-cols-[40%,_1fr]',
+                              'min-[432px]:grid-cols-none',
+                              'min-[432px]:grid-rows-[auto,1fr]',
+                          ]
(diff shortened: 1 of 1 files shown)


packages/gitbook/src/components/DocumentView/Table/RecordCard.tsx
```diff
@@ -54,13 +54,15 @@ export async function RecordCard(
...
                 // - If the file has a landscape aspect ratio, we display it normally
                 // - If the file is square or portrait, we display it left with 40% of the card width
-                coverIsLandscape
-                    ? 'grid-rows-[auto,1fr]'
-                    : [
-                          'grid-cols-[40%,_1fr]',
-                          'min-[432px]:grid-cols-none',
-                          'min-[432px]:grid-rows-[auto,1fr]',
-                      ],
+                cover
```
valentin0h, line 57:
```suggestion
                cover && coverIsLandscape
```

valentin0h (approved):

packages/gitbook/src/components/DocumentView/Table/RecordCard.tsx
zenoachtig: No I want none of these classes when there is no cover – combining the logic would still render the portrait classes if there is no cover (that's the current bug)

If there's a better way to write this logic statement I'm open to it, I don't like this double ternary

packages/gitbook/src/components/DocumentView/Table/RecordCard.tsx
```diff
@@ -54,13 +54,15 @@ export async function RecordCard(
...
+                    ? coverIsLandscape
+                        ? 'grid-rows-[auto,1fr]'
+                        : [
+                              'grid-cols-[40%,_1fr]',
+                              'min-[432px]:grid-cols-none',
+                              'min-[432px]:grid-rows-[auto,1fr]',
+                          ]
+                    : null,
```
taranvohra, line 65: Nested ternaries are a pain for readability but I guess it's fine here

taranvohra (approved):

packages/gitbook/src/components/DocumentView/Table/RecordCard.tsx
valentin0h: ha I missed the `null`

PR #2835, opened 2025-02-14
Flip card logic

zenoachtig:
From `coverIsLandscape` to the more specific `coverIsSquareOrPortrait`, which means our default can be more sensible. This fixes some bugs where `cover` values are `undefined`. Now we will only render the alternative card layout if we are certain we should, and we fall back to the normal layout.

### DIFF
--- a/packages/gitbook/src/components/DocumentView/Table/RecordCard.tsx
+++ b/packages/gitbook/src/components/DocumentView/Table/RecordCard.tsx
@@ -30,7 +30,7 @@ export async function RecordCard(
 
     const coverIsSquareOrPortrait =
         cover?.file?.dimensions &&
-        cover.file?.dimensions?.width / cover.file?.dimensions?.height < 1;
+        cover.file?.dimensions?.width / cover.file?.dimensions?.height <= 1;
 
     const body = (
         <div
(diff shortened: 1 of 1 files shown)


packages/gitbook/src/components/DocumentView/Table/RecordCard.tsx
```diff
@@ -28,9 +28,9 @@ export async function RecordCard(
         : null;
     const target = targetRef ? await context.resolveContentRef(targetRef) : null;
 
-    const coverIsLandscape =
+    const coverIsSquareOrPortrait =
         cover?.file?.dimensions &&
-        cover.file?.dimensions?.width / cover.file?.dimensions?.height > 1;
+        cover.file?.dimensions?.width / cover.file?.dimensions?.height < 1;
```
valentin0h, line 33: Is square `< 1`?

valentin0h (approved):

packages/gitbook/src/components/DocumentView/Table/RecordCard.tsx
zenoachtig: Ah damn yeah it should be

```suggestion
        cover.file?.dimensions?.width / cover.file?.dimensions?.height <= 1;
```

PR #2845, opened 2025-02-18
Fix export PDF from within a change request

taranvohra:


packages/gitbook/src/middleware.ts
```diff
@@ -528,8 +528,8 @@ async function lookupSiteOrSpaceInMultiIdMode(
     }
 
     // Extract the change request or revision ID from the path
-    const changeRequestId = eatPathId('~', 'changes');
-    const revisionId = eatPathId('~', 'revisions');
+    const changeRequestId = eatPathId('~changes');
```
taranvohra, line 531: https://github.com/GitbookIO/gitbook-x/blob/0ef9f594483da979f2bca5b1ba4db30034c5307e/packages/backend-gitbook-open/src/urls.ts#L33

We were parsing the path incorrectly in GBO that's why `changeRequestId` was undefined

gregberge (approved):

PR #2851, opened 2025-02-19
Show additional fields in OpenAPI block

nolannbiron:
Show additional fields in OpenAPI block like min, max, default

Before: 
CleanShot 2025-02-19 at 12 43 40@2x

After:
CleanShot 2025-02-19 at 12 55 17@2x

### DIFF
--- a/.changeset/shiny-adults-suffer.md
+++ b/.changeset/shiny-adults-suffer.md
@@ -1,0 +1,5 @@
+---
+'@gitbook/react-openapi': patch
+---
+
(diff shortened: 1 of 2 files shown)


gregberge (approved):
What about pattern? https://swagger.io/docs/specification/v3_0/data-models/data-types/#strings

nolannbiron:
> What about pattern? https://swagger.io/docs/specification/v3_0/data-models/data-types/#strings

`Pattern` are handled with `Example`

PR #2852, opened 2025-02-19
Display OpenAPI response examples in multiple languages

nolannbiron:
**DEMO with XML:**
 
CleanShot 2025-02-19 at 14 02 04@2x

### DIFF
--- a/.changeset/smooth-coats-sparkle.md
+++ b/.changeset/smooth-coats-sparkle.md
@@ -1,0 +1,5 @@
+---
+'@gitbook/react-openapi': patch
+---
+
+Display OpenAPI response examples in multiple languages


packages/react-openapi/src/OpenAPIResponseExample.tsx
```diff
@@ -143,3 +140,11 @@ function handleUnresolvedReference(
...
     return input;
 }
+
+function fetchSyntaxFromContentType(contentType: string) {
+    if (contentType.includes('json')) {
+        return 'json';
+    }
+
+    return contentType.split('/')[1] || 'json';
+}
```
gregberge, line 150: The approach is very primitive here.

gregberge (changes requested):
I am not sure it works to be honest and we have to allow to select the multiple languages. I will iterate on this one.

PR #2853, opened 2025-02-19
Upgrade Scalar client

gregberge:
Upgrade OpenAPI Scalar client "Try it" button.

I detected an issue, if we click on "Try it" twice, the second time, the method is not correctly selected.


nolannbiron (approved):

gregberge:
Waiting for the Scalar team to make it compatible with React v18 and to fix the bug of initial operation selection.

gregberge:
Done in another PR.

PR #2856, opened 2025-02-19
Support multiple examples and multiple responses example

gregberge:
CleanShot 2025-02-19 at 17 09 59@2x


nolannbiron:
I think we can remove the border radius for the CodeBlock inside

CleanShot 2025-02-19 at 17 18 47@2x

PR #2860, opened 2025-02-20
Add support for new OpenAPI ref

gregberge:
Add support for the new OpenAPI ref in OpenAPI operation blocks.

### DIFF
--- a/packages/gitbook/src/lib/api.ts
+++ b/packages/gitbook/src/lib/api.ts
@@ -209,6 +209,78 @@ export const getUserById = cache({
 });
 
 /**
+ * Get the latest version of an OpenAPI spec by its slug.
+ */
+export const getLatestOpenAPISpecVersion = cache({
+    name: 'api.getLatestOpenApiSpecVersion',
+    tag: (organization, openAPISpec) =>
+        getAPICacheTag({
+            tag: 'openapi',
+            organization,
+            openAPISpec,
+        }),
+    get: async (organizationId: string, slug: string, options: CacheFunctionOptions) => {
+        try {
+            const apiCtx = await api();
+            const response = await apiCtx.client.orgs.getLatestOpenApiSpecVersion(
+                organizationId,
+                slug,
+                {
+                    ...noCacheFetchOptions,
+                    signal: options.signal,
+                },
+            );
+            return cacheResponse(response, { revalidateBefore: 60 * 60 });
+        } catch (error) {
+            if (checkHasErrorCode(error, 404)) {
+                return {
+                    revalidateBefore: 5,
+                    data: null,
+                };
+            }
+
+            throw error;
+        }
+    },
+});
+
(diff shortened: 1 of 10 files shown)


packages/gitbook/src/lib/api.ts
```diff
@@ -208,6 +208,78 @@ export const getUserById = cache({
...
+/**
+ * Get the latest version of an OpenAPI spec by its slug.
+ */
+export const getLatestOpenAPISpecVersion = cache({
+    name: 'api.getLatestOpenApiSpecVersion',
+    tag: (organization, openAPISpec) =>
+        getAPICacheTag({
+            tag: 'openapi',
+            organization,
+            openAPISpec,
```
emmerich, line 220:
```suggestion
            slug,
```

it's the slug right?

emmerich (approved):

packages/gitbook/src/lib/references.tsx
```diff
@@ -272,6 +277,27 @@ export async function resolveContentRef(
...
         }
 
+        case 'openapi': {
+            if (!siteContext) {
+                return null;
+            }
+            const { organizationId } = siteContext;
+            const [openAPISpecVersion, openAPISpecVersionContent] = await Promise.all([
+                getLatestOpenAPISpecVersion(organizationId, contentRef.spec),
+                getLatestOpenAPISpecVersionContent(organizationId, contentRef.spec),
```
SamyPesse, line 287: This is inefficient, we should have optimized the content API to also return the URL

packages/react-openapi/src/OpenAPIOperation.tsx
```diff
@@ -29,7 +29,7 @@ export function OpenAPIOperation(props: {
 
     return (
         <div className={clsx('openapi-operation', className)}>
-            <div className="openapi-summary">
+            <div className="openapi-summary" id={operation.summary ? undefined : context.id}>
```
SamyPesse, line 32: Why only when `operation.summary` is undefined?

gregberge: Because else it's on the heading.

packages/gitbook/src/lib/references.tsx
gregberge: Yep, it's an optimization we should do.

gregberge: [RND-6305: Add URL to getLatestOpenAPISpecVersionContent](https://linear.app/gitbook-x/issue/RND-6305/add-url-to-getlatestopenapispecversioncontent)

packages/gitbook/src/lib/api.ts
gregberge: Yep and it's the organization ID, since we like not being precise here I continue the on the same vein 😅

PR #2861, opened 2025-02-20
Add link styles

zenoachtig:


packages/gitbook/src/components/DocumentView/InlineLink.tsx
```diff
@@ -23,9 +23,8 @@ export async function InlineLink(props: InlineProps<DocumentInlineLink>) {
     }
 
     return (
-        <Link
+        <StyledLink
```
clairechabas, line 26: For better readability (and easier browsing in folders) we should place the modifier at the end in or names. Could we rename this `LinkStyled`? (can be done in a separate PR ofc)

clairechabas (approved):

PR #2890, opened 2025-02-27
Change how a site in proxy mode is resolved

taranvohra:


packages/gitbook/src/middleware.ts
```diff
@@ -335,7 +342,10 @@ function getInputURL(request: NextRequest): {
     // to determine the site to serve.
     const xGitbookSite = request.headers.get('x-gitbook-site-url');
     if (xGitbookSite) {
-        mode = 'proxy';
+        return {
+            mode: 'proxy',
+            url: new URL(xGitbookSite),
```
taranvohra, line 347: From the `proxy` worker we send the entire URL that the user is trying to access, this is our `inputURL`

packages/gitbook/src/lib/links.ts
```diff
@@ -45,6 +45,11 @@ export async function getBasePath(): Promise<string> {
 export async function getHost(): Promise<string> {
     assertIsNotV2();
     const headersList = await headers();
+    const mode = headersList.get('x-gitbook-mode');
+    if (mode === 'proxy') {
+        return headersList.get('x-forwarded-host') ?? '';
```
taranvohra, line 50: We can't rely on `x-gitbook-host` for `proxy` mode because that value is overwritten by our content worker.

Instead we rely on the `forwarded-host`  because it's using the `inputURL` host https://github.com/GitbookIO/gitbook/blob/685325cdc7e5ef0b9d4f8a85b98a5b658ef8aec2/packages/gitbook/src/middleware.ts#L174

SamyPesse: But why is the request going through the content worker?

Why not directly make the request to `open.gitbook.com` ?

SamyPesse (approved):
Approving, but It looks a bit messy, should we lay down on paper the requests flow to make sure things are consistent?

packages/gitbook/src/lib/links.ts
taranvohra: Because of caching in content worker

PR #2892, opened 2025-02-27
Fix recursion in OpenAPISchemaAlternative

nolannbiron:
Before:
CleanShot 2025-02-27 at 15 35 50@2x

After:
CleanShot 2025-02-27 at 15 36 18@2x


packages/react-openapi/src/OpenAPISchema.tsx
```diff
@@ -173,6 +166,25 @@ function OpenAPISchemaAlternative(props: {
     const id = useId();
     const subProperties = getSchemaProperties(schema);
     const description = resolveDescription(schema);
+    const alternatives = getSchemaAlternatives(schema, new Set(circularRefs?.keys()));
+
+    if (alternatives?.[0]?.length && !subProperties?.length) {
```
gregberge, line 171: Not sure if our logic is good here, why are we ignoring all alternatives except the first one? Anyway let's merge it if it's a ISO fix.

gregberge (approved):

packages/react-openapi/src/OpenAPISchema.tsx
nolannbiron: The alternatives are typed like this :
```
type OpenAPISchemaAlternatives = [
    OpenAPIV3.SchemaObject[],
    OpenAPIV3.DiscriminatorObject | undefined
];
```

So we are checking if there is at least 1 alternative schema
- alternatives[0] = schema[]
- alternatives[1]  = discrimator

nolannbiron: we could change the typing and return an object

PR #2896, opened 2025-02-28
Fix `getProxyModeBasePath`

taranvohra:
There was a bug, I added test cases now


packages/gitbook/src/lib/paths.ts
```diff
@@ -18,3 +18,14 @@ export function removeTrailingSlash(path: string): string {
...
+/**
+ * Normalize a pathname to make it start with a slash
+ */
+export function normalizePathname(pathname: string): string {
+    if (!pathname.startsWith('/')) {
+        pathname = `/${pathname}`;
+    }
+
+    return pathname;
+}
```
taranvohra, line 31: Moved it from the middleware file

gregberge (approved):

PR #2908, opened 2025-03-03
Implement OpenAPI models blocks

nolannbiron:
First implementation for OpenAPI models blocks

### Multiple models
CleanShot 2025-03-05 at 13 20 47@2x

### Single model
CleanShot 2025-03-05 at 13 22 26@2x


packages/openapi-parser/src/traverse.ts
```diff
@@ -29,12 +29,20 @@ export async function traverse<T extends AnyObject | AnyObject[]>(
...
             const value = specification[key];
-            result[key] = await traverse(value, transform, [...path, key], seen);
+            const processed = await traverse(value, transform, [...path, key], seen);
+            return { key, value: processed, index };
         })
     );
 
+    // Promise.all does not guarantee the order of the results
+    // So we need to sort them to preserve the original order
+    promises.sort((a, b) => a.index - b.index);
```
nolannbiron, line 42: Tested the performance of this, it's not a big change. Keeping the original order of the spec is a good thing I think

packages/gitbook/src/lib/openapi/fetch.ts
gregberge, line 1: It makes all the code complicated. Please just separate methods: `resolveOpenAPIOperationBlock` and `resolveOpenAPIModelsBlock`.

packages/openapi-parser/src/traverse.ts
```diff
@@ -29,12 +29,20 @@ export async function traverse<T extends AnyObject | AnyObject[]>(
     }
 
     const keys = Object.keys(specification);
-    await Promise.all(
-        keys.map(async (key) => {
+    const promises = await Promise.all(
```
gregberge, line 32: These are not promises.

packages/react-openapi/src/memoDereferenceFilesystem.ts
gregberge, line 1: Rename the file and call it dereference.ts.

packages/react-openapi/src/models/OpenAPIModels.tsx
```diff
@@ -0,0 +1,91 @@
...
+        </div>
+    );
+}
+
+/**
+ * Root schema for OpenAPI models.
+ * It displays a single model or a disclosure group for multiple models.
+ */
+function OpenAPIRootModelsSchema(props: {
+    models: [string, OpenAPIV3.SchemaObject][];
```
gregberge, line 42: Please don't use an array if possible. It makes the code unreadable.

packages/react-openapi/src/utils.ts
```diff
@@ -10,11 +11,27 @@ export function createStateKey(key: string, scope?: string) {
     return scope ? `${scope}_${key}` : key;
 }
 
+/**
+ * Check if an object has a description. Either at the root level or in items.
```
gregberge, line 15: Don't do the check in array, it's weird and unexpected, why do you need it?

packages/react-openapi/src/utils.ts
```diff
@@ -50,9 +67,20 @@ export function resolveFirstExample(object: AnyObject) {
...
+        return formatExample(object.example);
+    }
+
+    // Resolve example from items if it exists
+    if (object.items && typeof object.items === 'object') {
+        if (Array.isArray(object.items)) {
+            return formatExample(object.items[0].example);
+        }
+        return formatExample(object.items.example);
     }
```
gregberge, line 82: Same it's unexpected to look in an array.

gregberge (changes requested):

packages/gitbook/src/lib/openapi/types.ts
```diff
@@ -0,0 +1,50 @@
...
+type FetchOpenAPIFilesystemError = {
+    error: OpenAPIParseError;
+    filesystem?: undefined;
+    specUrl?: undefined;
+};
+
+/**
+ * Resolved OpenAPI block result.
+ */
+export type ResolveOpenAPIBlockResult<T> =
```
nolannbiron, line 39: @gregberge I made this one generic but I can change, cause only data is different

gregberge (approved):

PR #2916, opened 2025-03-04
Implement route to invalidate data cache

SamyPesse:


packages/gitbook-v2/src/app/~gitbook/revalidate/route.ts
```diff
@@ -0,0 +1,42 @@
...
+}
+
+/**
+ * Revalidate cached data based on tags.
+ * The body should be a JSON with { tags: string[] }
+ */
+export async function POST(req: NextRequest) {
+    const json = (await req.json()) as JsonBody;
+
+    if (GITBOOK_APP_SECRET && req.headers.get('x-gitbook-secret') !== GITBOOK_APP_SECRET) {
```
gregberge, line 17: This is a weak security. The good security would be to sign the body with the secret.

gregberge (approved):

PR #2920, opened 2025-03-04
Fix OpenAPI spec recursion

gregberge:


packages/react-openapi/src/OpenAPISchema.tsx
```diff
@@ -21,84 +20,60 @@ export interface OpenAPISchemaPropertyEntry {
...
+            <OpenAPISchemaPresentation property={property} />
+            {(() => {
+                const parentCircularRef = parentCircularRefs.get(schema);
+
+                // Avoid recursing infinitely, and instead render a link to the parent schema
+                if (parentCircularRef) {
+                    return <OpenAPISchemaCircularRef id={parentCircularRef} schema={schema} />;
+                }
+
+                const circularRefs = parentCircularRefs.set(schema, id);
```
gregberge, line 51: I am mutating the parentCircularRefs which is bad but I don't understand why it does not work else...

I suspect streaming and Suspense to cut the rendering, start it again and never reach a point where it works.

nolannbiron (approved):

PR #2946, opened 2025-03-07
Rename OpenAPIModels to OpenAPISchemas

nolannbiron:


packages/react-openapi/src/types.ts
```diff
@@ -61,7 +61,7 @@ export type OpenAPIModel = {
     schema: OpenAPIV3.SchemaObject;
 };
 
-export interface OpenAPIModelsData {
-    /** Components schemas to be used for models */
-    models: OpenAPIModel[];
+export interface OpenAPISchemasData {
+    /** Components schemas to be used for schemas */
+    schemas: OpenAPIModel[];
```
gregberge, line 66: Oversight?

gregberge (approved):

packages/react-openapi/src/types.ts
nolannbiron: Forgot this one, replaced it with OpenAPISchema but it might be too generic ? wdyt ?

gregberge: We can change it later, is it exported anywhere or internal?

PR #2953, opened 2025-03-10
Info hint background and link color fixes

zenoachtig:
This PR fixes a few small issues with the hint block and links within hint blocks.

- The `info` hint block did not properly use the set `info` semantic colour, now it does.
- Links inside hint blocks should be coloured according to the semantic colour, with the exception of the `info` block, which uses the normal link colour. They now correctly follow this behaviour. (Fixes RND-6390)
- Links using the `accent` link style had an underline that was too thin, it's now explicitly set to `1px`.

# Before & After
Screenshot 2025-03-10 at 13 44 35
Screenshot 2025-03-10 at 13 44 27


packages/gitbook/src/components/DocumentView/Hint.tsx
```diff
@@ -104,20 +104,23 @@ const HINT_STYLES: {
...
         body: [
             'text-neutral-strong',
-            '[&_a]:text-warning',
-            '[&_a:hover]:text-warning-strong',
+            'links-default:[&_a]:text-warning',
+            'links-default:[&_a]:hover:text-warning-strong',
+            'links-default:[&_a]:decoration-warning/6',
+            'links-accent:[&_a]:decoration-warning',
+            'decoration-warning/6',
+            '[&_a:hover]:!text-warning-strong',
```
viktorrenkema, line 121: Should this one not also be applied to the 'danger' and 'success' variants right below?

viktorrenkema (approved):

packages/gitbook/src/components/DocumentView/Hint.tsx
zenoachtig: Whoops, this line here is a holdout, covered by the `'links-default:[&_a]:hover:text-warning-strong'` above. The `!` was a bit of a hack anyway, I'm using the customisation link classes correctly now. I'll remove this line.
```suggestion
```

PR #2960, opened 2025-03-11
Support Node.js v20 with @gitbook/icons

gregberge:


packages/icons/src/Icon.tsx
```diff
@@ -2,9 +2,9 @@
 
 import * as React from 'react';
 
-import { getIconAssetURL, useIcons } from './IconsProvider';
-import { getIconStyle } from './getIconStyle';
-import type { IconName, IconStyle } from './types';
+import { getIconAssetURL, useIcons } from './IconsProvider.js';
```
SamyPesse, line 5: There is a Typescript option to avoid having to put `.js` at every file.

We are not going to do this in our entire codebase, so let's be consistent here as well

SamyPesse (changes requested):

SamyPesse (approved):

PR #2968, opened 2025-03-11
Update target in shiki javascript engine

nolannbiron:


nolannbiron:
@gregberge the results when using ES2024 are inconsistent. I ran the test more than 100 times with a script, and up to 5% of the runs failed (because shiki return `console.` instead of `console`). 

However, when switching to ES2018 (as recommended for maximum compatibility in Shiki’s documentation), the tests run significantly faster (200ms instead of 700ms), and I can no longer reproduce the 5% failure.

gregberge (approved):

PR #2976, opened 2025-03-12
Fix Python code sample JSON body

rodrcastro:
When sending a JSON body in the request, it should use `json`. This PR changes it to make the code sample request correct.


nolannbiron:
@rodrcastro We are using `data` instead of `json` to handle any content type.

If you want to send `text/csv`, `text/plain` or `multipart`, you can’t use `json`, it’s just a shorthand that implicitly sets the content type to `application/json`.

`data` works with any content type, as long as you specify it in the headers.

nolannbiron:
but we can set it to `json` for `application/json`, if customers really want to see this instead of `data`

SamyPesse (approved):

PR #2977, opened 2025-03-12
Simplify deferred codeblock highlighting and fix streaming

SamyPesse:


packages/gitbook/src/components/DocumentView/CodeBlock/ClientCodeBlock.tsx
```diff
@@ -22,53 +20,56 @@ type ClientBlockProps = Pick<BlockProps<DocumentBlockCode>, 'block' | 'style'> &
...
+                setIsInViewport(true);
             }
-            isInViewportRef.current = isInViewport;
         },
         { rootMargin: '200px' }
     );
 
-    const handleScroll = useDebounceCallback(() => {
-        if (isInViewportRef.current) {
-            debouncedRunHighlight();
```
SamyPesse, line 67: @gregberge I don't understand what you tried to do here: if `isInViewportRef.current` is true, then you already ran the highlighting when it becomes visible, why defer another highlighting?

gregberge: I trigger highlighting on scroll and not from the observer, so it's smoother.

But the main issue on scroll is the outline, not that so I think you can simplify it.

gregberge: The main problem is triggering compute while the user scrolls. It's not good. I wanted to highlight when the user pause the scroll.

gregberge: But are you sure it's the issue? Because the blocks are empty if it was the issue they would be not highlighted but displayed at least.

gregberge: Because the initial presence in viewport was working so, I have some doubts on this fix.

gregberge (approved):

PR #2981, opened 2025-03-14
Add support for custom fonts

valentin0h:
Adds support for custom fonts in GitBook by dynamically generating `@font-face` rules from user-defined font configurations and preloading the key font sources. It integrates these custom fonts into the layout and OG image generation


packages/gitbook-v2/src/lib/env/globals.ts
```diff
@@ -64,6 +64,11 @@ export const GITBOOK_DISABLE_TRACKING = Boolean(
 export const GITBOOK_INTEGRATIONS_HOST =
     process.env.GITBOOK_INTEGRATIONS_HOST || 'integrations.gitbook.com';
 
+/**
+ * Hostname for fonts.
+ */
+export const GITBOOK_FONTS_URL = process.env.GITBOOK_FONTS_URL || 'https://fonts.gitbook.com';
```
SamyPesse, line 70: You need to update the `.github/workflows` to pass this env; and also add the value to the vault in 1Password. (So that staging can use the proper value)

packages/gitbook/src/components/RootLayout/CustomizationRootLayout.tsx
```diff
@@ -66,14 +67,37 @@ export async function CustomizationRootLayout(props: {
...
             <head>
                 {customization.privacyPolicy.url ? (
                     <link rel="privacy-policy" href={customization.privacyPolicy.url} />
                 ) : null}
+
+                {/* Custom font resources */}
+                {fontData.type === 'custom' && (
+                    <>
+                        {/* Font host preconnect */}
+                        <link rel="preconnect" href={GITBOOK_FONTS_URL} />
```
SamyPesse, line 82: I think it's better to use `ReactDOM.preconnect` (https://react.dev/reference/react-dom/preconnect), it can be called conditionnaly.

You should also use `ReactDOM.preload` (https://react.dev/reference/react-dom/preload)

packages/gitbook/src/components/RootLayout/CustomizationRootLayout.tsx
```diff
@@ -66,14 +67,37 @@ export async function CustomizationRootLayout(props: {
...
+                                rel="preload"
+                                href={url}
+                                as="font"
+                                type={format ? `font/${format}` : undefined}
+                                crossOrigin="anonymous"
+                            />
+                        ))}
+
+                        {/* Custom font CSS */}
+                        <style id="custom-font-styles">{fontData.cssDefinitions}</style>
```
SamyPesse, line 97: Why do we need an `id`?

packages/gitbook/src/components/RootLayout/CustomizationRootLayout.tsx
SamyPesse: It'll make sure it puts as high as possible in the HTML parsing

SamyPesse (commented):
Looks good, i'm waiting for the code to cleanup by removing the types

packages/gitbook/src/components/RootLayout/CustomizationRootLayout.tsx
valentin0h: Easier to debug; removed.

packages/gitbook/src/components/RootLayout/CustomizationRootLayout.tsx
```diff
@@ -48,6 +50,22 @@ export async function CustomizationRootLayout(props: {
     const mixColor = getTintMixColor(customization.styling.primaryColor, tintColor);
     const sidebarStyles = getSidebarStyles(customization);
     const { infoColor, successColor, warningColor, dangerColor } = getSemanticColors(customization);
+    const fontData = getFontData(customization.styling.font);
+
+    // Preconnect and preload custom fonts if needed
+    if (fontData.type === 'custom') {
+        ReactDOM.preconnect(GITBOOK_FONTS_URL);
```
emmerich, line 57: How does this work on the server? Preloading in Nextjs seems to have different recommendations: https://nextjs.org/docs/app/building-your-application/data-fetching/fetching#preloading-data

packages/gitbook/src/fonts/custom.ts
```diff
@@ -0,0 +1,56 @@
...
+
+                    if (source.format) {
+                        srcDefinition += ` format('${source.format}')`;
+                    }
+
+                    return srcDefinition;
+                })
+                .join(', ');
+
+            // We could use the font-family name here, but to avoid extra normalization we're using 'CustomFont'
```
emmerich, line 24: Hmm I wonder if this will have a wider impact as the underlying font data might have a different name. Should be fine for now though.

emmerich (commented):
Overall looks good, I'm not sure about the use of `ReactDOM.preload` in SSR though

packages/gitbook/src/fonts/custom.ts
valentin0h: I deliberately hardcoded the font family to “CustomFont” to avoid the complexity of normalizing the user input. I'm not sure it matters about the underlying font data names, as we always reference the custom font via our standardized CSS variable and class. Unless I'm missing something, lemme know.

packages/gitbook/src/components/RootLayout/CustomizationRootLayout.tsx
valentin0h: I'm not sure. The next preload mechanism appears to be more geared toward preloading next data rather than external resources like fonts. I haven't seen any examples in the wild using this pattern to preload custom fonts.

package.json
```diff
@@ -12,7 +12,7 @@
         "@codemirror/state": "6.4.1",
         "react": "18.3.1",
         "react-dom": "18.3.1",
-        "@gitbook/api": "0.99.0"
+        "@gitbook/api": "0.100.0"
```
SamyPesse, line 15: `"0.102.0"` in main now

packages/gitbook/src/components/RootLayout/CustomizationRootLayout.tsx
SamyPesse: @emmerich the link you are referring to is not related to this

The actual documentation is: https://react.dev/reference/react-dom/preconnect

The way it works is that Next will emit automatically an HTML tag, ex where we use it for other static assets

Screenshot 2025-03-21 at 20 53 23

When the browser is parsing the HTML, it'll start connecting to the server even before finding any CSS referencing this font (https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/preconnect)

With `ReactDOM.preload` below, it'll start fetching the actual font file as well.

SamyPesse (approved):
Looks great 👍

PR #2997, opened 2025-03-19
Disable animations on tests

zenoachtig:


gregberge (changes requested):
It's already the default: https://github.com/argos-ci/argos-javascript/blob/523719ebd3b4035daa1584b59f6e03d9aec10518/packages/playwright/src/screenshot.ts#L366

zenoachtig:
Closed in favour of #2999

PR #2998, opened 2025-03-19
v2: use static route for VA / adaptive content

SamyPesse:
This PR aims to use static routes for VA and adaptive content. The authentication is done in the middleware already with `getPublishedContentByUrl`, but now the site data from it are passed in the URL using a Rison encoded route parameter.

Since the `apiToken` from `getPublishedContentByUrl` are now stable over period of time, it should not cause cache misses.


packages/gitbook/src/app/middleware/(site)/(content)/layout.tsx
```diff
@@ -22,14 +23,14 @@ export default async function ContentLayout(props: { children: React.ReactNode }
...
     const context = await fetchLayoutData();
     const queryStringTheme = await getThemeFromMiddleware();
-    const visitorAuthToken = await getVisitorAuthTokenFromMiddleware();
+    const siteData = await getSiteURLDataFromMiddleware();
 
     return (
         <SiteLayout
             context={context}
             forcedTheme={queryStringTheme}
             withTracking={await shouldTrackEvents()}
```
gregberge, line 32: Sure we don't want `Promise.all` here?

gregberge (approved):

packages/gitbook/src/app/middleware/(site)/(content)/layout.tsx
SamyPesse: yes, it has no impact on perfs, it's just a `await headers()`, Next uses `await as a way to detect the calls and trigger the route to be dynamic

PR #3008, opened 2025-03-20
Support site announcement banner

zenoachtig:


packages/gitbook/src/components/Announcement/AnnouncementBanner.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+    const hasLink = announcement.link && contentRef?.href;
+    const closeable = announcement.style !== 'danger';
+
+    const Tag = hasLink ? Link : 'div';
+    const style = BANNER_STYLES[announcement.style];
+
+    if (!announcement.message) {
+        return null;
+    }
```
valentin0h, line 28: Is this allowed?

packages/gitbook/src/components/Announcement/style.css
valentin0h, line 1: We can remove it, it's done in the `global.css`

packages/gitbook/src/components/SpaceLayout/SpaceLayout.tsx
```diff
@@ -62,6 +63,7 @@ export function SpaceLayout(props: {
                 spaceId={context.space.id}
                 visitorAuthClaims={visitorAuthClaims}
             >
+                {customization.announcement?.enabled ? <Announcement context={context} /> : null}
```
valentin0h, line 66: I suggest so that we can handle the `null` returns in one place, in `Announcement`
```suggestion
                <Announcement context={context} />
```

packages/gitbook/src/components/Announcement/Announcement.tsx
```diff
@@ -0,0 +1,29 @@
+import { resolveContentRef } from '@/lib/references';
+import type { GitBookSiteContext } from '@v2/lib/context';
+import { AnnouncementBanner } from './AnnouncementBanner';
+
```
valentin0h, line 4: a comment here would be great

packages/gitbook/src/components/Announcement/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './Announcement';
+export * from './AnnouncementBanner';
```
valentin0h, line 2:
```suggestion
```

packages/gitbook/src/components/RootLayout/CustomizationRootLayout.tsx
```diff
@@ -96,6 +97,10 @@ export async function CustomizationRootLayout(props: {
                 {/* Inject custom font @font-face rules */}
                 {fontData.type === 'custom' ? <style>{fontData.fontFaceRules}</style> : null}
```
valentin0h, line 99: - [ ] @valentin0h add comment

packages/gitbook/src/components/primitives/StyledLink.tsx
```diff
@@ -2,36 +2,34 @@ import { type ClassValue, tcls } from '@/lib/tailwind';
...
+    'links-default:contrast-more:hover:text-primary-strong',
+
+    'links-accent:decoration-primary-subtle',
+    'links-accent:hover:decoration-[3px]',
+    'links-accent:hover:[text-decoration-skip-ink:none]',
+
+    'transition-all',
+    'duration-100',
+];
+
```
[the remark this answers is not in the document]
zenoachtig: We could use a shared class, but I'd prefer to keep the component styling inside this component rather than somewhere in `global.css`

valentin0h (approved):
looks good 👌 

we need to run changeset

packages/gitbook/src/components/Announcement/AnnouncementDismissedScript.tsx
```diff
@@ -0,0 +1,63 @@
...
+            const announcementStateStr = window.localStorage.getItem(storageKey);
+            const announcementState = announcementStateStr
+                ? JSON.parse(announcementStateStr)
+                : undefined;
+
+            if (announcementState && !announcementState.visible) {
+                const dismissedAt = announcementState.at;
+                const nowTime = new Date().getTime();
+
+                // Check if enough days have passed since dismissal
```
SamyPesse, line 34: We should anticipated this and included a timestamp or a "version id" in the announcement site data

So that instead of being based on date, we could make sure a new banner will display if you update the announcement.

One solution could be to create a hash of the announcement data and use this: if the hash is different than the one closed before, we show it

packages/gitbook/src/components/Announcement/AnnouncementDismissedScript.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Inject a script to read the local storage state for the announcement banner and apply the appropriate CSS class to the <html> element as early as possible.
+ * Bypasses react state to prevent flickering.
+ */
+export function AnnouncementDismissedScript() {
+    /**
+     * Read the local storage state for the announcement banner and apply the appropriate CSS class to the <html> element.
+     */
+    function checkStorageForDismissedScript(
```
SamyPesse, line 17: You can move this function out of the component as it's constant.

Usually I'd recommend having it in another script to make sure we don't have side effects of some imports.

packages/gitbook/src/components/primitives/StyledLink.tsx
```diff
@@ -2,36 +2,34 @@ import { type ClassValue, tcls } from '@/lib/tailwind';
 
 import { Link, type LinkProps } from '../primitives/Link';
 
+export const linkStyles = [
```
SamyPesse, line 5: Why do we exporrt these styles instead of using the component?

packages/gitbook/src/components/Announcement/AnnouncementDismissedScript.tsx
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    const scriptArgs = JSON.stringify([
+        ANNOUNCEMENT_STORAGE_KEY,
+        ANNOUNCEMENT_DAYS_TILL_RESET,
+        ANNOUNCEMENT_CSS_CLASS,
+    ]).slice(1, -1);
+
+    return (
+        <script
```
gregberge, line 56: nonce? I am not sure it will work like that with our CSP

SamyPesse (approved):

packages/gitbook/src/components/Announcement/AnnouncementDismissedScript.tsx
valentin0h: > One solution could be to create a hash of the announcement data and use this: if the hash is different than the one closed before, we show it

Yeah, I stayed away from it to keep the script lean. But can do.

zenoachtig: We can also still add it to the API if that's better

packages/gitbook/src/components/primitives/StyledLink.tsx
zenoachtig: Because we need them in the announcement banner to style the CTA `<span>`.

When you set a link on the announcement banner, the entire banner links to it, but we _also_ support a CTA that is styled like a link so people have a clear click target. However, you can't nest an `<a>` inside another `<a>` so that CTA is a `<span>` that looks like a styled link, hence uses the same styes. I didn't want to duplicate them or put them in a separate class in `globals.css`. I put this as a comment in the file but I guess I deleted it at some point.

Another option would be to use the `<StyledLinks>` component but make it conditionally not use a `<Link>` component within.

packages/gitbook/src/components/Announcement/AnnouncementDismissedScript.tsx
valentin0h: Yes let's do it on the API side separately.

packages/gitbook/src/components/Announcement/AnnouncementDismissedScript.tsx
valentin0h: It should work

https://github.com/GitbookIO/gitbook/blob/zeno/rnd-6578-site-announcement-banner-gbo-support/packages/gitbook/src/lib/csp.ts#L9

PR #3024, opened 2025-03-21
Fix support for site redirects that include section/variant paths

SamyPesse:


packages/gitbook/src/components/SitePage/fetch.ts
```diff
@@ -67,16 +68,31 @@ async function resolvePage(context: GitBookSiteContext, params: PagePathParams |
...
-                organizationId,
-                siteId: site.id,
-                source: rawPathname.startsWith('/') ? rawPathname : `/${rawPathname}`,
-                siteShareKey: shareKey,
-            })
-        );
-        if (resolvedSiteRedirect) {
-            return redirect(resolvedSiteRedirect.target);
+        const redirectPathname = withLeadingSlash(rawPathname);
+        if (/^\/[a-zA-Z0-9-_.\/]+[a-zA-Z0-9-_.]$/.test(redirectPathname)) {
```
taranvohra, line 72: Use `SITE_REDIRECT_SOURCE_PATH_PATTERN` from `@gitbook/api-client`

taranvohra (approved):

packages/gitbook/src/components/SitePage/fetch.ts
SamyPesse: This is only in gbx, it's not the same client here

PR #3037, opened 2025-03-25
Fix OpenAPI spec rendering

gregberge:


packages/gitbook/src/components/DocumentView/Blocks.tsx
```diff
@@ -65,7 +65,7 @@ export function UnwrappedBlocks<TBlock extends DocumentBlock>(props: UnwrappedBl
 
         return (
             <Block
-                key={node.key}
+                key={node.key || `${node.type}-${index}`}
```
SamyPesse, line 68: What is the case where we don't have a `key`? it indicates that we do `serialize()` without `preserveKey` in GBX, not a big issue but we need to make sure we don't depend on it anywhere else

packages/react-openapi/src/OpenAPISchema.tsx
```diff
@@ -1,3 +1,7 @@
+'use client';
+// This component does not use any client feature but we don't want to
+// render it server-side because it has recursion.
```
SamyPesse, line 3: Does it still mean that the first rendering on the client side will do all the recursion? or do we do them only when clicking/expanding?

It might impact client side perfs

SamyPesse (approved):
Looks good, but I have the impression we moved a perf bottleneck to the client side?

nolannbiron (approved):

packages/gitbook/src/components/DocumentView/Blocks.tsx
gregberge: The type is `key?` so this is needed, but yeah we should have keys here.

packages/react-openapi/src/OpenAPISchema.tsx
gregberge: Rendering is only when we expand now.

PR #3046, opened 2025-03-26
Safe parse OpenAPI JSON schema

nolannbiron:


packages/react-openapi/src/OpenAPISchema.tsx
```diff
@@ -172,12 +172,7 @@ export function OpenAPIRootSchemaFromServer(props: {
     schema: string;
     context: OpenAPIClientContext;
 }) {
-    return (
-        <OpenAPIRootSchema
-            schema={JSON.parse(props.schema, retrocycle())}
-            context={props.context}
-        />
-    );
+    return <OpenAPIRootSchema schema={safeJSONParse(props.schema)} context={props.context} />;
```
SamyPesse, line 175: I'm missing context here, does it mean we parse the entire props passed from a server component?

It must be huge in the HTML, why are we doing this?

SamyPesse (approved):
Approving to unblock

PR #3049, opened 2025-03-26
Replace $ref with $reference in json-decycle

nolannbiron:


packages/react-openapi/src/decycle.ts
```diff
@@ -0,0 +1,80 @@
...
+    return Object.defineProperties(json, {
+        decycle: {
+            value: (object: any, space: string | number): string =>
+                json.stringify(object, decycle(), space),
+        },
+        retrocycle: {
+            value: (s: string): any => json.parse(s, retrocycle()),
+        },
+    });
+};
```
gregberge, line 80: Dead code

gregberge (approved):

PR #3052, opened 2025-03-26
Add loading indicator

gregberge:
https://github.com/user-attachments/assets/38c6a841-b8ef-44a8-b4a5-60a4c0c092d9

### DIFF
--- a/packages/gitbook/src/components/RootLayout/CustomizationRootLayout.tsx
+++ b/packages/gitbook/src/components/RootLayout/CustomizationRootLayout.tsx
@@ -21,6 +21,7 @@ import {
     hexToRgb,
 } from '@gitbook/colors';
 import { IconStyle, IconsProvider } from '@gitbook/icons';
+import HolyLoader from 'holy-loader';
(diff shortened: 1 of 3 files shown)


packages/gitbook/src/components/RootLayout/CustomizationRootLayout.tsx
```diff
@@ -151,6 +151,7 @@ export async function CustomizationRootLayout(props: {
                     }
                 `}</style>
             </head>
+            <HolyLoader color="rgb(var(--primary-9))" height={1} />
```
taranvohra, line 154: Honestly I never liked these loaders. They're not very apparent and user friendly 😅. Why not try to use the skeleton grid we had in v1? It looks much better!

gregberge: I wanted to submit a PR with it to see if it works.

Now that we are sure we can detect the loading of a page we can adapt to do whatever we want!

PR #3060, opened 2025-03-27
Use the site canonical URL when rewriting in the middleware

taranvohra:


packages/gitbook-v2/src/middleware.ts
```diff
@@ -152,6 +152,10 @@ async function serveSiteRoutes(requestURL: URL, request: NextRequest) {
             );
         }
 
+        // We use the host/origin from the canonical URL to ensure the links are
+        // correctly generated when the site is proxied. e.g. https://proxy.gitbook.com/site/siteId/...
+        const siteCanonicalURL = new URL(siteURLData.canonicalUrl);
```
SamyPesse, line 157: Then let's rename `siteURL` to `inputURL` or something to avoid confusion

SamyPesse (approved):
Looks good!

PR #3063, opened 2025-03-28
Fix image resizing when using proxy by passing signing identifier in site URL data

SamyPesse:


packages/gitbook-v2/src/lib/context.ts
```diff
@@ -25,6 +25,30 @@ import { GITBOOK_URL } from './env';
...
+    | 'changeRequest'
+    | 'shareKey'
+    | 'siteSection'
+    | 'siteBasePath'
+    | 'basePath'
+> & {
+    /**
+     * Identifier used for image resizing.
+     */
+    imagesSiteId: string;
```
taranvohra, line 49:
```suggestion
    imageContextId: string;
```
Since it's only `siteId` for proxy mode but `host` normally so maybe something generic.

packages/gitbook-v2/src/middleware.ts
```diff
@@ -21,15 +21,17 @@ import {
...
+import type { SiteURLData } from './lib/context';
 
 export const config = {
     matcher: [
         '/((?!_next/static|_next/image|~gitbook/static|~gitbook/revalidate|~gitbook/monitoring|~scalar/proxy).*)',
     ],
 };
 
-type URLWithMode = { url: URL; mode: 'url' | 'url-host' };
+type URLWithMode = { url: URL; mode: 'url' | 'url-host' | 'url-proxy' };
```
taranvohra, line 34: Why we need this? We're not setting it anywhere

packages/gitbook-v2/src/lib/images/getImageResizingContextId.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Get the site identifier to use for image resizing for an incoming request.
+ * This identifier can be obtained before resolving the request URL.
+ */
+export function getImageResizingContextId(url: URL): string {
+    if (url.host === 'proxy.gitbook.site' || url.host === 'proxy.gitbook-staging.site') {
+        // For proxy requests, we extract the site ID from the pathname
+        // e.g. https://proxy.gitbook.com/site/siteId/...
+        const pathname = url.pathname.slice(1).split('/');
+        return pathname.slice(0, 2).join('/');
```
taranvohra, line 10: This includes `site/siteId`, are we sure we want that instead of just `siteId`?

taranvohra: Hm maybe that's good because that means the proxy URL format hasn't changed 👍

taranvohra (approved):

packages/gitbook-v2/src/lib/images/getImageResizingContextId.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Get the site identifier to use for image resizing for an incoming request.
+ * This identifier can be obtained before resolving the request URL.
+ */
+export function getImageResizingContextId(url: URL): string {
+    if (url.host === 'proxy.gitbook.site' || url.host === 'proxy.gitbook-staging.site') {
+        // For proxy requests, we extract the site ID from the pathname
+        // e.g. https://proxy.gitbook.com/site/siteId/...
```
taranvohra, line 8:
```suggestion
        // e.g. https://proxy.gitbook.site/site/siteId/...
```

PR #3064, opened 2025-03-28
Support for x-stability property

nolannbiron:
DEMO :

https://github.com/user-attachments/assets/efacedd3-00de-46ae-a15c-10b664f05c60

### DIFF
--- a/packages/react-openapi/src/OpenAPIOperation.tsx
+++ b/packages/react-openapi/src/OpenAPIOperation.tsx
@@ -1,6 +1,10 @@
 import clsx from 'clsx';
 
-import type { OpenAPICustomOperationProperties, OpenAPIV3 } from '@gitbook/openapi-parser';
+import type {
+    OpenAPICustomOperationProperties,
+    OpenAPIStability,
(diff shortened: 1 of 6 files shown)


packages/react-openapi/src/OpenAPIOperation.tsx
```diff
@@ -29,14 +33,24 @@ export function OpenAPIOperation(props: {
...
+                        )}
+                        {operation['x-stability'] && (
+                            <OpenAPIOperationStability stability={operation['x-stability']} />
+                        )}
+                    </div>
+                )}
                 {operation.summary
                     ? context.renderHeading({
                           deprecated: operation.deprecated ?? false,
+                          stability: operation['x-stability'],
```
gregberge, line 49: Why do you need that here?

nolannbiron: I've put the tags (deprecated/stability) above the heading, so I need both to style the heading in case a tag is displayed

gregberge (approved):

PR #3111, opened 2025-04-07
Sanitize the back URL used when rendering PDF

SamyPesse:
Also cleans up a lot of code from v1


packages/gitbook/src/lib/v1.ts
gregberge, line 1: Where do you use the added methods?

gregberge (approved):

packages/gitbook/src/lib/v1.ts
SamyPesse: There are only used in this file, I'm compressing a bit more the code to make sure that v1 only stuff are only in `v1.ts`

PR #3112, opened 2025-04-07
Reduce the payload from code blocks

gregberge:
**Before**
CleanShot 2025-04-07 at 16 49 02@2x

**After**
CleanShot 2025-04-07 at 16 49 59@2x

Fix RND-6822

### DIFF
--- a/packages/gitbook/src/components/DocumentView/CodeBlock/tree.ts
+++ b/packages/gitbook/src/components/DocumentView/CodeBlock/tree.ts
@@ -1,0 +1,35 @@
+import type { DocumentBlockCode, DocumentInlineAnnotation, DocumentText } from '@gitbook/api';
+
+export type LightNode = (
+    | {
+          text: string;
+      }
+    | {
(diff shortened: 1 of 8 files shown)


packages/gitbook/src/components/DocumentView/CodeBlock/tree.ts
```diff
@@ -0,0 +1,35 @@
+import type { DocumentBlockCode, DocumentInlineAnnotation, DocumentText } from '@gitbook/api';
+
+export type LightNode = (
```
SamyPesse, line 3: I was wondering if we have a better name for it, and if at some point we should merge this as something we could do in `getDocument` to also solve https://linear.app/gitbook-x/issue/RND-6497/optimize-document-and-revision-data-cache-by-reducing-data-size

SamyPesse: Fine for now

SamyPesse (approved):

PR #3119, opened 2025-04-08
Fix ogimage generation crashing when site is using a custom WOFF2 font

SamyPesse:


packages/gitbook/src/routes/ogimage.tsx
```diff
@@ -89,12 +89,15 @@ export async function serveOGImage(baseContext: GitBookSiteContext, params: Page
             await Promise.all(
                 primaryFontWeights.map((face) => {
                     const { weight, sources } = face;
-                    if (sources.length === 0) {
+                    const source = sources[0];
```
gregberge, line 92: Why are we taking the first source like that? Shouldn't we iterate? What are the sources?

SamyPesse: I' not sure tbh, I'm just adapting the code not to crash. Maybe @emmerich knows?

PR #3126, opened 2025-04-09
Fix the site canonical URL to include the visitor token if necessary

taranvohra:
Otherwise VA sites using the proxy would not remove the token from the URL

### DIFF
--- a/packages/gitbook-v2/src/middleware.ts
+++ b/packages/gitbook-v2/src/middleware.ts
@@ -17,6 +17,7 @@ import { serveResizedImage } from '@/routes/image';
 import {
     DataFetcherError,
     getPublishedContentByURL,
+    getSiteCanonicalURL,
     getVisitorAuthBasePath,
     normalizeURL,
     throwIfDataError,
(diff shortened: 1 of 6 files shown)


packages/gitbook-v2/src/middleware.ts
```diff
@@ -146,20 +147,17 @@ async function serveSiteRoutes(requestURL: URL, request: NextRequest) {
...
         // The token is stored in a cookie that is set on the redirect response
         //
         const incomingURL = mode === 'url' ? requestURL : siteCanonicalURL;
-        const requestURLWithoutToken = normalizeVisitorAuthURL(incomingURL);
-        if (
-            requestURLWithoutToken !== incomingURL &&
-            requestURLWithoutToken.toString() !== incomingURL.toString()
-        ) {
+        const incomingURLWithoutToken = normalizeVisitorAuthURL(incomingURL);
+        if (incomingURLWithoutToken.toString() !== incomingURL.toString()) {
```
taranvohra, line 158: For proxy requests our `incoming` URL is the `siteCanonicalURL` but it does not include the va token.  So a VA site when proxied would remove the token from the URL after a successful resolution. But with this PR it fixes that

taranvohra:
Closing in favor of #3127

PR #3127, opened 2025-04-09
Fix incoming URL for requests that were proxied

taranvohra:
Otherwise VA sites using the proxy feature would not get rid of the token from the main URL

### DIFF
--- a/packages/gitbook-v2/src/lib/data/urls.test.ts
+++ b/packages/gitbook-v2/src/lib/data/urls.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, it } from 'bun:test';
 
-import { getIncomingURL, getURLLookupAlternatives, normalizeURL } from './urls';
+import { getURLLookupAlternatives, normalizeURL } from './urls';
 
 describe('getURLLookupAlternatives', () => {
     it('should return all URLs up to the root', () => {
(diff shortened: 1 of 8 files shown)


packages/gitbook-v2/src/middleware.ts
```diff
@@ -148,18 +148,15 @@ async function serveSiteRoutes(requestURL: URL, request: NextRequest) {
...
 
+        const incomingURL = getIncomingURL(requestURL, mode, siteCanonicalURL);
         //
         // Make sure the URL is clean of any va token after a successful lookup
         // The token is stored in a cookie that is set on the redirect response
         //
-        const incomingURL = mode === 'url' ? requestURL : siteCanonicalURL;
-        const requestURLWithoutToken = normalizeVisitorAuthURL(incomingURL);
-        if (
-            requestURLWithoutToken !== incomingURL &&
```
taranvohra, line 158: `URL` check is useless since the string comparison is what we needed

packages/gitbook-v2/src/lib/data/urls.ts
```diff
@@ -110,6 +112,26 @@ export function getURLLookupAlternatives(input: URL) {
     return { urls: alternatives, basePath, changeRequest, revision };
 }
 
+/**
+ * Get the incoming URL for a request including requests
+ * that may be proxied.
+ */
+export function getIncomingURL(
```
SamyPesse, line 119: I don't understand why we make it a function, instead of just fixing it in the middleware and having an e2e. As a function it's a lot more confusing to connect the dots: what does "Incomig URL" means globally?

`incomingURL` was just a variable name in the middleware, and now it's being introduced as a big thing

packages/gitbook-v2/src/lib/data/urls.test.ts
```diff
@@ -364,6 +364,32 @@ describe('getURLLookupAlternatives', () => {
     });
 });
 
+describe('getIncomingURL', () => {
```
SamyPesse, line 367: I think these tests are useless, and instead we should have a e2e test

SamyPesse (approved):

PR #3134, opened 2025-04-10
Fix cache revalidation for computed revisions

SamyPesse:
We considered all revisions to be immutable, even when they were computed. With this PR, we'll leverage the cache-tags provided by the API, to ensure the tag is invalidated when the sources of the computations are updated.

This PR should be combined with one in the API to pass tags for every revision APIs.

On Cloudflare, we'll still use `unstable_cache` and it'll not work properly for computed content (until https://github.com/opennextjs/opennextjs-aws/pull/820 is ready).
On Vercel, it'll use `use cache` and the tags are based on the API response.


packages/gitbook/src/lib/api.ts
```diff
@@ -427,6 +427,7 @@ export const getRevisionPages = cache({
             revisionId,
             {
                 metadata: fetchOptions.metadata,
+                computed: false,
```
gregberge, line 430: You should call computed: true here

SamyPesse: ha yes

gregberge (approved):

PR #3140, opened 2025-04-11
Introduce slim documents

gregberge:
We introduce slim documents, it's a lighter version of the document
computed in GitBook Open to lightweight to document stored in cache and
transferred to client. The idea is to store in this document only what
is needed to display a document in GitBook Open.

Fix RND-6822


emmerich (commented):
My first thought is: why doesn't this live in the API?

gregberge:
@emmerich it's tight to GBO and what we display in it. I think it's easier to maintain here. @SamyPesse didn't want it in the API

SamyPesse:
@emmerich It should not be done in the API, because the choice of what data is important or not to be stored in the cache is relative to what GBO needs to do.

Our API should always respond with "all" the properties needed for different use-cases (some properties are useless on revision pages for GBO, but needed for our users using the API).

GBO can then choose to pick only some properties to select and cache; depending on what is being used.

In the API, we should just get rid of some deprecated properties on `Revision`.

gregberge:
This PR is now outdated, closing it.

PR #3142, opened 2025-04-11
Fix computed revision cache in v1

gregberge:


packages/gitbook/src/lib/api.ts
```diff
@@ -381,10 +394,16 @@ const getAPIContextId = async () => {
...
+    name: 'api.getRevision.v3',
+    tag: (spaceId, revisionId, fetchOptions) =>
+        // Temporary hack to make it work with OpenAPI on v1
+        fetchOptions.tags?.[0] ??
+        getCacheTag({
+            tag: 'revision',
+            space: spaceId,
+            revision: revisionId,
+        }),
+    tagImmutable: false,
```
SamyPesse: Should we create 2 cache function to avoid churning all existing cache and keeping `tagImmutable`?

I'm slightly worried about deploying such a changes and getting 100k requests to our API at once.

Basically I think `getRevision` should stay the same cache parameters and we pass `computed: false`, and then we have the other function for the second case?

packages/gitbook/src/lib/api.ts
```diff
@@ -414,10 +439,16 @@ export const getRevision = cache({
  * Get all the pages in a revision of a space.
  */
 export const getRevisionPages = cache({
-    name: 'api.getRevisionPages.v4',
-    tag: (spaceId, revisionId) =>
-        getCacheTag({ tag: 'revision', space: spaceId, revision: revisionId }),
-    tagImmutable: true,
+    name: 'api.getRevisionPages.v5',
```
SamyPesse, line 442: Same thing about cache churning

SamyPesse (commented):
Looks good, but we need to find a way to preserve all existing cache (99.999% of the site have no openapi).

packages/gitbook/src/lib/api.ts
gregberge: I added them, but in v1 we will switch to the other function, so all v1 will be refetched because if the result is computed we need the tag (currently we don't have them).

packages/gitbook/src/lib/api.ts
gregberge: Same here.

packages/gitbook/src/lib/api.ts
```diff
@@ -440,10 +476,15 @@ export const getRevisionPages = cache({
 
         return cacheResponse(response, {
             ...(fetchOptions.metadata ? cacheTtl_7days : cacheTtl_1day),
-            data: response.data.pages,
+            data: { ...response.data, tags: getResponseCacheTags(response) },
         });
     },
-    getKeyArgs: (args) => [args[0], args[1]],
+    getKeyArgs: ([spaceId, revisionId, fetchOptions]) => {
+        if (fetchOptions.computed === false) {
```
SamyPesse, line 483: You don't seem to pass `computed: fetchOptions.computed` in the actual API call

packages/gitbook/src/lib/api.ts
```diff
@@ -405,18 +424,35 @@ export const getRevision = cache({
...
+        return cacheResponse(response, {
+            ...(fetchOptions.metadata ? cacheTtl_7days : cacheTtl_1day),
+            data: {
+                ...response.data,
+                tags: getResponseCacheTags(response),
+            },
+        });
+    },
+    getKeyArgs: ([spaceId, revisionId, fetchOptions]) => {
+        if (fetchOptions.computed === false) {
```
SamyPesse, line 436: Same thing, where do you use this property?

SamyPesse (commented):
Overall looks good, but we'll have to monitor very closely after deployment, so I'm not approving yet as I'm not on the laptop.

Also it looks like the code is incomplete

packages/gitbook/src/lib/api.ts
gregberge: Same

packages/gitbook/src/lib/api.ts
```diff
@@ -370,6 +371,18 @@ interface GetRevisionOptions {
...
+
+    /**
+     * Whether to fetch the revision as a computed revision.
+     * @default true
+     */
+    computed?: boolean;
+
+    /**
+     * Additional tags to add to the cache entry.
+     * It's only used for v1, once in v2 we can get rid of it.
```
SamyPesse, line 383: This comment is not relaly relevant, this code is only v1 anyway

packages/gitbook/src/lib/api.ts
SamyPesse, line 436: I think it's opposite; you want to use `[spaceId, revisionId]` when `computed = false`, because that's the first call we do; and we want to re-use existing existing cache.

Otherwise here, the call with `computed: false`, will try a new cache key, and bust the cache for all spaces.

packages/gitbook/src/lib/api.ts
SamyPesse, line 483: Same thing about the cache key

packages/gitbook/src/lib/v1.ts
```diff
@@ -134,9 +135,29 @@ async function getDataFetcherV1(): Promise<GitBookDataFetcher> {
 
         getRevision(params) {
             return wrapDataFetcherError(async () => {
-                return getRevision(params.spaceId, params.revisionId, {
+                const { tags, ...revision } = await getRevision(params.spaceId, params.revisionId, {
```
SamyPesse, line 138: It confirms by other comments, you are starting with `computed: false`, so we need to make sure the cache key stays the same as before in this case.

SamyPesse (commented):
At this stage, it'll still bust the cache for all spaces

gregberge:
> At this stage, it'll still bust the cache for all spaces

@SamyPesse it's fixed.

packages/gitbook/src/lib/api.ts
```diff
@@ -440,10 +475,15 @@ export const getRevisionPages = cache({
 
         return cacheResponse(response, {
             ...(fetchOptions.metadata ? cacheTtl_7days : cacheTtl_1day),
-            data: response.data.pages,
+            data: { ...response.data, tags: getResponseCacheTags(response) },
```
SamyPesse, line 478: Isn't it an array usually ?

gregberge:
As I said @SamyPesse I don't think it's a good idea to do that.

PR #3164, opened 2025-04-15
Fix dropdown overlay not staying fixed

zenoachtig:
This is meant to be a quick fix to solve a bug. We should probably move to a more properly supported Dropdown element like react-aria or Radix. But this works for now.

# Before
https://github.com/user-attachments/assets/2405f21b-dabc-457b-8463-5d2bb01aecb2

# After

https://github.com/user-attachments/assets/f624f421-7961-4314-9f2c-052e910f7d8b

### DIFF
--- a/packages/gitbook/src/components/Header/Dropdown.tsx
+++ b/packages/gitbook/src/components/Header/Dropdown.tsx
@@ -10,18 +10,22 @@ export type DropdownButtonProps<E extends HTMLElement = HTMLElement> = Omit<
     'ref'
 >;
 
-/**
- * Button with a dropdown.
- */
-export function Dropdown<E extends HTMLElement>(props: {
+export type DropdownProps<E extends HTMLElement = HTMLElement> = {
     /** Content of the button */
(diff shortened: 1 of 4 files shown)


packages/gitbook/src/components/Header/Dropdown.tsx
```diff
@@ -10,18 +10,22 @@ export type DropdownButtonProps<E extends HTMLElement = HTMLElement> = Omit<
...
     /** Content of the button */
     button: (buttonProps: DropdownButtonProps<E>) => React.ReactNode;
     /** Content of the dropdown */
     children: React.ReactNode;
     /** Custom styles */
     className?: ClassValue;
-}) {
-    const { button, children, className } = props;
+    /** Whether to use fixed positioning (for fixed navbars) */
+    useFixedPosition?: boolean;
```
taranvohra, line 21: I am not sure to understand why we need a special prop for this. The default behaviour of the dropdown is it's fixed and anchored to it's parent no matter what the situation is. Exposing a prop like this leaks implementation details that is not very clear for the caller of this API

PR #3173, opened 2025-04-18
Support for reusable content across spaces.

emmerich:
This PR adds support for reusable content across spaces. It relies on the API token passed from GitBook to resolve the reference.


taranvohra (approved):

packages/gitbook/src/lib/references.tsx
```diff
@@ -41,8 +41,8 @@ export interface ResolvedContentRef {
     file?: RevisionFile;
     /** Page document resolved from the content ref */
     page?: RevisionPageDocument;
-    /** Resolved reusable content, if the ref points to reusable content on a revision. */
-    reusableContent?: RevisionReusableContent;
+    /** Resolved reusable content, if the ref points to reusable content on a revision. Also contains the space and revision used for resolution. */
+    reusableContent?: RevisionReusableContent & { space: string; revision: string };
```
SamyPesse, line 45: I'd recommend against extending interface and instead using 3 propa in an interface

SamyPesse (approved):

PR #3175, opened 2025-04-22
Have all nodes within a text-cells adhere to alignments

viktorrenkema:
Fixes issue raised by aimlapi.com, text cells can receive additional markup including captions. We weren't accounting for those with the alignments, so now we are.

~closes https://linear.app/gitbook-x/issue/RND-6658/annotations-in-table-cells-with-multi-line-text-makes-it-center

Before
CleanShot 2025-04-22 at 10 12 56@2x

After
CleanShot 2025-04-22 at 10 13 06@2x

### DIFF
--- a/packages/gitbook/src/components/DocumentView/Table/RecordColumnValue.tsx
+++ b/packages/gitbook/src/components/DocumentView/Table/RecordColumnValue.tsx
@@ -115,7 +115,7 @@ export async function RecordColumnValue<Tag extends React.ElementType = 'div'>(
                 return <Tag className={tcls(['w-full', verticalAlignment])}>{''}</Tag>;
             }
 
-            const alignment = getColumnAlignment(definition);
+            const horizontalAlignment = `[&_*]:${getColumnAlignment(definition)} ${getColumnAlignment(definition)}`;
 
             return (
                 <Blocks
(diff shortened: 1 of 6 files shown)


packages/gitbook/src/components/DocumentView/Table/RecordColumnValue.tsx
```diff
@@ -130,8 +130,9 @@ export async function RecordColumnValue<Tag extends React.ElementType = 'div'>(
                         'lg:space-y-3',
                         'leading-normal',
                         verticalAlignment,
-                        alignment === 'right' ? 'text-right' : null,
-                        alignment === 'center' ? 'text-center' : null,
+                        alignment === 'right' ? '[&_*]:text-right' : null,
+                        alignment === 'center' ? '[&_*]:text-center' : null,
+                        alignment === 'left' ? '[&_*]:text-left' : null,
```
gregberge, line 135:
```suggestion
                        alignment === 'right' ? '*:text-right' : null,
                        alignment === 'center' ? '*:text-center' : null,
                        alignment === 'left' ? '*:text-left' : null,
```

I think it's equivalent, also as I said to @clairechabas it would be better to use an extensive switch to statically check that if we add an alignment we don't forget that place in the code.

gregberge (approved):

packages/gitbook/src/components/DocumentView/Table/RecordColumnValue.tsx
viktorrenkema: Like this: https://github.com/GitbookIO/gitbook/pull/3175/commits/bb8b4aa4487eeb168eca7e4fa33149e2ae8533cb?

Also, I checked but *: does not impact the child node while `[&_*]` seems to be better supported by tailwind.

PR #3187, opened 2025-04-24
Update docs.gitbook.com to gitbook.com/docs

addisonschultz:
This PR updates the mention of docs.gitbook.com to gitbook.com/docs in preparation for when we switch over to using a proxied domain.

Would love a check on this - I’m not sure if this also affects any tests that rely on the old subdomain (docs.gitbook.com)

Please leave as a draft until we switch over to the subpath domain.


.github/CONTRIBUTING.md
```diff
@@ -63,9 +63,9 @@ To start your local version of GitBook, run the command `bun dev`.
 
 When running the development server, published GitBook sites can be rendered through your local version at `http://localhost:3000/`.
 
-For example, our published docs can be viewed using the local version by visiting `http://localhost:3000/docs.gitbook.com` after running the development server.
+For example, our published docs can be viewed using the local version by visiting `http://localhost:3000/gitbook.com/docs` after running the development server.
```
taranvohra, line 66: Were you able to test a proxied site locally like this?

addisonschultz: @talagluck We’ll need to test this before we go live

talagluck: @taranvohra I've just tested this and it doesn't work quite right - it redirects to the app, rather than the docs site. I compared http://localhost:3000/gitbook.com/docs with http://localhost:3000/docs.photoroom.com which works as expected.

taranvohra: I've fixed that part

taranvohra (approved):

PR #3191, opened 2025-04-28
Allow /~gitbook/image and /~gitbook/icon despite querystring parameters for SEO in robots.txt

jpreynat:
Because currently these paths aren't allowed via the `robots.txt` file, preventing favicons to appear in search results, and images too.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -9,6 +9,7 @@
     ],
     "tailwindCSS.classAttributes": ["class", "className", "style", ".*Style"],
     "prettier.enable": false,
+    "editor.formatOnSave": true,
     "editor.defaultFormatter": "biomejs.biome",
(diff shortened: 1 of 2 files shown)


.vscode/settings.json
```diff
@@ -9,6 +9,7 @@
     ],
     "tailwindCSS.classAttributes": ["class", "className", "style", ".*Style"],
     "prettier.enable": false,
+    "editor.formatOnSave": true,
```
viktorrenkema, line 12: This is intended?

viktorrenkema (approved):

.vscode/settings.json
jpreynat: Yes, because otherwise it doesn't format on save in my VSCode and that's a pain 😓

PR #3192, opened 2025-04-29
Schemas disclosure label causing client error

nolannbiron:


packages/react-openapi/src/schemas/OpenAPISchemaItem.tsx
```diff
@@ -0,0 +1,34 @@
...
+    name: string;
+    schema: OpenAPIV3.SchemaObject;
+    context: OpenAPIClientContext;
+}) {
+    const { schema, context, name } = props;
+
+    return (
+        <OpenAPIDisclosure
+            className="openapi-schemas-disclosure"
+            key={name}
```
gregberge, line 22: Why?

gregberge (approved):

packages/react-openapi/src/schemas/OpenAPISchemaItem.tsx
nolannbiron: This needs to be a client component because tString can return React component

PR #3202, opened 2025-04-30
Adapt site resolution/lookup to pass visitor unsigned claims in addition to token

spastorelli:
This PR aims to add site resolution/lookup to pass visitor unsigned claims in addition to token when calling `getPublishedContentByUrl`. The unsigned claims can originate either from a `gitbook-visitor-public-*` cookie or `visitor.*` query parameters passed to the site url.

Pending https://github.com/GitbookIO/gitbook-x/pull/17625 to be merged.

**TODO:**
 - [x] add method to extracts unsigned claims from public cookie or url param and another one that return an object with both token and unassigned to be passed to `getPublishedContentByUrl` call
 - [x] add method to site resolution lookup module that pass the retrieved visitor payload (token + unsigned claims) and call new `resolvePublishedContentByUrl` endpoint.
 - [x] update site resolution in v2 code to use the new `resolvePublishedContentByUrl` (only for select sites for initially to allow us test)

### DIFF
--- a/packages/gitbook-v2/src/lib/data/lookup.ts
+++ b/packages/gitbook-v2/src/lib/data/lookup.ts
@@ -11,25 +11,14 @@ interface LookupPublishedContentByUrlInput {
     url: string;
     redirectOnError: boolean;
     apiToken: string | null;
+    visitorPayload: SiteVisitorPayload;
 }
-
-type FetchLookupAPIResponse =
-    | {
-          data?: undefined;
-          error: Error;
-      }
-    | {
-          data: Awaited<ReturnType<GitBookAPI['urls']['resolvePublishedContentByUrl']>>;
-          error?: undefined;
-      };
 
 /**
  * Lookup a content by its URL using the GitBook resolvePublishedContentByUrl API endpoint.
  * To optimize caching, we try multiple lookup alternatives and return the first one that matches.
  */
-export async function resolvePublishedContentByUrl(
-    input: LookupPublishedContentByUrlInput & { visitorPayload: SiteVisitorPayload }
-) {
+export async function resolvePublishedContentByUrl(input: LookupPublishedContentByUrlInput) {
     return lookupPublishedContentByUrl({
         url: input.url,
         fetchLookupAPIResult: ({ url, signal }) => {
@@ -62,11 +51,7 @@ export async function resolvePublishedContentByUrl(
  * @deprecated use resolvePublishedContentByUrl.
  *
  */
-export async function getPublishedContentByURL(
-    input: LookupPublishedContentByUrlInput & {
-        visitorAuthToken: string | null;
-    }
-) {
+export async function getPublishedContentByURL(input: LookupPublishedContentByUrlInput) {
     return lookupPublishedContentByUrl({
         url: input.url,
         fetchLookupAPIResult: ({ url, signal }) => {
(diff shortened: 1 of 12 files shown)


packages/gitbook-v2/src/lib/data/lookup.ts
```diff
@@ -1,55 +1,115 @@
...
+          data: Awaited<ReturnType<GitBookAPI['urls']['resolvePublishedContentByUrl']>>;
+          error?: undefined;
+      };
+
+/**
+ * Lookup a content by its URL using the GitBook resolvePublishedContentByUrl API endpoint.
+ * To optimize caching, we try multiple lookup alternatives and return the first one that matches.
+ */
+export async function resolvePublishedContentByUrl(
+    input: LookupPublishedContentByUrlInput & { visitorPayload: SiteVisitorPayload }
```
SamyPesse, line 31: Why is `visitorPayload` not part of `LookupPublishedContentByUrlInput`?

packages/gitbook/src/lib/visitors.ts
```diff
@@ -30,6 +31,25 @@ type VisitorAuthCookieValue = {
...
+    | undefined
+    | { [key: string]: ClaimPrimitive }
+    | ClaimPrimitive[];
+
+/**
+ * The result of a visitor info lookup that can include:
+ *   - a visitor token (JWT)
+ *   - a record of visitor public/unsigned claims (JSON object)
+ */
+export type VisitorPayloadLookup = {
```
SamyPesse, line 48: Instead of creating another new type, can we use `SiteVisitorPayload`?

packages/gitbook/src/lib/visitors.ts
```diff
@@ -82,6 +121,106 @@ export function getVisitorToken({
     }
 }
 
+/**
+ * Get the visitor unsigned/public claims for the request. They can either be in `visitor.` query
```
SamyPesse, line 125: Should it be `gitbook.` instead of `visitor.`?

packages/gitbook-v2/src/lib/data/lookup.ts
```diff
@@ -1,55 +1,115 @@
...
 import type { DataFetcherResponse } from './types';
 import { getURLLookupAlternatives, stripURLSearch } from './urls';
 
+interface LookupPublishedContentByUrlInput {
+    url: string;
+    redirectOnError: boolean;
+    apiToken: string | null;
+}
+
+type FetchLookupAPIResponse =
```
SamyPesse, line 16: Just use `DataFetcherResponse` instead of duplicating types

packages/gitbook-v2/src/lib/data/lookup.ts
```diff
@@ -1,55 +1,115 @@
...
+        fetchLookupAPIResult: ({ url, signal }) => {
+            const api = apiClient({ apiToken: input.apiToken });
+            return trace(
+                {
+                    operation: 'resolvePublishedContentByUrl',
+                    name: url,
+                },
+                () =>
+                    tryCatch(
+                        api.urls.resolvePublishedContentByUrl(
```
SamyPesse, line 44: Have we tested that the CF caching is working as expected on these POST?

SamyPesse (approved):

packages/gitbook/src/lib/visitors.ts
taranvohra: Not really because the lookup here also returns info about where the token was retrieved from (params, cookies etc)

packages/gitbook/src/lib/visitors.ts
taranvohra: We should reserve `gitbook.` for something else. Let's keep `visitor.` for clarity

packages/gitbook-v2/src/lib/data/lookup.ts
taranvohra: Yes I had ran some tests and the cache was busted properly

PR #3204, opened 2025-05-01
Update opennext and leverage "use cache" on CF

SamyPesse:


conico974:
I think the only thing missing here would be to wait for this https://github.com/opennextjs/opennextjs-cloudflare/pull/666 to be merged and released so that we could use npm release instead of `pkg.pr.new`

packages/gitbook-v2/package.json
```diff
@@ -5,6 +5,7 @@
     "dependencies": {
         "@gitbook/api": "^0.115.0",
         "@gitbook/cache-tags": "workspace:*",
+        "@opennextjs/cloudflare": "https://pkg.pr.new/@opennextjs/cloudflare@666",
```
SamyPesse, line 8: Same questionhttps://github.com/GitbookIO/gitbook-x/pull/17784

SamyPesse (commented):
I think we can merge and deploy it already to test with a bit of traffic

packages/gitbook-v2/package.json
conico974: Right, i didn't realize it was a dev deps, but you're using `getCloudflareContext`, shouldn't this be a normal deps ?

SamyPesse: True, then let's remove it from `devDependencies`. It's just semantic at the end

PR #3205, opened 2025-05-01
Add reusable-content cache tag.

emmerich:
I want to change the caching strategy for reusable content to be based on the reusable content itself. Right now it's based on the containing space/revision. I'm adding this cache tag so I can build things internally to support it.


emmerich:
After some thought, the space tag will be fine.

PR #3206, opened 2025-05-01
Upgrade tailwind v4

zenoachtig:


gregberge:
There are lot of changes in Argos, we should aim for a ISO migration here.

zenoachtig:
> There are lot of changes in Argos, we should aim for a ISO migration here.

I've now gotten it as close to 1:1 as I could. I've adjusted some margins to be close to what it was before. I've gone through Argos extensively, the detected changes are all just small shifts in margin because I had to re-do the page header margins because some of the `space-*` utilities changed how they worked.

PR #3207, opened 2025-05-02
Missing top-level required OpenAPI alternatives

nolannbiron:


gregberge (commented):
I don't think it's good. A required parent does not mean children are required. Also we have ancestors to know that.

nolannbiron:
@gregberge You can require children in parent like this : 
```
xxSchema:
      required:
        - name
        - type
      allOf:
        - $ref: '#/components/schemas/...'
        - ...
```

I'll try to get this from ancestors 👍🏻

gregberge:
Hmm ok so it's valid. Yeah let's use the latest ancestor and we are good to go.

packages/react-openapi/src/OpenAPISchema.tsx
```diff
@@ -492,7 +493,8 @@ export function getSchemaAlternatives(
  */
 function mergeAlternatives(
     alternativeType: AlternativeType,
-    schemasOrRefs: OpenAPIV3.SchemaObject[]
+    schemasOrRefs: OpenAPIV3.SchemaObject[],
+    parentRequired?: string[]
```
gregberge, line 497: Use ancestors instead.

gregberge (approved):

PR #3212, opened 2025-05-06
Fix resolution of links in reusable content

SamyPesse:


packages/gitbook/src/components/DocumentView/ReusableContent.tsx
```diff
@@ -46,14 +46,15 @@ export async function ReusableContent(props: BlockProps<DocumentBlockReusableCon
     // Create a new context for reusable content block, including
     // the data fetcher with the token from the block meta and the correct
     // space and revision pointers.
-    const reusableContentContext: GitBookSpaceContext = {
-        ...context.contentContext,
-        dataFetcher,
-        space: resolved.reusableContent.space,
-        revisionId: resolved.reusableContent.revision,
-        pages: [],
```
emmerich, line 54: It'll fix links in same-space reusable content but I think we have an issue with relative links in cross-space. Imagine:

```
SpaceA
  PageA
  RucA
    <link to PageA>

SpaceB
  PageB
    <ref to RucA>
```

When you render SpaceB, it'll try to resolve the link to PageA but be unable to as it's relative. I think this is a different issue though.

emmerich (approved):

PR #3220, opened 2025-05-07
Fix sitemap URLs

gregberge:
- Fix URLs in sitemap.xml

### DIFF
--- a/packages/gitbook-v2/src/lib/links.ts
+++ b/packages/gitbook-v2/src/lib/links.ts
@@ -93,6 +93,11 @@ export function createLinker(
         },
 
         toAbsoluteURL(absolutePath: string): string {
+            // If the path is already a full URL, we return it as is.
+            if (URL.canParse(absolutePath)) {
+                return absolutePath;
+            }
+
             if (!servedOn.host) {
                 return absolutePath;
             }
--- a/.changeset/healthy-parrots-call.md
+++ b/.changeset/healthy-parrots-call.md
@@ -1,0 +1,6 @@
+---
+"gitbook-v2": patch
+"gitbook": patch
+---
+
+Fix URL in sitemap


packages/gitbook-v2/src/lib/links.ts
```diff
@@ -93,6 +93,10 @@ export function createLinker(
         },
 
         toAbsoluteURL(absolutePath: string): string {
+            if (absolutePath.startsWith('http://') || absolutePath.startsWith('https://')) {
```
SamyPesse, line 96: Let's use URL.canParse instead

packages/gitbook/src/lib/v1.ts
```diff
@@ -48,16 +48,21 @@ export async function getV1BaseContext(): Promise<GitBookBaseContext> {
...
+        protocol,
         host,
         spaceBasePath: basePath,
         siteBasePath: siteBasePath,
     });
 
     // On V1, we use hard-navigation between different spaces because of layout issues
-    linker.toLinkForContent = (url) => {
-        return url;
+    linker.toLinkForContent = (rawURL) => {
```
SamyPesse, line 61: Why do we change this?

SamyPesse (approved):

packages/gitbook/src/lib/v1.ts
gregberge: I wanted to fix the fact that in proxy mode URL are wrong, but it seems to break things.

taranvohra: Proxy mode is only supported in v2, it was on purpose

gregberge: I am not talking about this proxy mode. In dev the URL are wrong for sections, sitemap and other places. So I was trying to fix that.

gregberge: I removed it from this PR.

PR #3231, opened 2025-05-13
Persist visitor params values in public visitor session cookie

spastorelli:
This PR aims to persist visitor params values in public visitor session cookie.

### DIFF
--- a/packages/gitbook-v2/src/middleware.ts
+++ b/packages/gitbook-v2/src/middleware.ts
@@ -9,6 +9,7 @@ import { removeLeadingSlash, removeTrailingSlash } from '@/lib/paths';
 import {
     type ResponseCookies,
     getPathScopedCookieName,
+    getResponseCookieForVisitorParams,
     getResponseCookiesForVisitorAuth,
     getVisitorPayload,
     normalizeVisitorAuthURL,
@@ -109,7 +110,7 @@ async function serveSiteRoutes(requestURL: URL, request: NextRequest) {
                 url: siteRequestURL.toString(),
                 visitorPayload: {
                     jwtToken: visitorToken?.token ?? undefined,
-                    unsignedClaims,
+                    unsignedClaims: unsignedClaims.all,
                 },
                 // When the visitor auth token is pulled from the cookie, set redirectOnError when calling getPublishedContentByUrl to allow
                 // redirecting when the token is invalid as we could be dealing with stale token stored in the cookie.
@@ -121,8 +122,13 @@ async function serveSiteRoutes(requestURL: URL, request: NextRequest) {
                 apiToken,
             })
         );
-        const cookies: ResponseCookies = [];
 
+        //
+        // Retrieve the response cookie for visitor params that were passed to the URL to persist them in a session cookie.
+        //
+        const visitorParamsCookie = getResponseCookieForVisitorParams(unsignedClaims);
+
+        const cookies: ResponseCookies = visitorParamsCookie ? [visitorParamsCookie] : [];
         //
         // Handle redirects
         //
--- a/packages/gitbook-v2/src/middleware.ts
+++ b/packages/gitbook-v2/src/middleware.ts
@@ -9,9 +9,8 @@ import { removeLeadingSlash, removeTrailingSlash } from '@/lib/paths';
 import {
     type ResponseCookies,
     getPathScopedCookieName,
-    getResponseCookieForVisitorParams,
     getResponseCookiesForVisitorAuth,
-    getVisitorPayload,
+    getVisitorData,
     normalizeVisitorAuthURL,
 } from '@/lib/visitors';
 import { serveResizedImage } from '@/routes/image';
@@ -96,7 +95,7 @@ async function serveSiteRoutes(requestURL: URL, request: NextRequest) {
     //
     // Detect and extract the visitor authentication token from the request
     //
-    const { visitorToken, unsignedClaims } = getVisitorPayload({
+    const { visitorToken, unsignedClaims, visitorParamsCookie } = getVisitorData({
         cookies: request.cookies.getAll(),
         url: siteRequestURL,
     });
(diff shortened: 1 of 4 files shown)


packages/gitbook/src/lib/visitors.ts
```diff
@@ -221,6 +228,28 @@ function parseVisitorQueryParamValue(value: string): ClaimPrimitive {
...
+ */
+export function getResponseCookieForVisitorParams(
+    visitorUnsignedClaims: VisitorPayloadLookup['unsignedClaims']
+): ResponseCookie | undefined {
+    if (Object.keys(visitorUnsignedClaims.fromVisitorParams).length === 0) {
+        return undefined;
+    }
+
+    return {
+        name: VISITOR_UNSIGNED_CLAIMS_PREFIX,
```
spastorelli, line 242: Using the `gitbook-visitor-public` but happy to change to store the params in a separate/dedicated one.

packages/gitbook-v2/src/middleware.ts
```diff
@@ -121,8 +122,13 @@ async function serveSiteRoutes(requestURL: URL, request: NextRequest) {
                 apiToken,
             })
         );
-        const cookies: ResponseCookies = [];
 
+        //
+        // Retrieve the response cookie for visitor params that were passed to the URL to persist them in a session cookie.
+        //
+        const visitorParamsCookie = getResponseCookieForVisitorParams(unsignedClaims);
+
```
taranvohra, line 130: Why not return it from `getVisitorPayload`? We should avoid too many visitor methods

taranvohra:
```js
const {visitorToken, visitorPuvlicCookie, visitorUnsignedClaims} = getVisitorPayload();
```

packages/gitbook/src/lib/visitors.ts
```diff
@@ -47,7 +48,10 @@ type ClaimPrimitive =
  */
 export type VisitorPayloadLookup = {
     visitorToken: VisitorTokenLookup;
-    unsignedClaims: Record<string, ClaimPrimitive>;
+    unsignedClaims: {
+        all: Record<string, ClaimPrimitive>;
+        fromVisitorParams: Record<string, ClaimPrimitive>;
+    };
```
taranvohra, line 54: I feel we should only be returning `visitorClaims` which are all the `unsigned` claims. The distinction should not be needed for the caller. 

And then like I said above it should also return `visitorPublicCookie` (if any) that contains claims from the params etc

packages/gitbook/src/lib/visitors.ts
```diff
@@ -221,6 +228,28 @@ function parseVisitorQueryParamValue(value: string): ClaimPrimitive {
...
+    }
+
+    return {
+        name: VISITOR_UNSIGNED_CLAIMS_PREFIX,
+        value: JSON.stringify(visitorUnsignedClaims.fromVisitorParams),
+        options: {
+            httpOnly: true,
+            sameSite: process.env.NODE_ENV === 'production' ? 'none' : undefined,
+            secure: process.env.NODE_ENV === 'production',
+            maxAge: VISITOR_PARAMS_COOKIE_MAX_AGE,
```
taranvohra, line 248: Should it not be a session cookie?

packages/gitbook-v2/src/middleware.ts
spastorelli: Seperation of concerns. It's also following the same pattern that we use for the visitor auth token. One to perform the lookup and another one to retrieve the cookie response. Not too sure how is it problematic to have 2 seperate methods?

spastorelli: it's also odd imo to have a function called `getVisitorPayload` return a cookie.

packages/gitbook/src/lib/visitors.ts
spastorelli: Yes good point not sure what why I did this 🤦

packages/gitbook-v2/src/middleware.ts
taranvohra: Visitor auth is a bit different because setting the cookie for it is not very straight forward. You actually need the `basePath` that comes after the resolution API call hence separate function for it.

My intention with the suggestion was that `getVisitorPayload` (or maybe `getVisitorData`?) should be an API that should output all relevant info related to a visitor for a particular request.

spastorelli: Ok I see your point on the API. Changing it.

packages/gitbook/src/lib/visitors.ts
```diff
@@ -221,6 +231,27 @@ function parseVisitorQueryParamValue(value: string): ClaimPrimitive {
...
+): ResponseCookie | undefined {
+    if (Object.keys(visitorParamsClaims).length === 0) {
+        return undefined;
+    }
+
+    return {
+        name: VISITOR_UNSIGNED_CLAIMS_PREFIX,
+        value: JSON.stringify(visitorParamsClaims),
+        options: {
+            httpOnly: true,
```
taranvohra, line 248: Why `httpOnly`? It means only the server can read it and not client side scripts

packages/gitbook/src/lib/visitors.ts
```diff
@@ -44,10 +44,12 @@ type ClaimPrimitive =
  * The result of a visitor info lookup that can include:
  *   - a visitor token (JWT)
  *   - a record of visitor public/unsigned claims (JSON object)
+ *   - a session cookie response to persist any visitor query params across navigations.
  */
 export type VisitorPayloadLookup = {
```
taranvohra, line 49:
```suggestion
export type VisitorDataLookup = {
```

packages/gitbook/src/lib/visitors.ts
```diff
@@ -128,9 +135,10 @@ export function getVisitorToken({
 export function getVisitorUnsignedClaims(args: {
     cookies: RequestCookies;
     url: URL | NextRequest['nextUrl'];
-}): Record<string, ClaimPrimitive> {
+}): { all: Record<string, ClaimPrimitive>; fromVisitorParams: Record<string, ClaimPrimitive> } {
```
taranvohra, line 138:
```suggestion
}): { all: Record<string, ClaimPrimitive>; params: Record<string, ClaimPrimitive> } {
```
It's better to add some comments to these properties to signify their origin

taranvohra (approved):

PR #3233, opened 2025-05-13
Dynamic TOC height

zenoachtig:
This PR adds a client-side available height calculation for the TableOfContents component. This solves two issues:
- When the announcement banner is enabled, the TOC's static height is too high to fit in the screen initially. When scrolling down the banner disappears out of frame, which means the TOC needs to grow to take the available height (causing issues like RND-7090)
- When reaching the footer, the TOC bumps up against it and moves visually up the screen, with its top part cut off. We should subtract the footer's visible height from the TOC height.

Important to note: to prevent layout shift the TOC still has a number of `top-*` and `h-*` classes added to it, to set its initial positioning as close as possible on the server. Once the client loads we override these classes with the dynamic height through CSS variables.

# Before
https://github.com/user-attachments/assets/eed57279-331a-4b33-a56a-1315b432ce2f

# After
https://github.com/user-attachments/assets/9c2b0a7a-95cd-4bda-91ff-d41d661b1b9e

Fixes RND-7090


packages/gitbook/src/components/TableOfContents/TableOfContentsScript.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+            // Set sticky top position based on header
+            const headerHeight = header?.offsetHeight ?? 0;
+            root.style.setProperty('--toc-top-offset', `${headerHeight}px`);
+
+            // Start with full viewport height minus header
+            let height = window.innerHeight - headerHeight;
+
+            // Subtract visible banner (if any)
+            if (banner && banner?.computedStyleMap().get('display') !== 'none') {
```
conico974, line 27: `computedStyleMap` doesn't work on Firefox https://developer.mozilla.org/en-US/docs/Web/API/Element/computedStyleMap

zenoachtig: Had no idea, thank you! Changed in favour of the classic `window.getComputedStyle()`

conico974: Me neither to be honest :smile: , i just tested on Firefox and it crashed right away

PR #3239, opened 2025-05-16
Custom Incremental Cache

conico974:
This PR add a custom incremental cache instead of the one from cloudflare.

It does not use the `BUILD_ID` as part of the key for the data cache. 
It also add tracing for the different methods of the incremental cache. (This is broken right now as esbuild replace `NODE_ENV` with development during compilation of the config file for the edge runtime, fix needs to be done in aws)

### DIFF
--- a/packages/gitbook-v2/openNext/incrementalCache.ts
+++ b/packages/gitbook-v2/openNext/incrementalCache.ts
@@ -9,8 +9,6 @@ import type {
 import { getCloudflareContext } from '@opennextjs/cloudflare';
 
 export const BINDING_NAME = 'NEXT_INC_CACHE_R2_BUCKET';
-export const PREFIX_ENV_NAME = 'NEXT_INC_CACHE_R2_PREFIX';
-export const FALLBACK_BUILD_ID = 'no-build-id';
 export const DEFAULT_PREFIX = 'incremental-cache';
 
 export type KeyOptions = {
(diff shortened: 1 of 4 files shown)


packages/gitbook-v2/openNext/incrementalCache.ts
```diff
@@ -0,0 +1,180 @@
...
+    name = 'GitbookIncrementalCache';
+
+    protected localCache: Cache | undefined;
+
+    async get<CacheType extends CacheEntryType = 'cache'>(
+        key: string,
+        cacheType?: CacheType
+    ): Promise<WithLastModified<CacheValue<CacheType>> | null> {
+        const cacheKey = this.getR2Key(key, cacheType);
+        return trace(
```
conico974, line 29: We could either wait for the next release in cloudflare to fix tracing not being recorded, or we could release like that, it should not have any impact besides logs not outputted to the console

packages/gitbook-v2/openNext/incrementalCache.ts
SamyPesse, line 1: Why not have this in `src` or even `src/lib` ? especially as you use `@/lib` import in it

Also fine to keep it here if we really want to isolate

packages/gitbook-v2/openNext/incrementalCache.ts
```diff
@@ -0,0 +1,180 @@
...
+import { getCloudflareContext } from '@opennextjs/cloudflare';
+
+export const BINDING_NAME = 'NEXT_INC_CACHE_R2_BUCKET';
+export const DEFAULT_PREFIX = 'incremental-cache';
+
+export type KeyOptions = {
+    cacheType?: CacheEntryType;
+};
+
+class GitbookIncrementalCache implements IncrementalCache {
```
SamyPesse, line 19: It looks like you had to re-implement the whole R2 cache?

Any chance we could extend it instead?
Or at least we should comment the differences

SamyPesse (approved):
Looks great!

packages/gitbook-v2/openNext/incrementalCache.ts
conico974: Yeah, that was the idea to isolate OpenNext stuff in its own thing. If you prefer to have it in `src` i'm fine with it as well

packages/gitbook-v2/openNext/incrementalCache.ts
conico974: Yeah i had this idea as well, while we could do that (but it would require some change in cloudflare, the class are not exported from there), i think it just make sense that we have our own implementation here where we could tweak things exactly as we need in the future.
This one is actually a mix of the R2 cache and the regional cache

PR #3240, opened 2025-05-17
Pass spec validation errors through OpenAPIParseError

nolannbiron:


gregberge (commented):
Can you give more context about it? What do we want to do here?

nolannbiron:
@gregberge yes, I want to get scalar validation errors to display them in GBX. If a spec is invalid, we should tell the user what's invalid (feedback from Bird) so they don't have to find where the problem is in their spec. Currently in GBX we just say "Your spec is invalid" and that's it

PR #3249, opened 2025-05-20
Add temporary logging for Cloudflare debugging

conico974:
OG Image generation seems to fail for bot once deployed with this error `Error: Invalid content type: text/html; charset=utf-8`

Same URL, when accessed manually works. Haven't been able to reproduce.

This just add a bunch of logs to help narrow what could go wrong here

### DIFF
--- a/packages/gitbook/src/routes/ogimage.tsx
+++ b/packages/gitbook/src/routes/ogimage.tsx
@@ -312,12 +312,24 @@ async function loadCustomFont(input: { url: string; weight: 400 | 700 }) {
 }
 
 /**
+ * Temporary function to log some data on Cloudflare.
+ * TODO: remove this when we found the issue
+ */
+function logOnCloudflareOnly(message: string) {
+    if (process.env.DEBUG_CLOUDFLARE === 'true') {
+        // biome-ignore lint/suspicious/noConsole: <explanation>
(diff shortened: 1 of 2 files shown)


packages/gitbook/src/routes/ogimage.tsx
```diff
@@ -311,13 +311,25 @@ async function loadCustomFont(input: { url: string; weight: 400 | 700 }) {
...
+}
+
 /**
  * Fetch a resource from the function itself.
  * To avoid error with worker to worker requests in the same zone, we use the `WORKER_SELF_REFERENCE` binding.
  */
 async function fetchSelf(url: string) {
     const cloudflare = getCloudflareContext();
     if (cloudflare?.env.WORKER_SELF_REFERENCE) {
+        logOnCloudflareOnly(`Fetching self: ${url}`);
```
SamyPesse, line 332: You could log in all cases, this code already only runs on Cloudflare.

SamyPesse (approved):

packages/gitbook/src/routes/ogimage.tsx
conico974: Yeah but it was more to not forgot to remove it later. This way once i remove the function, i'll remove that as well

PR #3253, opened 2025-05-21
Fix OGImage icon fetch on cloudflare

conico974:
When a request for OG Image arrives for a custom domain, the requested favicon is requested from the same domain (i.e. `https://thedomain.com/~gitbook/icon`)
In Cloudflare, we were using a service binding, but by doing so, this bypass setting the `x-gitbook-url` header and thus the request will end up not being able to match `getSiteUrlFromRequest` https://github.com/GitbookIO/gitbook/blob/ba0094a862e28b83728cc05c2503c4b652cf93ed/packages/gitbook-v2/src/middleware.ts#L382-L391
We should be fine without the service binding since we set `global_fetch_strictly_public` compatibility flag

The icon will return a 404 and the OG Image Generation will fail with a 500

Should fix RND-7140

### DIFF
--- a/packages/gitbook/src/routes/ogimage.tsx
+++ b/packages/gitbook/src/routes/ogimage.tsx
@@ -9,7 +9,6 @@ import { getAssetURL } from '@/lib/assets';
 import { filterOutNullable } from '@/lib/typescript';
 import { getCacheTag } from '@gitbook/cache-tags';
 import type { GitBookSiteContext } from '@v2/lib/context';
-import { getCloudflareContext } from '@v2/lib/data/cloudflare';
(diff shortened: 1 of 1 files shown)


SamyPesse (approved):
Makes sense, the alternative was to set the header as well here, but I think it's better to unify between Vercel and Cloudflare on the way we fetch things here

conico974:
> Makes sense, the alternative was to set the header as well here, but I think it's better to unify between Vercel and Cloudflare on the way we fetch things here

I've started doing exactly that at first, but there is not much benefits in doing so (technically it's slightly faster), it just complicate things.

PR #3254, opened 2025-05-22
External middleware in cloudflare

conico974:
The goal of this PR is to deploy the routing/middleware layer in its own worker.
This should make cold start for ISR/SSG route that has been generated before really fast compared to now (On par or even better than in Vercel)
It requires #3265 to be merged before release.

Will close RND-7163

## What's changed

- The routing/middleware are in a separate worker. (ISR will be served from here)
- The server and middleware worker are "linked" together. In preview, we use the preview url, in staging and prod, we use the `Cloudflare-Workers-Version-Overrides` headers (this requires the 2 versions to be uploaded first)
- Deployment are now a multi step process to ensure that the middleware will always access the correct server
- It doesn't use the binding anymore for the API, in some case it can cause a `Response closed due to connection limit` because the limit are shared between the workers and all its binding, especially for big pages. (This should close RND-7189)
- This will also reduce memory consumption by the server worker since the middleware is in a different worker (it will help with RND-7141)

One of the downsides of this method is that changes to `routes` in `wrangler.jsonc` won't work automatically. It requires running `wrangler triggers deploy`

Also, we can't really test the Durable Object without deploying (DO don't work with `preview url` )


.github/actions/gradual-deploy-cloudflare/action.yaml
```diff
@@ -0,0 +1,83 @@
...
+      - id: deploy_server
+        name: Deploy server to Cloudflare at 0%
+        uses: cloudflare/wrangler-action@v3.14.0
+        with:
+            apiToken: ${{ inputs.apiToken }}
+            accountId: ${{ inputs.accountId }}
+            workingDirectory: ./
+            wranglerVersion: '4.10.0'
+            environment: ${{ inputs.environment }}
+            command: versions deploy ${{ steps.extract_current_version.outputs.version_id }}@100% ${{ inputs.serverVersionId }}@0% -y --config ./packages/gitbook-v2/openNext/customWorkers/defaultWrangler.jsonc
```
jpreynat, line 54: Argh, too bad there isn't a "--no-traffic-update" flag 😅

jpreynat (approved):
Looking great 👌 

Hopefully at some point the whole orchestration could be better handled at a lower level by opennext itself, but this should definitely be a great improvement in perfs for us

.github/actions/gradual-deploy-cloudflare/action.yaml
conico974: Yeah this would have been very handy. This should get better in the future :crossed_fingers:

conico974:
> Hopefully at some point the whole orchestration could be better handled at a lower level by opennext itself, but this should definitely be a great improvement in perfs for us

Yeah this will get better, but to make it work in a more generic way, it requires a bunch of change, and not only to OpenNext

conico974:
I did realize something while working on another issue. Migrations also won't work and actually make wrangler crash.
I'll include something in another PR for that (it's not really needed here)

SamyPesse (approved):

packages/gitbook-v2/openNext/customWorkers/middleware.js
```diff
@@ -0,0 +1,41 @@
...
+    async fetch(request) {
+        return runWithCloudflareRequestContext(request, this.env, this.ctx, async () => {
+            // - `Request`s are handled by the Next server
+            const reqOrResp = await middlewareHandler(request, this.env, this.ctx);
+            if (reqOrResp instanceof Response) {
+                return reqOrResp;
+            }
+
+            if (this.env.STAGE !== 'preview') {
+                reqOrResp.headers.set(
```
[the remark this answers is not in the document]
conico974: It's to make sure that the middleware sends data to the right server worker while we do the migration https://developers.cloudflare.com/workers/configuration/versions-and-deployments/gradual-deployments/#version-affinity

PR #3256, opened 2025-05-22
Open image zoon viewer on click for any image size

clairechabas:
The image viewer should open on click whatever the image size is. Otherwise when the viewer doesn't open there's no clear reason for the user who clicked why that is which is confusing.


SamyPesse (commented):
Shared my feedback on Slack, I think this is the wrong call

PR #3280, opened 2025-06-04
Add circular corners and depth styles

zenoachtig:


packages/gitbook/src/components/DocumentView/Block.tsx
SamyPesse, line 1: We should revert the changes in this file. I guess it's caused by the column block, the solution is to add a `case 'column':`

SamyPesse (approved):
Make sure all `@gitbook/api` in all package.json points to the latest version

and we should add some tests for the customization options

PR #3282, opened 2025-06-07
Revamp mobile navigation

nolannbiron:


packages/gitbook/src/components/Header/mobile-menu/MobileMenuSheet.tsx
```diff
@@ -0,0 +1,33 @@
+'use client';
+
+import {
+    Sheet,
+    SheetContent,
+    SheetDescription,
+    SheetHeader,
+    SheetTitle,
+} from '@/components/utils/Sheet';
```
SamyPesse: Maybe this should be in primitives instead of utils?

I generally don't like "utils" folder as they mean nothing and everything

packages/gitbook/src/components/utils/Sheet.tsx
SamyPesse, line 1: We should move this primitives and maybe commend a bit on the main component and what it represents

packages/gitbook/src/components/Header/HeaderMobileMenuButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+    const toggleNavigation = () => {
+        setOpen(!open);
+    };
+
+    // biome-ignore lint/correctness/useExhaustiveDependencies: Close the navigation when navigating to a page
+    useEffect(() => {
+        if (!open) return;
+        setOpen(false);
+    }, [pathname]);
```
SamyPesse: It'd be good to do something proper with react here instead of hacking around the deps.

A Ref can be used to store the latest value of open

Or maybe we call it everytime

packages/gitbook/src/components/SpaceLayout/SpaceLayout.tsx
```diff
@@ -11,6 +11,8 @@ import { getSpaceLanguage } from '@/intl/server';
 import { t } from '@/intl/translate';
 import { tcls } from '@/lib/tailwind';
 
+import { MobileMenuSheet } from '@/components/Header/mobile-menu/MobileMenuSheet';
```
SamyPesse: Let's move this out of `Header` if we want to export this way; we should avoid deep imports

```suggestion
import { MobileMenuSheet } from '@/components/MobileMenu';
```

packages/gitbook/src/components/SpaceLayout/SpaceLayout.tsx
```diff
@@ -11,6 +11,8 @@ import { getSpaceLanguage } from '@/intl/server';
...
+import { TOCScrollContent } from '@/components/TableOfContents/TOCScrollContent';
```
SamyPesse: Same thing

packages/gitbook/src/components/SpaceLayout/SpaceLayout.tsx
```diff
@@ -66,6 +68,27 @@ export function SpaceLayout(props: {
             >
                 <Announcement context={context} />
                 <Header withTopHeader={withTopHeader} context={context} />
+                <MobileMenuSheet>
```
SamyPesse: Does it mean EVERY html will contain the TOC twice?

This looks impactful on perfs

SamyPesse (commented):
I think it can negatively impact perfs as it means the TOC will be rendered twice; we have spaces with thousands of pages, which can result in a big HTML chunk just for mobile

nolannbiron:
@SamyPesse Probably not the sexiest approach, but this way we can preserve the initial HTML (desktop TOC) and switch to the Sheet TOC for mobile, so no duplicate

I’m not a big fan of relying on `useIsMobile`, but given that we’re passing a `context` props (not serializable) to a lot of components, it’s tricky to implement a clean client-side solution without breaking things

I still think it's a good improvement and it won't change anything on desktop

packages/gitbook/src/components/Header/HeaderMobileMenuButton.tsx
```diff
@@ -0,0 +1,43 @@
...
+    const { open, setOpen } = useMobileMenuSheet();
+
+    const toggleNavigation = () => {
+        setOpen(!open);
+    };
+
+    // biome-ignore lint/correctness/useExhaustiveDependencies: Close the navigation when navigating to a page
+    useEffect(() => {
+        setOpen(false);
+    }, [pathname]);
```
gregberge: Why do we have this logic in the MobileMenuButton and not elsewhere? A Provider or something like that?

gregberge: You will re-render the button when the pathname change for nothing.

packages/gitbook/src/components/TableOfContents/TableOfContents.tsx
```diff
@@ -1,21 +1,23 @@
...
 }) {
-    const { innerHeader, context, header } = props;
-    const { space, customization, pages } = context;
+    const { header, children } = props;
+    const isMobile = useIsMobile();
+
+    // If the screen is mobile, we use the mobile menu sheet to display the table of contents.
+    if (isMobile) {
+        return <MobileMenuSheet>{children}</MobileMenuSheet>;
+    }
```
gregberge: Why do we need that. I don't understand why we don't keep the same element and we just switch its position to absolute or something like that? In the app, in mobile I use another layout but we could keep the same component like in desktop actually (I think I will refactor it at some point). I think we could do something smarter here.

gregberge (commented):
I tested it, the animation is not great and the backdrop is blinking and disappearing. 

It's still need to be polished before being ready to be merged.

https://github.com/user-attachments/assets/fa45b906-cace-4cb9-ad63-4083d2762f72

packages/gitbook/src/components/MobileMenu/MobileMenuScript.tsx
```diff
@@ -0,0 +1,26 @@
...
+    }, [pathname]);
+
+    useEffect(() => {
+        // If the menu is open, we add a class to the body to prevent scrolling
+        if (open) {
+            document.body.style.overflow = 'hidden';
+        } else {
+            document.body.style.overflow = 'auto';
+        }
+    }, [open]);
```
gregberge: It's a bit more complicated than that:

https://github.com/adobe/react-spectrum/blob/main/packages/%40react-aria/overlays/src/usePreventScroll.ts#L44

Can you use that insteand?

PR #3292, opened 2025-06-09
revert to use withoutConcurrentExecution

conico974:


packages/gitbook-v2/src/lib/data/cloudflare.ts
```diff
@@ -11,3 +12,12 @@ export function getCloudflareContext() {
 
     return getCloudflareContextOpenNext();
 }
+
+export function getCloudflareRequestGlobal() {
+    const context = getCloudflareContext();
+    if (!context) {
+        return cache(() => ({}));
```
SamyPesse: Should it be outside the function call?

Also maybe it doesn't have to be cloudflare specific anymore?

packages/gitbook-v2/src/lib/data/memoize.ts
```diff
@@ -0,0 +1,87 @@
+/**
+ * Wrap a function by preventing concurrent executions of the same function.
+ * With a logic to work per-request in Cloudflare Workers.
+ */
+export function withoutConcurrentExecution<ArgsType extends any[], ReturnType>(
+    getGlobalContext: () => object | null | undefined,
```
SamyPesse: We could remove this option and just use the `cache(() => ({}))` here

packages/gitbook-v2/src/lib/data/memoize.ts
```diff
@@ -0,0 +1,87 @@
...
+    wrapped: (key: string, ...args: ArgsType) => Promise<ReturnType>
+): (cacheKey: string, ...args: ArgsType) => Promise<ReturnType> {
+    const globalPromiseCache = new WeakMap<object, Map<string, Promise<ReturnType>>>();
+
+    return (key: string, ...args: ArgsType) => {
+        const globalContext = getGlobalContext() ?? globalThis;
```
SamyPesse: This can be dangerous, we should make sure we never use `globalThis`

SamyPesse (approved):
Looks good, but when I think about it, we might still need to manage concurrency inside the `use cache`: imagine that the Node.js process receive 10 requests to generate pages in the same space, right now it'll result in 10 API requests to the same space for each of the top level requests.
I think we should manage concurrency at 2 levels:
- Wrapping in the `use cache`: per-request using `cache(() => {})`
- Wrapping the API call in the `use cache`: global in the process

packages/gitbook-v2/src/lib/data/cloudflare.ts
conico974: Yeah totally

PR #3293, opened 2025-06-10
Rework full-width layout, add support for full-width page option

zenoachtig:
- Rename `page-full-width` to `site-full-width`. This class will be added whenever any element on the page is full-width, and the site layout has to adapt to it.
- Individual blocks can be set to full-width and expand, while other blocks remain bound to the default centred container.
- Setting the full-width page option increases the container width from `3xl` to `screen-2xl`.

### DIFF
--- a/packages/gitbook/src/components/DocumentView/Blocks.tsx
+++ b/packages/gitbook/src/components/DocumentView/Blocks.tsx
@@ -70,8 +70,8 @@ export function UnwrappedBlocks<TBlock extends DocumentBlock>(props: UnwrappedBl
                 style={[
                     'mx-auto w-full decoration-primary/6',
                     node.data && 'fullWidth' in node.data && node.data.fullWidth
-                        ? 'max-w-screen-xl'
-                        : 'max-w-3xl',
+                        ? 'max-w-screen-2xl'
+                        : 'page-full-width:ml-0 max-w-3xl',
                     blockStyle,
                 ]}
                 isEstimatedOffscreen={isOffscreen}
--- a/.changeset/long-cameras-protect.md
+++ b/.changeset/long-cameras-protect.md
@@ -1,0 +1,5 @@
+---
+"gitbook": minor
+---
+
+Rework full-width layout, add support for full-width page option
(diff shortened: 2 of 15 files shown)


packages/gitbook/src/components/DocumentView/Blocks.tsx
```diff
@@ -70,8 +70,8 @@ export function UnwrappedBlocks<TBlock extends DocumentBlock>(props: UnwrappedBl
                 style={[
                     'mx-auto w-full decoration-primary/6',
                     node.data && 'fullWidth' in node.data && node.data.fullWidth
-                        ? 'max-w-screen-xl'
-                        : 'max-w-3xl',
+                        ? 'max-w-screen-2xl'
+                        : 'page-full-width:ml-0 max-w-3xl',
```
conico974, line 74: Shouldn't it be `site-full-width` ?

zenoachtig: No, they will serve different purposes.
- `site-full-width` is applied any time the site _needs_ to grow wider to accommodate a bigger block. It's applied for OpenAPI blocks, fullwidth blocks, and for the (soon to be added) fullwidth page option. It basically signals that some UI should no longer be restricted to the container, but go full width instead.
- `page-full-width` is applied _only_ when the (soon to be added) fullwidth **page option** is turned on. It signals that the page's container should be wider.

It's a subtle difference and not the best naming (hardest thing in programming…) but we need both to make different layouts possible.

This way, you can have _both_ a narrow container with a single breakout block (like a big table) that leaves other blocks entered _and_ a wide container where all blocks float left (which is what we need for PayPal).

conico974: Ho ok, i was confused because of the PR description.
I'll take a look tomorrow as to why the e2e test are failing, it seems very random

zenoachtig: Yeah apologies, I wrote it before the final version 😄 I've updated it now.

conico974: No worries.
I found the issue in the e2e in cloudflare about the 404, i'll try to come up with a quick fix in our repo, but ideally a correct fix should go in `@opennextjs/aws`.
For the PDF ones, it has something to do with samy's test on the variables

PR #3295, opened 2025-06-10
Support dynamic expressions

SamyPesse:


packages/gitbook/src/lib/v1.ts
```diff
@@ -241,6 +242,18 @@ function getDataFetcherV1(apiTokenOverride?: string): GitBookDataFetcher {
             );
         },
 
+        getRevisionPageDocument(params) {
```
conico974, line 245: Do we need it on V1?

packages/gitbook-v2/src/lib/data/pages.ts
```diff
@@ -26,5 +41,30 @@ export async function getPageDocument(
         );
     }
 
+    // Pre-fetch the document to start filling the cache before we migrate to this API.
+    if (isInPercentRollout(space.id, 10)) {
```
conico974, line 45: We could do it per site instead here. It probably does not matter that much though.
The idea with that is to increase this value overtime right ?

conico974 (approved):

packages/gitbook/src/lib/v1.ts
SamyPesse: yes I think so, I'm not sure when we'll ship v2 and otherwise we'll be stuck not shipping expressions to all users until we move out of it.

packages/gitbook-v2/src/lib/data/pages.ts
SamyPesse: yes it doesn't matter. It's just to avoid sending too much API requests at once; and instead gradually increase it as our cache fills up

PR #3296, opened 2025-06-11
return null for 404 cached page in the incremental cache

conico974:
This is a temp fix until it is properly fixed in OpenNext


conico974:
> I don't fully get this one, but trust you

FYI, returning null means that we ignore the cache for this one. This is probably what we want to do anyway for 404 on cache interception

PR #3301, opened 2025-06-11
Make heading from stepper visible in outline

gregberge:


gregberge:
> It's hard to know if it works well, should we have some unit tests for it?

See Argos

PR #3303, opened 2025-06-11
Add docs.testgitbook.com to ADAPTIVE_CONTENT_HOSTS list

spastorelli:


packages/gitbook-v2/src/middleware.ts
```diff
@@ -41,6 +41,7 @@ const ADAPTIVE_CONTENT_HOSTS = [
     'docs.gitbook.com',
     'adaptive-docs.gitbook-staging.com',
     'enriched-content-playground.gitbook-staging.io',
+    'docs.testgitbook.com',
```
SamyPesse, line 44: Why do we still need this list?

spastorelli: We should not need it anymore and planning to remove soon.  We are goint to sync with @taranvohra and check that all is good before getting it of it.

PR #3309, opened 2025-06-12
Use dataFetcher itself as the key for preventing concurrent execution

conico974:
Use the datafetcher object itself as the key to prevent concurrent execution instead of `React.cache(() => ({}))`


conico974:
This one is not necessary anymore

PR #3316, opened 2025-06-13
Add margin to adjacent buttons

zenoachtig:


packages/gitbook/src/components/DocumentView/InlineButton.tsx
```diff
@@ -22,6 +22,7 @@ export async function InlineButton(props: InlineProps<api.DocumentInlineButton>)
             label={inline.data.label}
             // TODO: use a variant specifically for user-defined buttons.
             variant={inline.data.kind}
+            className="[&:has(+&)]:mr-2 [&:has(+&)]:mb-2" // Add margin to offset adjacent buttons
```
gregberge: You are adding useless margin for buttons that does not need margins (you will get +2 margin at the bottom, so your card will not look correct for example).

gregberge (approved):

PR #3318, opened 2025-06-13
Experiment with hiding sections without a matching site space

SamyPesse:


packages/gitbook/src/components/SiteSections/encodeClientSiteSections.ts
```diff
@@ -57,6 +61,33 @@ function encodeSection(context: GitBookSiteContext, section: SiteSection) {
...
+    const withNavigateOnlyIfEquivalent = section.id === 'sitesc_4jvEm';
+
+    if (!withNavigateOnlyIfEquivalent) {
+        return true;
+    }
+
+    const { siteSpace: currentSiteSpace } = context;
+    if (section.siteSpaces.length === 1) {
+        return true;
+    }
```
gregberge, line 85: Not sure about that @SamyPesse

SamyPesse: It doesn't matter right now since this demo has multiple languages in all site sections; we can revisit it when we make it a proper option.

My thinking was: if there is only one site space, then it means it's not "multi-variants", so no point on filtering

gregberge: It also means, you select spanish on one section and all your site is only translated in english (without variants added yet), so you should hide english to respect the setting. But yeah we can revisit it.

PR #3319, opened 2025-06-13
Use Promise.all whenever possible

conico974:
This should decrease TTFB for page that needs to be generated

Close RND-7316

### DIFF
--- a/packages/gitbook/src/components/SiteLayout/SiteLayout.tsx
+++ b/packages/gitbook/src/components/SiteLayout/SiteLayout.tsx
@@ -102,30 +102,37 @@ export async function generateSiteLayoutMetadata(context: GitBookSiteContext): P
     const customIcon = 'icon' in customization.favicon ? customization.favicon.icon : null;
 
     const faviconSize = 48;
-    const icons = await Promise.all([
-        {
-            url: customIcon?.light
-                ? await getResizedImageURL(imageResizer, customIcon.light, {
-                      width: faviconSize,
-                      height: faviconSize,
-                  })
-                : linker.toAbsoluteURL(
-                      linker.toPathInSpace('~gitbook/icon?size=small&theme=light')
-                  ),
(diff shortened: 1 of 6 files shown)


packages/gitbook/src/components/SiteLayout/SiteLayout.tsx
```diff
@@ -102,7 +102,7 @@ export async function generateSiteLayoutMetadata(context: GitBookSiteContext): P
     const customIcon = 'icon' in customization.favicon ? customization.favicon.icon : null;
 
     const faviconSize = 48;
-    const icons = [
+    const icons = await Promise.all([
```
emmerich: You'll need to convert the inner objects to async functions:

```typescript
await Promise.all([
   (async () => {
     return {
       url: await getResizedImageURL
     }
   })(),
```

or rearrange the code a bit. Right now you're still doing it in serial.

conico974: Ho yeah you're right

SamyPesse (approved):
It makes sense, but I don't think it'll have much impact, I think we should look more at how Next.js process components rendered as siblings that are suspending: is it serial or parallel?

conico974:
> It makes sense, but I don't think it'll have much impact, I think we should look more at how Next.js process components rendered as siblings that are suspending: is it serial or parallel?

Totally agree, it's just the first step. But in some cases it could still save a couple hundreds of milliseconds, that's an easy win.

One suspended component will not render it's children before actually reaching the JSX component. For some nested deep in multiple `<Suspense />`, we have to wait a lot of time before starting to fetch even though we could fetch it right away.

I'm working on a PR that will prefetch things ASAP in the component tree (basically in the layout/page), and we then use these prefetched data instead of fetching them directly in the component.
The ultimate step would probably be to try to get all the necessary data in one single API call (or as few as possible)

PR #3334, opened 2025-06-16
Don't cache unexpected API errors for more than a few minutes

SamyPesse:


packages/gitbook-v2/src/lib/data/api.ts
```diff
@@ -206,8 +206,8 @@ export function createDataFetcher(
 
 const getUserById = cache(async (input: DataFetcherInput, params: { userId: string }) => {
     'use cache';
-    return trace(`getUserById(${params.userId})`, async () => {
-        return wrapDataFetcherError(async () => {
+    return wrapCacheDataFetcherError(async () => {
+        return trace(`getUserById(${params.userId})`, async () => {
```
SamyPesse, line 210: It swaps both so that we can see the message of the trace as "failed in" instead of succeeded

packages/gitbook-v2/src/lib/data/errors.ts
```diff
@@ -89,6 +90,22 @@ export async function wrapDataFetcherError<T>(
     }
 }
 
+/**
+ * Wrap an async execution to handle errors and return a DataFetcherResponse.
+ * This should be used inside 'use cache' functions.
+ */
+export async function wrapCacheDataFetcherError<T>(
+    fn: () => Promise<T>
```
conico974, line 98: This function should handle all the `cacheLife`. If it fails we set it to minutes, otherwise we set it to whatever we initially wanted

SamyPesse: Why ? we don't always know the cache lifetime that we want before actually getting the API response.

conico974: Mostly to be safe, my idea is to not have a `cacheLife` applied in `fn` and this one not applied.
My idea was to have a second parameter to this function, like an `onSuccess` callback and only call it after the if

SamyPesse: I thought that if only the latest call to `cacheLife` was taken into consideration; I'll check, but I'd prefer to keep the API simple

SamyPesse: @conico974 I prefer to keep it like this for now, it's simpler and we can see how it goes

conico974: It's probably fine to be honest, it's just me being overly cautious.
FYI multiple `cacheLife` should in theory always choose the smallest value.

PR #3336, opened 2025-06-16
Fix rendering of ogimage when logo or icon are AVIF images

SamyPesse:
It fixes crashes when generating docs.hex-rays.com/~gitbook/ogimage/fNhZm7Q9TxnP9XbAa969

### DIFF
--- a/packages/gitbook/src/routes/ogimage.tsx
+++ b/packages/gitbook/src/routes/ogimage.tsx
@@ -8,6 +8,7 @@ import { ImageResponse } from 'next/og';
 import { type PageParams, fetchPageData } from '@/components/SitePage';
 import { getFontSourcesToPreload } from '@/fonts/custom';
 import { getAssetURL } from '@/lib/assets';
+import { getExtension } from '@/lib/paths';
 import { filterOutNullable } from '@/lib/typescript';
 import { getCacheTag } from '@gitbook/cache-tags';
 import type { GitBookSiteContext } from '@v2/lib/context';
@@ -32,7 +33,6 @@ export async function serveOGImage(baseContext: GitBookSiteContext, params: Page
     }
 
     // Compute all text to load only the necessary fonts
-    const contentTitle = customization.header.logo ? '' : site.title;
     const pageTitle = page
         ? page.title.length > 64
             ? `${page.title.slice(0, 64)}...`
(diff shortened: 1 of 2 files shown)


packages/gitbook/src/routes/ogimage.tsx
```diff
@@ -164,10 +164,28 @@ export async function serveOGImage(baseContext: GitBookSiteContext, params: Page
                 )
             )
         );
+        if (!iconImage) {
+            throw new Error('Icon image should always been fetchable');
```
emmerich:
```suggestion
            throw new Error('Icon image should always be fetchable');
```

packages/gitbook/src/routes/ogimage.tsx
```diff
@@ -311,19 +323,46 @@ async function getWithCache<T>(key: string, fn: () => Promise<T>) {
...
+    }
+
     const response = await fetch(url);
 
+    // Filter out unsupported image types
     const contentType = response.headers.get('content-type');
-    if (!contentType || !contentType.startsWith('image/')) {
-        throw new Error(`Invalid content type: ${contentType}`);
+    if (!contentType || !SUPPORTED_IMAGE_TYPES.some((type) => contentType.includes(type))) {
+        return null;
```
emmerich, line 365: Should we continue to log the error to find image types we've missed?

packages/gitbook/src/routes/ogimage.tsx
```diff
@@ -311,19 +323,46 @@ async function getWithCache<T>(key: string, fn: () => Promise<T>) {
...
+];
+
 /**
  * Fetch an image from a URL and return a base64 encoded string.
  * We do this as @vercel/og is otherwise failing on SVG images referenced by a URL.
  */
 async function fetchImage(url: string) {
+    // Skip early some images to avoid fetching them
+    const parsedURL = new URL(url);
+    if (UNSUPPORTED_IMAGE_EXTENSIONS.includes(getExtension(parsedURL.pathname).toLowerCase())) {
```
emmerich, line 356: Should we continue to log the error to find image types we've missed?

packages/gitbook/src/routes/ogimage.tsx
```diff
@@ -334,8 +373,7 @@ async function fetchImage(url: string) {
     try {
         const { width, height } = imageSize(buffer);
         return { src, width, height };
-    } catch (error) {
-        console.error(`Error reading image size: ${error}`);
-        return { src };
+    } catch {
+        return null;
```
emmerich, line 377: Should we continue to log the error?

packages/gitbook/src/routes/ogimage.tsx
SamyPesse: I don't think so, we'll not go over all logs and test the images one by one anyway

packages/gitbook/src/routes/ogimage.tsx
SamyPesse: No, because it's a case where it'll generate an image, so we'll never do anything with it.

PR #3338, opened 2025-06-17
Move style to css

conico974:
This is one of the steps needed for RND-7319.
We move link tailwind classes to a single css class which can reduce a lot the size of the RSC.
This PR alone make it go for one of the RSC from 18.4Mb to 13.2Mb

### DIFF
--- a/packages/gitbook-v2/package.json
+++ b/packages/gitbook-v2/package.json
@@ -25,7 +25,7 @@
     },
     "scripts": {
         "generate": "rm -rf ./public && cp -r ../gitbook/public ./public",
-        "dev:v2": "env-cmd --silent -f ../../.env.local next --turbopack",
+        "dev:v2": "env-cmd --silent -f ../../.env.local next",
         "build": "next build",
         "build:v2": "next build",
         "start": "next start",
--- a/packages/gitbook/src/components/DocumentView/Table/RecordCard.tsx
+++ b/packages/gitbook/src/components/DocumentView/Table/RecordCard.tsx
@@ -7,11 +7,13 @@ import {
 import { LinkBox, LinkOverlay } from '@/components/primitives';
 import { Image } from '@/components/utils';
 import { resolveContentRef } from '@/lib/references';
-import { type ClassValue, tcls } from '@/lib/tailwind';
+import { tcls } from '@/lib/tailwind';
 
 import { RecordColumnValue } from './RecordColumnValue';
 import type { TableRecordKV, TableViewProps } from './Table';
 import { getRecordValue } from './utils';
+
+import styles from './recordCard.module.css';
 
 export async function RecordCard(
     props: TableViewProps<DocumentTableViewCards> & {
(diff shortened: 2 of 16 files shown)


packages/gitbook/src/components/DocumentView/Table/RecordCard.tsx
```diff
@@ -143,35 +145,20 @@ export async function RecordCard(
...
             // a link is rendered inside a link, which is not allowed in HTML.
             // It causes an hydration error in React.
-            <LinkBox href={target.href} className={tcls(style, 'hover:before:ring-tint-12/5')}>
+            <LinkBox
+                href={target.href}
+                className={tcls(
+                    styles.card,
+                    'hover:before:ring-tint-12/5',
+                    'group',
+                    'before:grid-area-1-1'
```
SamyPesse, line 159: Why not move these ones as well?

packages/gitbook/src/components/RootLayout/globals.css
```diff
@@ -151,6 +151,18 @@
             z-index: 20;
         }
     }
+
+    .styled-link {
```
SamyPesse: Why not use CSS modules for these ones?

It makes the DX better (hot reload, etc)

packages/gitbook/src/components/TableOfContents/PageLinkItem.tsx
```diff
@@ -8,6 +8,8 @@ import { tcls } from '@/lib/tailwind';
 
 import { TOCPageIcon } from './TOCPageIcon';
 
+import style from './toc.module.css';
```
SamyPesse: Sometimes you name it `style` and sometimes `styles`

packages/gitbook/src/components/DocumentView/Table/RecordCard.tsx
conico974: `@apply` doesn't work on these ones, tailwind is not letting me use them there. I'll add a small comment

packages/gitbook/src/components/RootLayout/globals.css
conico974: Good point, i'll move them to a CSS modules

packages/gitbook-v2/package.json
```diff
@@ -25,7 +25,7 @@
     },
     "scripts": {
         "generate": "rm -rf ./public && cp -r ../gitbook/public ./public",
-        "dev:v2": "env-cmd --silent -f ../../.env.local next --turbopack",
+        "dev:v2": "env-cmd --silent -f ../../.env.local next",
```
conico974, line 28: There is a bug in next that make css modules fails with turbopack https://github.com/vercel/next.js/issues/79535

conico974:
CSS modules are broken, and @apply doesn't work for every class for some reason.
I'm closing this and will implement a solution using a StyleProvider.

PR #3339, opened 2025-06-17
Improve performance of InlineLinkTooltip

emmerich:
InlineLinkTooltip was having a negative effect on the loading and rendering time of pages for two reasons:
1. Despite being entirely client-side/interaction-based, it was rendered on the server, included in the RSC payload, and hydrated.
2. On pages with 1000s of links, it was causing a very slow browser experience. I think optimizing for this size of page is a bigger task and quite niche.

This PR addresses both of these points:

1. The tooltip is now marked as a client component, and is loaded lazily post-hydration to avoid having an impact on initial render. There's no SEO impact as the tooltips don't contain meaningful SEO content. Even on pages with only a handful of links this is a net benefit.
2. We disable the tooltip on pages with more than 100 links.


packages/gitbook/src/lib/document.ts
```diff
@@ -30,6 +30,32 @@ export function hasFullWidthBlock(document: JSONDocument): boolean {
...
 
+export function hasMoreThan(
+    document: JSONDocument | DocumentBlock,
+    check: (block: DocumentBlock | DocumentInline) => boolean,
+    limit = 1
+): boolean {
+    let count = 0;
+
+    function traverse(node: JSONDocument | DocumentBlock): boolean {
+        for (const child of 'nodes' in node ? node.nodes : []) {
```
conico974, line 44: I think you should also look into `node.fragments`

packages/gitbook/src/components/DocumentView/InlineLink/InlineLink.tsx
```diff
@@ -0,0 +1,153 @@
...
+            return {
+                pageId: resolved.page?.id ?? inline.data.ref.page ?? context.page.id,
+                spaceId: inline.data.ref.space ?? context.space.id,
+            };
+        }
+    })();
+
+    return (
+        <InlineLinkTooltip
+            breadcrumbs={breadcrumbs}
```
conico974, line 138: One thing we could consider as well, is reduce the props key themselves to a minimum. They also get copied for every single instance of the client component.

packages/gitbook/src/components/DocumentView/InlineLink.tsx
```diff
@@ -1,61 +0,0 @@
-import { type DocumentInlineLink, SiteInsightsLinkPosition } from '@gitbook/api';
-
-import { resolveContentRef } from '@/lib/references';
-import { Icon } from '@gitbook/icons';
-import { StyledLink } from '../primitives';
-import type { InlineProps } from './Inline';
-import { InlineLinkTooltip } from './InlineLinkTooltip';
```
gregberge, line 7: You forgot to remove that file right?

packages/gitbook/src/components/DocumentView/DocumentView.tsx
```diff
@@ -28,6 +28,14 @@ export interface DocumentContext {
...
      */
     wrapBlocksInSuspense?: boolean;
+
+    /**
+     * True if link previews should be rendered.
+     * This is used to limit the number of link previews rendered in a document.
+     * If false, no link previews will be rendered.
+     * @default false
+     */
+    shouldRenderLinkPreviews?: boolean;
```
gregberge, line 38: For me limiting the number of links is the only thing with real impact here. The rest of the code is overly complex for very small result.

packages/gitbook/src/lib/document.ts
emmerich: yes 👍

packages/gitbook/src/components/DocumentView/InlineLink/InlineLink.tsx
emmerich: Right, I can do that in a follow-up

packages/gitbook/src/components/DocumentView/DocumentView.tsx
emmerich: It has a decent result on the RSC payload even on pages with less than 10 links, but generally agree with you

packages/gitbook/src/components/DocumentView/InlineLink.tsx
emmerich: No?

packages/gitbook/src/components/DocumentView/DocumentView.tsx
gregberge: I have no doubt but the involved complexity does not worth it IMO. But honestly in this case the doc is just too long.

packages/gitbook/src/components/DocumentView/InlineLink.tsx
gregberge: Hmm weird but if you are sure, anyway not a problem.

CleanShot 2025-06-18 at 11 16 27@2x

PR #3340, opened 2025-06-17
Fix contributing workflow

addisonschultz:


talagluck (commented):
I've gone through the setup steps, but I get a 404 when I try and go to http://localhost:3000/gitbook.com/docs. I also have a bunch of failures of the e2e tests ( but all of the bun commands run).

This is on the latest pull of the branch used in this PR

PR #3343, opened 2025-06-17
fix variant space change when pathname equals space-path

conico974:
If the current page path equals the space path, the `SpacesDropdown` won't work for this space and you'll stay stuck on this specific space.
In this PR, for these cases, we always set the target url to the root of the other spaces

### DIFF
--- a/packages/gitbook/src/components/Header/SpacesDropdown.tsx
+++ b/packages/gitbook/src/components/Header/SpacesDropdown.tsx
@@ -74,6 +74,7 @@ export function SpacesDropdown(props: {
                         url: getSiteSpaceURL(context, otherSiteSpace),
                     }}
                     active={otherSiteSpace.id === siteSpace.id}
+                    currentSpacePath={siteSpace.path}
                 />
             ))}
         </DropdownMenu>
--- a/packages/gitbook/src/components/Header/SpacesDropdownMenuItem.tsx
+++ b/packages/gitbook/src/components/Header/SpacesDropdownMenuItem.tsx
@@ -6,12 +6,19 @@ import { joinPath } from '@/lib/paths';
 import { useCurrentPagePath } from '../hooks';
 import { DropdownMenuItem } from './DropdownMenu';
 
-function useVariantSpaceHref(variantSpaceUrl: string) {
+function useVariantSpaceHref(variantSpaceUrl: string, currentSpacePath: string, active = false) {
     const currentPathname = useCurrentPagePath();
+
+    if (!active && currentPathname.startsWith(currentSpacePath)) {
+        // If the variant space is not active, we need to ensure that we return to the variant space URL
+        // without the current pathname, otherwise we get stuck in the current space.
+        return variantSpaceUrl;
+    }
 
     if (URL.canParse(variantSpaceUrl)) {
         const targetUrl = new URL(variantSpaceUrl);
         targetUrl.pathname = joinPath(targetUrl.pathname, currentPathname);
+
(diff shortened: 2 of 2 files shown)


packages/gitbook/src/components/Header/SpacesDropdown.tsx
```diff
@@ -74,6 +74,7 @@ export function SpacesDropdown(props: {
                         url: getSiteSpaceURL(context, otherSiteSpace),
                     }}
                     active={otherSiteSpace.id === siteSpace.id}
+                    currentSpacePath={siteSpace.path}
```
SamyPesse: Thinking about the RSC issue, even here impact is minimal; should we have only one `SpacesDropdownMenuItems` component that is client side and iterate over the array of variants; instead of serializing some props N times?

packages/gitbook/src/components/Header/SpacesDropdownMenuItem.tsx
```diff
@@ -6,12 +6,19 @@ import { joinPath } from '@/lib/paths';
 import { useCurrentPagePath } from '../hooks';
 import { DropdownMenuItem } from './DropdownMenu';
 
-function useVariantSpaceHref(variantSpaceUrl: string) {
+function useVariantSpaceHref(variantSpaceUrl: string, currentSpacePath: string, active = false) {
     const currentPathname = useCurrentPagePath();
 
+    if (!active && currentPathname.startsWith(currentSpacePath)) {
```
SamyPesse: I don't really understand the bug and the logic here.

But I guess it should not match the case `currentPathname = 'hello-world'` and `currentSpacePath = 'hello'`? 

```suggestion
    if (!active && currentPathname.startsWith(currentSpacePath + '/')) {
```
```suggestion
    if (!active && currentPathname.startsWith(currentSpacePath)) {
```
```suggestion
    if (!active && currentPathname.startsWith(currentSpacePath)) {
```

packages/gitbook/src/components/Header/SpacesDropdownMenuItem.tsx
```diff
@@ -6,12 +6,19 @@ import { joinPath } from '@/lib/paths';
...
+        // If the variant space is not active, we need to ensure that we return to the variant space URL
+        // without the current pathname, otherwise we get stuck in the current space.
```
SamyPesse: This is quite confusing and it might be worth explaining it better

SamyPesse: I honestly don't understand it haha

SamyPesse (approved):

packages/gitbook/src/components/Header/SpacesDropdownMenuItem.tsx
conico974: Haha yeah, i'll update this one. I think the best is just to give an example.

In this case there was 1 space, `developer` which had `developer` as its path and the first page was also `developer`, which means the path is something like `/developer/developer`, but here we don't know if the space will have the page we want so we "reconstruct" an url, by just swapping the space path.
And in the case of the default space, this path is empty and we end up in the original path (i.e. `developer`).
I'll have to ask chatGPT to write this one :smile:

packages/gitbook/src/components/Header/SpacesDropdown.tsx
conico974: Probably a good idea, i don't think this one will have a big impact, but we might as well start making things this way everytime we can

PR #3344, opened 2025-06-17
Add fullwidth page option

zenoachtig:


packages/gitbook/src/components/DocumentView/Blocks.tsx
```diff
@@ -47,6 +47,23 @@ type UnwrappedBlocksProps<TBlock extends DocumentBlock> = DocumentContextProps &
     isOffscreen?: boolean;
 };
 
+/* Blocks that can be full width are automatically expanded on full-width pages.
+ * Ideally we'd rely on the block type to determine if it can be full width, but
+ * the block's `fullWidth` property does not differentiate between `undefined` and `false`.
+ * So instead we hardcode a list of blocks that can be full width. */
```
SamyPesse, line 53: I don't understand this comment, why would these blocks differ from others when the page is full-width?

packages/gitbook/src/components/PageBody/PageBody.tsx
```diff
@@ -40,7 +40,7 @@ export function PageBody(props: {
               LINK_PREVIEW_MAX_COUNT
           )
         : false;
-    const pageFullWidth = page.id === 'wtthNFMqmEQmnt5LKR0q';
+    const pageFullWidth = page.layout.width === 'wide';
```
SamyPesse: We should name it `pageWidthWide` instead and rename the tailwind modifier to match the name

packages/gitbook/src/components/DocumentView/Blocks.tsx
zenoachtig: Because they automatically grow to expand the size. If we didn't do that, the page option would only affect the page's _container_, and you would still have to make each block full-width individually. 

This way we prevent the option of having a wide page with a lot of space on the right, because any fullwidth block (tabs, tables, ...) grows automatically.

zenoachtig: Actually I'm landing on a different solution to this after discussion with @BrettJephson so let's wait until his part is taken care of. Basically instead of hardcoding stuff here we'll just set the components to the right width and in turn read them out on GBO.

SamyPesse: This is still very confusing, and smells very bad that we have an hardcoded set of block types...

I don't have time to deep dive, but yeah it smells fishy

SamyPesse: What does the following mean?

> the block's `fullWidth` property does not differentiate between `undefined` and `false`.

SamyPesse (approved):
Approving, but I'm concerned that the logic is twisted with different behaviour based on a set of block types.

@BrettJephson how is it done in the editor? any idea on how to make this better?

packages/gitbook/src/components/DocumentView/Blocks.tsx
zenoachtig: I agree it's far from perfect and I am very open to better approaches.

> What does the following mean?
> 
> > the block's `fullWidth` property does not differentiate between `undefined` and `false`.

I need to set a block's max width based on if its `fullWidth` prop:
- `fullWidth=undefined`: the block cannot expand, so it should always stay narrow.
- `fullWidth=false`: the block can expand, but is not set to expand by the user. By default we do nothing. On wide pages, we should automatically expand it so the user doesn't have to toggle it manually on every single block.
- `fulLWidth=true`: the block is set to expand by the user. We need to expand it.

The problem is this: the API only returns `undefined` and `true`. If a block is set to expand, the `fullWidth` prop is removed entirely instead of being set to `false`, even though it's an optional boolean so can be `true|false|undefined`.

This means I cannot know on the front-end if I need to show a block narrow (because blocks without `fullWidth` should remain narrow) or expand it (because blocks with `fullWidth=true` _and_ `fullWidth=false` should expand).

Two solutions I see:
1. There's probably a non-strict falsy check somewhere that is removing `fullWidth` instead of setting it to `false` deliberately. If we change that I can read out all three states and remove this hardcoded array.
2. We can explicitly save each block as `fullWidth=true` when the page is set to `wide` layout. This way the front-end can just expand whatever block on which `fullWidth` exists, and keep the rest narrow. But we're then actually editing blocks' markup based on the page they're in, that feels strange.

I think 1 is the easiest solution. But I'm also very open to other ideas I haven't thought of yet.
cc @BrettJephson

PR #3349, opened 2025-06-17
Consistently show variant selector in section bar if site has sections

zenoachtig:


packages/gitbook/src/components/SpaceLayout/SpaceLayout.tsx
```diff
@@ -125,7 +125,7 @@ export function SpaceLayout(props: {
                                             sections={encodeClientSiteSections(context, sections)}
                                         />
                                     )}
-                                    {isMultiVariants && (
+                                    {isMultiVariants && !sections && (
                                         <SpacesDropdown
```
viktorrenkema, line 129: Dont have full context of the code, but I'd expect that if you are now showing the dropdown under more conditions here, you need to make sure it doesn't then _also_ show up in the prior place. With this change, I expect youll have the SpacesDropdown show in 2 places under some conditions, no?

zenoachtig: This line change actually _reduces_ the number of cases where this particular selector shows — we now only show the variant selector in the sidebar if there are no sections (and thus no section bar). 

In `Header.tsx` we used to only show the variant selector if there is no TOC, now we always show the variant selector in the header's section bar. 

Please check my logic but I believe taken together they exactly cover for each other :D

viktorrenkema: Spun up the PR locally and tested it, and I stand corrected :)

viktorrenkema (approved):

PR #3351, opened 2025-06-18
Fix an issue where PDF export URLs were not bringing their query params.

emmerich:
When we rewrite the URL in the middleware, we were losing the query params of the original request. This fixes that.

Here's an example of it working now: https://gitbook-v2-7p90urrx2-gitbook.vercel.app/url/gitbook.com/docs/~gitbook/pdf?page=DfnNkU49mvLe2ythHAyx&only=yes&limit=100


packages/gitbook-v2/src/middleware.ts
```diff
@@ -259,6 +259,12 @@ async function serveSiteRoutes(requestURL: URL, request: NextRequest) {
         console.log(`rewriting ${request.nextUrl.toString()} to ${route}`);
 
         const rewrittenURL = new URL(`/${route}`, request.nextUrl.toString());
+
+        // Ensure all searchParams are brought over to the rewritten URL
+        request.nextUrl.searchParams.forEach((value, key) => {
```
emmerich: Investigating this made me think - how did other features that rely on searchParams (like VA and search) work?

VA works because earlier in the middleware we pull out the token from the URL and set it on the cookie.
Search works because it's all done client-side looking at the current URL.

PDF doesn't work because we rely on the searchParams in the `Page` component post-middleware.

But it makes me wonder: should I include _all_ searchParams in the rewritten URL?

SamyPesse: In theory it should be fine as Next will not take them into consideration for the ISR key (if you use `searchParams`, the page becomes dynamic).

SamyPesse: Why not just:

```
rewrittenURL.search = request.nextUrl.search;
```

?

 Usually it works

packages/gitbook/e2e/internal.spec.ts
```diff
@@ -80,6 +80,14 @@ const testCases: TestsCase[] = [
                 },
                 screenshot: false,
             },
+            {
+                name: 'PDF (all pages)',
+                url: '~gitbook/pdf?limit=100',
+            },
+            {
+                name: 'PDF (only current page)',
+                url: '~gitbook/pdf?page=Bw7LjWwgTjV8nIV4s7rs&only=yes&limit=100',
```
SamyPesse: This might slow down tests a lot no?

Can you remove some of these tests and also put `limit=1` instead?

emmerich: No? I mean it's another test so another screenshot so yeah it'll take longer. I reduced the limits to make it faster but I don't want to remove the test, it'd be too easy to introduce a regression.

packages/gitbook-v2/src/middleware.ts
emmerich: Right, a bit more destructive but we aren't doing anything with `rewrittenURL.search` anyways.

PR #3352, opened 2025-06-18
Implement a StylesProvider

conico974:
This PR introduce a StylesProvider that we can use to pass tailwind css classes to client component.
It reduces the size of the RSC.
With this PR and #3339 , we can in some worst case scenario go from 41.8Mb to 14.3.
And we can iterate on this PR to improve this even further

### DIFF
--- a/.changeset/rich-buses-hunt.md
+++ b/.changeset/rich-buses-hunt.md
@@ -1,0 +1,5 @@
+---
+"gitbook-v2": patch
+---
+
+Fix an issue where PDF export URLs were not keeping their query params.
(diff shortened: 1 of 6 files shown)


gregberge (changes requested):
We already have a solution for that: CSS or Tailwind utilities. Introducing another thing like that is not the way to go.

packages/gitbook/src/components/primitives/StyleProvider.tsx
SamyPesse, line 1: Why do we need a context if the styles are constant? 😅 

Maybe we just need to a file with the class names, and a `type NamedStyle = keyof styles;` and then accept this as styles in some places?

conico974: Yeah that's true, don't know why i didn't think of that :smile:

PR #3353, opened 2025-06-18
Fix Mentions with broken links showing as empty in published content.

emmerich:
Bug report came in because of PDF export. It's easier to find these broken links if they're visible (that's what we do for inline links).

The orginal source of the bug is cross-space links being exported to PDF, but that's a larger fix and not a priority right now.

### DIFF
--- a/packages/gitbook/src/components/DocumentView/Mention.tsx
+++ b/packages/gitbook/src/components/DocumentView/Mention.tsx
@@ -15,7 +15,7 @@ export async function Mention(props: InlineProps<DocumentInlineMention>) {
         : null;
 
     if (!resolved) {
-        return <span className="underline">Broken link</span>;
+        return <span className="underline">⚠️ Page not found</span>;
     }
 
     return (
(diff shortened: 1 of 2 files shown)


packages/gitbook/src/components/DocumentView/Mention.tsx
```diff
@@ -15,7 +15,7 @@ export async function Mention(props: InlineProps<DocumentInlineMention>) {
         : null;
 
     if (!resolved) {
-        return null;
+        return <span className="underline">Broken link</span>;
```
SamyPesse: Is it actually good to expose them to the end reader?

emmerich: Yes, I think so. Otherwise it's very hard to track them down as an author. It's what we do for inline links.

As a visitor the experience is about the same. "Broken link" or a big empty space, both look bad. At least broken link is something that can be actioned on.

Crazy idea: render a link that the visitor can click that takes them to a "Notify the author" page in GitBook 😉

emmerich:
After a bit of reflection I changed my mind.  We should flag this internally in GitBook, not to the reader. Mentions are often kept "contextless" and used like tags, so it doesn't really break the flow of the document.

PR #3356, opened 2025-06-19
Update SpacesDropdown to manage active state based on current space

conico974:
Pass the current space ID to the `SpacesDropdownMenuItems` to correctly set the active state for each item.

### DIFF
--- a/packages/gitbook/src/components/Header/SpacesDropdown.tsx
+++ b/packages/gitbook/src/components/Header/SpacesDropdown.tsx
@@ -70,9 +70,9 @@ export function SpacesDropdown(props: {
                     id: space.id,
                     title: space.title,
                     url: getSiteSpaceURL(context, space),
+                    isActive: space.id === siteSpace.id,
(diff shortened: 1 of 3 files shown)


packages/gitbook/src/components/Header/SpacesDropdown.tsx
```diff
@@ -72,6 +72,7 @@ export function SpacesDropdown(props: {
                     url: getSiteSpaceURL(context, space),
                 }))}
                 curPath={siteSpace.path}
+                curSpaceId={siteSpace.id}
```
SamyPesse: Why don't we just pass `active` in `slimSpaces`?

SamyPesse (approved):

packages/gitbook/src/components/Header/SpacesDropdown.tsx
conico974: Yeah we could do that. I'll change it.

PR #3357, opened 2025-06-19
Filter out malicious request

conico974:
This PR aims to filter out some malicious request directly in the middleware : 
- Invalid server actions
- Request for hostname containing a port

### DIFF
--- a/packages/gitbook-v2/src/middleware.ts
+++ b/packages/gitbook-v2/src/middleware.ts
@@ -64,10 +64,11 @@ export async function middleware(request: NextRequest) {
 
 async function validateServerActionRequest(request: NextRequest) {
     // We need to reject incorrect server actions requests
-    if (request.headers.has('next-action')) {
+    // We do not do it in cloudflare workers as there is a bug that prevents us from reading the request body.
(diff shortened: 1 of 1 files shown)


packages/gitbook-v2/src/middleware.ts
```diff
@@ -56,6 +62,25 @@ export async function middleware(request: NextRequest) {
...
 }
 
+async function validateServerActionRequest(request: NextRequest) {
+    // We need to reject incorrect server actions requests
+    // We do not do it in cloudflare workers as there is a bug that prevents us from reading the request body.
+    if (request.headers.has('next-action') && process.env.GITBOOK_RUNTIME !== 'cloudflare') {
+        // We just test that the json body is parseable
+        try {
+            const clonedRequest = request.clone();
+            await clonedRequest.json();
```
gregberge, line 72: So now we parse the JSON twice right? I don't think it's a good idea for perf.

conico974: Only for server actions, and most of the time the JSON will be very small.

gregberge (approved):

PR #3359, opened 2025-06-19
Fix CodeBlock rendering

gregberge:
- Move CSS to global to avoid conflict (also CSS imported are not
  supported in dev)
- Fix a very weird layout issue that makes our lines to use a grid (very
  weird)

Tested JSON5, it works great.


gregberge:
> Something is broken if you add lines

Thanks, I missed the subgrid thing

conico974 (approved):

PR #3362, opened 2025-06-19
Scroll to active TOC when clicking a link

gregberge:


conico974 (approved):
Way better UX this way

PR #3364, opened 2025-06-19
Allow to zoom image on mobile if relevant

gregberge:


packages/gitbook/src/components/utils/ZoomImage.tsx
```diff
@@ -28,14 +30,9 @@ export function ZoomImage(
 
     // Only allow zooming when image will not actually be larger and on mobile
     React.useEffect(() => {
-        if (isTouchDevice()) {
```
SamyPesse, line 31: Why allow zoom on mobile? It'a a bad experience usually compared to zoom with the fingers

gregberge: Because it was asked by a customer: [RND-7089: Allow zooming images on mobile](https://linear.app/gitbook-x/issue/RND-7089/allow-zooming-images-on-mobile)

And I think it's relevant because when you have small images like that it's good to be able to just touch it to expand it. It does not prevent from zooming with two fingers, so it's just a +.

SamyPesse (approved):

PR #3369, opened 2025-06-20
Properly redirect to the root page

conico974:
Fix the linker methods to never return an empty path.
Close RND-7230

### DIFF
--- a/packages/gitbook-v2/src/lib/links.test.ts
+++ b/packages/gitbook-v2/src/lib/links.test.ts
@@ -35,7 +35,7 @@ describe('toPathInSpace', () => {
         expect(variantInSection.toPathInSpace('some/path/')).toBe('/section/variant/some/path');
     });
 
-    it('should not add a trailing slash', () => {
+    it('should not add an unnecessary trailing slash', () => {
         // The index page should not be an empty path
         expect(root.toPathInSpace('')).toBe('/');
         expect(variantInSection.toPathInSpace('')).toBe('/section/variant');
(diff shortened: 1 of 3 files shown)


emmerich (approved):

packages/gitbook-v2/src/lib/links.test.ts
```diff
@@ -36,7 +36,8 @@ describe('toPathInSpace', () => {
     });
 
     it('should not add a trailing slash', () => {
```
SamyPesse: Well this test losts its meaning 😅

packages/gitbook-v2/src/lib/links.test.ts
```diff
@@ -53,7 +54,8 @@ describe('toPathInSite', () => {
...
```
SamyPesse: Same thing

packages/gitbook-v2/src/lib/links.ts
```diff
@@ -128,7 +128,8 @@ export function createLinker(
 function joinPaths(prefix: string, path: string): string {
     const prefixPath = prefix.endsWith('/') ? prefix : `${prefix}/`;
     const suffixPath = path.startsWith('/') ? path.slice(1) : path;
-    return removeTrailingSlash(prefixPath + suffixPath);
+    const pathWithoutTrailingSlash = removeTrailingSlash(prefixPath + suffixPath);
+    return pathWithoutTrailingSlash === '' ? '/' : pathWithoutTrailingSlash;
```
SamyPesse, line 132: I'm really really not sure that the fix should be there

i think it'll break the case where right now we have `/sitebasepath/` as a base path and instead of doing a link to `/sitebasepath/`, it'll do it to `/`.

I'm not sure how we can do to cover this case with a test somewhere

packages/gitbook-v2/src/lib/links.test.ts
conico974: We need to ensure that it doesn't return `""` for no basePath, i'll rename the test to `should not add an unnecessary trailing slash`

packages/gitbook-v2/src/lib/links.ts
conico974: It should not do that actually, it only applies if both the prefixPath and the suffixPath are empty.
And given where they are used, it cannot happen (unless the data provided are incorrect, but that's a totally different issue)

PR #3376, opened 2025-06-23
Optional label in OpenAPI x-codeSamples

nolannbiron:


packages/react-openapi/src/OpenAPICodeSample.tsx
```diff
@@ -253,15 +253,11 @@ function getCustomCodeSamples(props: {
...
-                        typeof sample.label === 'string' &&
-                        typeof sample.source === 'string' &&
-                        typeof sample.lang === 'string'
-                    );
+                    return typeof sample.source === 'string' && typeof sample.lang === 'string';
                 })
                 .map((sample, index) => ({
                     key: `custom-sample-${sample.lang}-${index}`,
-                    label: sample.label,
+                    label: sample.label ?? `Sample ${index + 1}`,
```
emmerich:
```suggestion
                    label: sample.label ?? sample.lang,
```

I believe in the Redocly implementation, the `lang` is used if the label is not defined - can you confirm it?

gregberge:
```suggestion
                    label: sample.label || `Sample ${index + 1}`,
```

emmerich (approved):
Approving assuming the fallback to `lang` is correct!

PR #3377, opened 2025-06-23
Avoid unnecessary proxy URL lookups

conico974:
Right now we run `lookupPublishedContentByUrl` on these url `https://proxy.gitbook.site` and `https://proxy.gitbook.site/sites`


packages/gitbook-v2/src/lib/data/urls.ts
```diff
@@ -39,6 +39,14 @@ export function getURLLookupAlternatives(input: URL) {
             return;
         }
 
+        // We don't want to push the url if it is the root proxy URL (either https://proxy.gitbook.site or https://proxy.gitbook.site/sites)
+        if (
+            adding.hostname === 'proxy.gitbook.site' &&
+            (adding.pathname === '/' || adding.pathname === '/sites/')
+        ) {
+            return;
+        }
```
SamyPesse: We should not hardcode these values, but instead combine this with `packages/gitbook-v2/src/lib/proxy.ts`

SamyPesse: Also we should add tests for it in `packages/gitbook-v2/src/lib/data/urls.test.ts`

I guess the goal is to only return one alternative here instead of multiple, correct? The test should represent this

conico974: Right i missed the proxy file, I'll make the changes

packages/gitbook-v2/src/lib/data/urls.test.ts
```diff
@@ -362,6 +362,28 @@ describe('getURLLookupAlternatives', () => {
             ],
         });
     });
+
+    it('should skip proxy root URLs', () => {
+        expect(
+            getURLLookupAlternatives(new URL('https://proxy.gitbook.site/site_123/doc1'))
```
SamyPesse: I think the real case is more something like `https://proxy.gitbook.site/sites/site_foo/hello/world`

SamyPesse (approved):

PR #3380, opened 2025-06-23
Remove CSS module from table and simplify layout

gregberge:


gregberge:
I will close it, let's wait for the migration on 2c to have CSS module working first.

PR #3382, opened 2025-06-23
Optimize fetching of revisions

SamyPesse:
Second step of https://github.com/GitbookIO/gitbook/pull/3379

It ensures we always fetch revisions the same way for a space, maximizing the odds of cache hits.


packages/gitbook-v2/src/lib/data/revisions.ts
```diff
@@ -0,0 +1,56 @@
+import type { Revision, RevisionFile, RevisionReusableContent, Space } from '@gitbook/api';
+import * as React from 'react';
+import type { GitBookDataFetcher } from './types';
+
+const getRevisionReusableContents = React.cache((revision: Revision) => {
```
conico974, line 4: I think we should use https://github.com/GitbookIO/gitbook/blob/4f7c0eea9fae4ce035086bea2fa3c58efffb4b0b/packages/gitbook-v2/src/lib/cache.ts#L8 instead of `React.cache`

SamyPesse: No it's not needed as in that case it can use `Object.is`

Our custom cache is more expensive as it needs to hash the object, which is huge here

conico974 (approved):

PR #3385, opened 2025-06-24
Hooks and APIs for AI v2

SamyPesse:


packages/gitbook/src/components/AI/server-actions/chat.ts
```diff
@@ -0,0 +1,117 @@
...
+    previousResponseId,
+    options,
+}: {
+    message: string;
+    messageContext: AIMessageContext;
+    previousResponseId?: string;
+    options?: RenderAIMessageOptions;
+}) {
+    const context = await getServerActionBaseContext();
+    const siteURLData = await getSiteURLDataFromMiddleware();
```
conico974, line 54: You could get it from the context directly

SamyPesse: No, because `getServerActionBaseContext` returns a very basic context, not fetching anything.
`getSiteURLDataFromMiddleware` returns the data coming from the headers.

Basically these 2 methods are fast and do not do any HTTP requests

conico974 (approved):
Just one small optimization

packages/gitbook/src/components/AI/server-actions/chat.ts
conico974: Yeah but `getServerActionBaseContext` already calls `getSiteURLDataFromMiddleware`, you could return it from there

PR #3387, opened 2025-06-24
Show section tabs when there is a single section group present

zenoachtig:


gregberge (approved):
A bit cryptic written like that but yeah good 👌

PR #3390, opened 2025-06-24
Fix ScrollSectionsList with suspended component

conico974:
Enhance the ScrollSectionsList component to work correctly with suspended components by integrating a loading state context. 

Before this PR and in SSR, suspended elements could be improperly observed.

It also removes `<Suspense />` in `Block` which can decrease the size of the RSC by up to 2Mb in some worst case scenario. (This doesn't have a big impact on user now since most page are ISR/SSG and thus Suspense won't work in these cases)

### DIFF
--- a/packages/gitbook/src/components/DocumentView/Block.tsx
+++ b/packages/gitbook/src/components/DocumentView/Block.tsx
@@ -1,5 +1,4 @@
 import type { DocumentBlock, JSONDocument } from '@gitbook/api';
-import React from 'react';
 
 import {
     SkeletonCard,
@@ -47,7 +46,7 @@ export interface BlockProps<Block extends DocumentBlock> extends DocumentContext
 }
 
 export function Block<T extends DocumentBlock>(props: BlockProps<T>) {
-    const { block, style, isEstimatedOffscreen, context } = props;
+    const { block } = props;
 
     const content = (() => {
         switch (block.type) {
@@ -116,17 +115,7 @@ export function Block<T extends DocumentBlock>(props: BlockProps<T>) {
         }
     })();
 
-    if (!isEstimatedOffscreen || context.wrapBlocksInSuspense === false) {
-        // When blocks are estimated to be on the initial viewport, we render them immediately
-        // to avoid a flash of a loading skeleton.
-        return content;
-    }
-
-    return (
-        <React.Suspense fallback={<BlockSkeleton block={block} style={style} />}>
-            {content}
-        </React.Suspense>
-    );
+    return content;
 }
 
 /**
--- a/packages/gitbook/src/components/PageAside/ScrollSectionsList.tsx
+++ b/packages/gitbook/src/components/PageAside/ScrollSectionsList.tsx
@@ -7,6 +7,7 @@ import type { DocumentSection } from '@/lib/document-sections';
 import { tcls } from '@/lib/tailwind';
 
 import { HEADER_HEIGHT_DESKTOP } from '../layout';
+import { useBodyLoaded } from '../primitives/LoadingStateProvider';
 import { AsideSectionHighlight } from './AsideSectionHighlight';
 
 /**
@@ -30,9 +31,12 @@ export function ScrollSectionsList(props: { sections: DocumentSection[] }) {
         });
     }, [sections]);
 
+    const bodyLoaded = useBodyLoaded();
+
     const activeId = useScrollActiveId(ids, {
         rootMargin: `-${HEADER_HEIGHT_DESKTOP}px 0px -40% 0px`,
         threshold: SECTION_INTERSECTING_THRESHOLD,
+        additionalEffects: bodyLoaded,
     });
 
     return (
--- a/packages/gitbook/src/components/PageBody/PageBody.tsx
+++ b/packages/gitbook/src/components/PageBody/PageBody.tsx
@@ -11,6 +11,7 @@ import { DocumentView, DocumentViewSkeleton } from '../DocumentView';
 import { TrackPageViewEvent } from '../Insights';
 import { PageFeedbackForm } from '../PageFeedback';
 import { DateRelative } from '../primitives';
+import { SuspenseLoadedHint } from '../primitives/LoadingStateProvider';
 import { PageBodyBlankslate } from './PageBodyBlankslate';
 import { PageCover } from './PageCover';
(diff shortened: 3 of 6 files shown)


bangaom59 (approved):

packages/gitbook/src/components/DocumentView/Block.tsx
```diff
@@ -116,17 +115,7 @@ export function Block<T extends DocumentBlock>(props: BlockProps<T>) {
         }
     })();
 
-    if (!isEstimatedOffscreen || context.wrapBlocksInSuspense === false) {
-        // When blocks are estimated to be on the initial viewport, we render them immediately
-        // to avoid a flash of a loading skeleton.
-        return content;
-    }
```
SamyPesse, line 123: Why do you remove this?

Isn't it useful in the dynamic route case?

packages/gitbook/src/components/PageAside/ScrollSectionsList.tsx
```diff
@@ -7,6 +7,7 @@ import type { DocumentSection } from '@/lib/document-sections';
 import { tcls } from '@/lib/tailwind';
 
 import { HEADER_HEIGHT_DESKTOP } from '../layout';
+import { useBodyLoaded } from '../primitives/LoadingStateProvider';
```
SamyPesse: I think we should do more import as:

```suggestion
import { useBodyLoaded } from '@/components/primitives';
```

packages/gitbook/src/components/primitives/LoadingStateProvider.tsx
```diff
@@ -0,0 +1,50 @@
...
+ */
+export function LoadingStateProvider({
+    children,
+}: {
+    children: React.ReactNode;
+}) {
+    const [bodyLoaded, setBodyLoaded] = useState(false);
+
+    return (
+        <LoadingStateProviderContext.Provider value={{ bodyLoaded, setBodyLoaded }}>
```
SamyPesse, line 29: You can't do this, it'll re-render every component that listens to this

Use `useMemo`

Otherwise you can also use `zustand`

packages/gitbook/src/components/PageAside/ScrollSectionsList.tsx
```diff
@@ -30,9 +31,12 @@ export function ScrollSectionsList(props: { sections: DocumentSection[] }) {
         });
     }, [sections]);
 
+    const bodyLoaded = useBodyLoaded();
+
     const activeId = useScrollActiveId(ids, {
         rootMargin: `-${HEADER_HEIGHT_DESKTOP}px 0px -40% 0px`,
         threshold: SECTION_INTERSECTING_THRESHOLD,
+        additionalEffects: bodyLoaded,
```
SamyPesse: I think we should name it `active` or `enabled`

and only scrolled when it's `true`

packages/gitbook/src/components/DocumentView/Block.tsx
conico974: It's barely useful, you still have the top Loader at the `PageBody` level, 99% of the component dont't need that.
If we really need this it should be on the problematic component (Maybe things like integration or something like that.

Now each paragraph for example is surrounded by a `<Suspense>`

packages/gitbook/src/components/primitives/LoadingStateProvider.tsx
conico974: This one is not supposed to be accessed as is. You're supposed to use `useBodyLoaded` which use `useMemo`, but i could just do it here i guess

packages/gitbook/src/components/DocumentView/Block.tsx
SamyPesse: It makes sense, it was needed when we were streaming the page, and even a paragraph could suspend as it has links in it.

But now dynamic pages are a edge-case and as long as we have a suspense boundary over the entire docs, it's fine!

packages/gitbook/src/components/PageBody/PageBody.tsx
```diff
@@ -12,6 +12,7 @@ import { TrackPageViewEvent } from '../Insights';
 import { PageFeedbackForm } from '../PageFeedback';
 import { CurrentPageProvider } from '../hooks/useCurrentPage';
 import { DateRelative } from '../primitives';
+import { SuspenseLoadedHint } from '../primitives/LoadingStateProvider';
```
SamyPesse: We can group both.

I'm not sure how we could add more lint rules for imports; feel free to ignore my comments on this as the code already has plenty of messy imports

packages/gitbook/src/components/hooks/useScrollActiveId.ts
```diff
@@ -8,9 +8,10 @@ export function useScrollActiveId(
     options: {
         rootMargin?: string;
         threshold?: number;
-    } = {}
+        active: boolean;
+    } = { active: true }
 ) {
-    const { rootMargin, threshold = 0.5 } = options;
+    const { rootMargin, threshold = 0.5, active } = options;
```
SamyPesse: Actually `active` is a bit confusing as a named compared to `enabled` as there is `activeId` in it :D but it's fine

packages/gitbook/src/components/hooks/useScrollActiveId.ts
```diff
@@ -70,7 +71,7 @@ export function useScrollActiveId(
         return () => {
             observer.disconnect();
         };
-    }, [ids, threshold, rootMargin]);
+    }, [ids, threshold, rootMargin, active]);
```
SamyPesse: I think you forgot to do `if (!active) { return }` in the `useEffect`

packages/gitbook/src/components/primitives/LoadingStateProvider.tsx
SamyPesse: It doesn't matter if the client component that uses this context uses `useMemo`.

If you do `value={{ something }}` in aReact context provder, it means every time the provider re-renders, all the clients that calls the associated `useContext` will re-render; because it uses `Object.is` to compare values.

```suggestion
        <LoadingStateProviderContext.Provider value={React.useMemo(() => ({ bodyLoaded, setBodyLoaded }), [bodyLoaded, setBodyLoaded])}>
```

SamyPesse (approved):

PR #3392, opened 2025-06-25
Fix an issue with link resolution in cross-space reusable content.

emmerich:
Bit of a complex one. If Space A references a reusable content to Space B, and that reuc has a relative link to a page, the link was not resolved.

The content ref will be relative, like `{ kind: 'page', page: 'pageId' }` so we won't find it in the pages array.

When we render ReusableContent we empty the `pages` array to reduce API requests and prevent broken links across spaces that are not published in a site. But with recent changes, we have the pages available to us out of the box.

- ReusableContent now includes `revision.pages` in its context so all content refs are resolved as if they were in Space B.
- ReusableContent creates the appropriate linker - ensuring links will be absolute if they're across spaces.

### DIFF
--- a/.changeset/slow-boxes-approve.md
+++ b/.changeset/slow-boxes-approve.md
@@ -1,0 +1,5 @@
+---
+'gitbook': patch
+---
+
+Include page group children under the .md route
--- a/packages/gitbook/e2e/internal.spec.ts
+++ b/packages/gitbook/e2e/internal.spec.ts
@@ -465,6 +465,22 @@ const testCases: TestsCase[] = [
         ],
     },
     {
+        name: '[page].md',
+        skip: process.env.ARGOS_BUILD_NAME !== 'v2-vercel',
+        contentBaseURL: 'https://gitbook.gitbook.io/test-gitbook-open/',
+        tests: [
+            {
+                name: 'blocks.md',
+                url: 'blocks.md',
+                screenshot: false,
+                run: async (_page, response) => {
+                    expect(response?.status()).toBe(200);
+                    expect(response?.headers()['content-type']).toContain('text/markdown');
+                },
+            },
+        ],
+    },
+    {
         name: 'Site subdirectory (proxy)',
         skip: process.env.ARGOS_BUILD_NAME !== 'v2-vercel',
         contentBaseURL: 'https://nextjs-gbo-proxy.vercel.app/documentation/',
--- a/packages/gitbook/src/routes/markdownPage.ts
+++ b/packages/gitbook/src/routes/markdownPage.ts
@@ -1,10 +1,16 @@
 import type { GitBookSiteContext } from '@/lib/context';
-import { throwIfDataError } from '@/lib/data';
+import { throwIfDataError } from '@/lib/data/errors';
 import { resolvePagePath } from '@/lib/pages';
-import { RevisionPageType } from '@gitbook/api';
+import { type RevisionPage, type RevisionPageDocument, RevisionPageType } from '@gitbook/api';
+import { pMapIterable } from 'p-map';
+
+// We limit the concurrency to 100 to avoid reaching limit with concurrent requests
+// or file descriptor limits.
+const MAX_CONCURRENCY = 100;
 
 /**
- * Generate a markdown version of a page.
+ * Generate a markdown version of a page with streaming for better performance.
+ * For pages with many children, this streams the output to avoid memory issues.
  */
 export async function servePageMarkdown(context: GitBookSiteContext, pagePath: string) {
     const pageLookup = resolvePagePath(context.revision.pages, pagePath);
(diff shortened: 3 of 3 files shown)


packages/gitbook/src/components/DocumentView/ReusableContent.tsx
```diff
@@ -46,23 +47,32 @@ export async function ReusableContent(props: BlockProps<DocumentBlockReusableCon
...
+        if (context.contentContext.space.id === reusableContent.space.id) {
+            return context.contentContext;
+        }
+
+        // References inside reusable content from a different space need to resolve in the parent space.
+        // Create a linker that ensures links are resolved with the correct parent, and are kept absolute.
+        const ctx = await createLinkerForSpace(reusableContent.space.id, context.contentContext);
+
+        if (!ctx) {
+            throw new Error(`Could not create context for space ${reusableContent.space.id}`);
```
emmerich: i shouldn't throw here, i'll fix it

packages/gitbook/src/components/DocumentView/ReusableContent.tsx
```diff
@@ -46,23 +48,56 @@ export async function ReusableContent(props: BlockProps<DocumentBlockReusableCon
...
+        }
+
+        // References inside reusable content from a different space need to resolve in the parent space.
+        // Create a linker that ensures links are resolved with the correct parent, and are kept absolute.
+        const ctx = await createLinkerForSpace(reusableContent.space.id, context.contentContext);
+
+        if (!ctx) {
+            // TODO: we should never have to reach this point - it means we couldn't resolve the space context for the reusable content
+            // but that should never happen as we've already fetched it at this point.
+            // Rather than throw, resolving using app URLs if needed.
```
SamyPesse: Hum I think we should throw or not render the reusable content.

Having in-app url there will be hard to catch and understand

SamyPesse: Also it's a complex logic

packages/gitbook/src/lib/links.ts
```diff
@@ -66,49 +66,63 @@ export function createLinker(
 
         /** The base path of the site */
         siteBasePath: string;
-    }
+    },
+    linkerOptions: {
+        /** If true, any links returned by this linker will be absolute. */
+        alwaysAbsolute?: boolean;
```
SamyPesse: I'm slightly concerned about this as it means the linker will return something that is not expected from it.

`toPathInSpace` is supposed to return a path, not a URL.

I'll think of an alternative

SamyPesse: Also instead of putting options in the linker

it should be a function `withAbsoluteURL(linker: Linker): Linker`

It'll be cleaner and easier to add tests for it

packages/gitbook/src/lib/links.ts
```diff
@@ -66,49 +66,63 @@ export function createLinker(
...
+            return wrapInAbsolute(normalizedPath.slice(servedOn.siteBasePath.length));
         },
 
         toPathForPage({ pages, page, anchor }) {
             return linker.toPathInSpace(getPagePath(pages, page)) + (anchor ? `#${anchor}` : '');
         },
 
+        toAbsoluteURL(absolutePath: string): string {
+            return toAbsoluteURL(absolutePath);
+        },
```
SamyPesse:
```suggestion
        toAbsoluteURL,
```

packages/gitbook/src/lib/references.tsx
```diff
@@ -360,6 +354,43 @@ async function resolveContentRefInSpace(
...
+            {
+                label: ctx.space.title,
+                href: ctx.baseURL.toString(),
+            },
+            ...(resolved.ancestors ?? []),
+        ].filter(filterOutNullable),
+    };
+}
+
+export async function createLinkerForSpace(
```
SamyPesse: We should comment to explain what it does, as the function name can be interpreted in different ways.

packages/gitbook/src/lib/references.tsx
```diff
@@ -360,6 +354,43 @@ async function resolveContentRefInSpace(
...
+    spaceId: string,
+    context: GitBookAnyContext
+): Promise<{
+    spaceContext: GitBookSpaceContext;
+    linker: GitBookLinker;
+    space: Space;
```
SamyPesse: Why do you need `space` here? isn't it included in `spaceContext`?

packages/gitbook/src/lib/references.tsx
```diff
@@ -360,6 +354,43 @@ async function resolveContentRefInSpace(
...
+    baseURL: URL;
```
SamyPesse, line 376: What is `baseURL`? pleae comment if it's not also something you can get from the context

packages/gitbook/src/lib/links.ts
emmerich: That's hard to do though, because the linker is created in `ReusableContent`, `resolveContentRef` doesn't know whether it's in a reuc or not so it has no way to figure out whether it needs absolute URLs or not.

We'd have to put something on the context to say "you're in an environment that requires absolute URLs" which isn't _awful_ but I did think keeping it in the linker was a more elegant solution.

packages/gitbook/src/components/DocumentView/ReusableContent.tsx
emmerich: alright I'll return null but I think I do worry it'll lead to silent bugs

packages/gitbook/src/lib/references.tsx
emmerich: yeah, removed it

packages/gitbook/src/lib/references.tsx
emmerich: This code already existed, I just pulled it into its own function. createLinkerInSpace is going to resolve the best space for the given spaceId. It returns the baseURL so when we resolve a ref in a different space we have the correct root-level ancestor. See https://github.com/GitbookIO/gitbook/pull/3392/files#diff-b146299347c6848e8024df124628d5a8bbbaa9cc9e62da3dc4680758f48060bbR378

again, this is how the code worked before

packages/gitbook/src/components/DocumentView/ReusableContent.tsx
SamyPesse: Then we should throw an error? I think resolving in-app is the worst of the 3 options

SamyPesse (approved):

PR #3399, opened 2025-06-26
Support external images in cards cover

nolannbiron:


packages/gitbook/src/components/DocumentView/Table/utils.ts
```diff
@@ -4,10 +4,9 @@ import assertNever from 'assert-never';
 /**
  * Get the value for a column in a record.
  */
-export function getRecordValue<T extends number | string | boolean | string[] | ContentRef>(
-    record: DocumentTableRecord,
-    definitionId: string
-): T {
+export function getRecordValue<
+    T extends number | string | boolean | string[] | ContentRef | ContentRef[],
```
SamyPesse, line 8: Why can it be an array of ContentRef?

Why would a cover column referenes multiples content refs

nolannbiron: it's to handle `images` type, similar to files, but ContentRef instead of string to get the difference between file and url content-ref

PR #3401, opened 2025-06-27
Correctly handle OpenAPI errors

gregberge:


packages/gitbook/src/lib/openapi/resolveOpenAPIOperationBlock.ts
```diff
@@ -55,8 +55,8 @@ async function baseResolveOpenAPIOperationBlock(
 
         return { data, specUrl };
     } catch (error) {
-        if (error instanceof OpenAPIParseError) {
-            return { error };
+        if (error instanceof Error && error.name === 'OpenAPIParseError') {
```
gregberge, line 58: Not sure why it didn't work, but realm, etc... Safer like that.

nolannbiron (approved):

PR #3402, opened 2025-06-27
Fix markdown page generation for groups

nolannbiron:


packages/gitbook/src/lib/pages.ts
```diff
@@ -1,53 +1,99 @@
...
 } from '@gitbook/api';
+import type { ListItem, Paragraph, RootContent } from 'mdast';
 
 export type AncestorRevisionPage = RevisionPageDocument | RevisionPageGroup;
 
+interface ResolvePageOptions {
+    /**
+     * Whether to include page groups in the results.
+     */
+    includePageGroup?: boolean;
```
SamyPesse: The naming is a bit confusing, as I first thought you meant "in the ancestors".
```suggestion
    /**
     * Whether to return a page group if it matches the path. If not it'll resolve to the first document in it.
     * @default false
     */
    returnPageGroup?: boolean;
```

packages/gitbook/src/lib/pages.ts
```diff
@@ -1,53 +1,99 @@
...
-        for (const page of pages) {
-            if (page.type === RevisionPageType.Link || page.type === RevisionPageType.Computed) {
-                continue;
-            }
+    pagePath: string,
+    options?: { includePageGroup: true }
+):
+    | { page: RevisionPageGroup | RevisionPageDocument; ancestors: AncestorRevisionPage[] }
+    | undefined;
+export function resolvePagePath(
```
SamyPesse: I'm usually not a big fan of functions with multiple signatures.

I don't know if possible in this case, but it's usually better to have composition:

```
function resolvePagePath(rootPages, pagePath) {
   const page = resolveAnyPagePath(rootPages, pagepath)
   if (page.type === 'group') {
     return resolveFirstPageDocument(...);
   }

   return page;
}

function resolveAnyPagePath(...)
```

packages/gitbook/src/lib/pages.ts
```diff
@@ -214,3 +260,55 @@ function flattenPages(
 
     return result;
 }
+
+export async function getPagesTree(
```
SamyPesse: If we export it, we should comment it and its props

Especially as the props names are not clear

SamyPesse: Looking more into it; I really don't think this code should be there, it's specific to markdown.

Move it back to `llms.ts` and import the function from there in `markdownPage.ts`

SamyPesse:
```suggestion
export async function getMarkdownForPagesTree(
```

packages/gitbook/src/lib/pages.ts
```diff
@@ -8,17 +8,68 @@ import {
 
 export type AncestorRevisionPage = RevisionPageDocument | RevisionPageGroup;
 
+type ResolvedPageDocument = {
```
SamyPesse: Haha I'm a pain in the ass, but this could be a nice type with a generic

```ts
type ResolvedPagePath<Page extends RevisionPageDocument | RevisionPageGroup> = {
  page: Page;
  ancestors: AncestorRevisionPage[];
}
```

packages/gitbook/src/routes/llms.ts
```diff
@@ -140,46 +141,68 @@ async function getNodesFromSiteSpaces(
...
+        })
+    );
+    const nodes: RootContent[] = [];
+    if (options.heading) {
+        nodes.push({
+            type: 'heading',
+            depth: 2,
+            children: [{ type: 'text', value: options.heading }],
+        });
+    }
```
SamyPesse: This doesn't need to be done in this function; the place that calls this function can just put the heading before calling the function

Better to limit the scope of each low-level function

packages/gitbook/src/routes/markdownPage.ts
```diff
@@ -1,123 +1,117 @@
...
-/**
- * Fetch markdown from a page.
- */
-async function fetchMarkdown(
-    context: GitBookSiteContext,
-    page: RevisionPageDocument
-): Promise<string> {
-    const markdown = await throwIfDataError(
-        context.dataFetcher.getRevisionPageMarkdown({
+    const document = await throwIfDataError(
```
SamyPesse: Hum this is actually going to be quite expensive, as it means for each page, we'll fetch the content of the page twice, once as JSON, once as Markdown.

Can't we detect the "empty" case from the actual markdown fetched?

packages/gitbook/src/routes/markdownPage.ts
```diff
@@ -1,123 +1,127 @@
...
+    const tree = fromMarkdown(pageMarkdown, {
+        extensions: [frontmatter(['yaml']), gfm()],
+        mdastExtensions: [frontmatterFromMarkdown(['yaml']), gfmFromMarkdown()],
+    });
+
+    // Remove frontmatter
+    remove(tree, 'yaml');
+
+    // If the page has no content or only a title, it is empty
+    return tree.children.length <= 1 && tree.children[0]?.type === 'heading';
```
SamyPesse: Maybe we can do something quicker here:

```ts
function isOnlyHeading(markdown: string): boolean {
  // Remove frontmatter if present
  const cleaned = markdown
    .trim()
    .replace(/^---\n[\s\S]*?\n---\n?/g, '')
    .trim();

  // Match a single heading line, possibly followed by empty lines
  return /^#{1,6} .+\n*$/i.test(cleaned);
}
```

(Generated from ChatGPT)

To avoid parsing a long markdown pages (takes CPU) when we just want to check this.

Maybe we can compose the 2 solutions into something efficient:

```ts
import { unified } from 'unified';
import remarkParse from 'remark-parse';
import { Root } from 'mdast';

/**
 * Returns true if the markdown only contains frontmatter and a single heading.
 */
export function isOnlyHeadingMarkdown(markdown: string): boolean {
  // Fast path: try to quickly detect obvious matches
  if (/^[ \t]*# .+$/m.test(markdown)) {
    const stripped = markdown
      .trim()
      .replace(/^---\n[\s\S]*?\n---\n?/g, '')
      .trim();

    // If there's a single heading line and nothing else (or empty lines)
    if (/^#{1,6} .+\s*$/.test(stripped) && !/\n\S+/g.test(stripped.split('\n').slice(1).join('\n'))) {
      return true;
    }
  }

  // Fallback: parse with remark for safety
  const tree = unified().use(remarkParse).parse(markdown) as Root;

  let seenHeading = false;

  for (const node of tree.children) {
    if (node.type === 'yaml') continue;

    if (node.type === 'heading') {
      if (seenHeading) return false;
      seenHeading = true;
      continue;
    }

    // Allow empty whitespace-only text nodes (e.g., extra newlines)
    if (node.type === 'paragraph' && node.children.length === 1 && node.children[0].type === 'text') {
      if (node.children[0].value.trim() === '') continue;
    }

    // Anything else is disallowed
    return false;
  }

  return seenHeading;
}
```

(Generated by ChatGPT as well).

SamyPesse (approved):

packages/gitbook/src/routes/markdownPage.ts
nolannbiron: I tweaked a bit the second one (always remove frontmatter, readability)

PR #3411, opened 2025-06-30
Error handling for AI chat

nolannbiron:
CleanShot 2025-07-01 at 00 58 02@2x

### DIFF
--- a/packages/gitbook/src/components/AI/useAIChat.tsx
+++ b/packages/gitbook/src/components/AI/useAIChat.tsx
@@ -11,10 +11,6 @@ import { useAIMessageContextRef } from './useAIMessageContext';
 export type AIChatMessage = {
     role: AIMessageRole;
     content: React.ReactNode;
-    /**
-     * When true, the message represents an error returned by the client-side logic.
-     */
-    error?: boolean;
 };
 
 export type AIChatState = {
--- a/packages/gitbook/src/components/AI/useAIChat.tsx
+++ b/packages/gitbook/src/components/AI/useAIChat.tsx
@@ -154,6 +154,7 @@ export function useAIChatController(): AIChatController {
                                     // Mark as not loading when the response is finished
                                     // Even if the stream might continue as we receive 'response_followup_suggestion'
                                     loading: false,
+                                    error: false,
                                 }));
                                 break;
                             }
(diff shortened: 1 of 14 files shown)


packages/gitbook/src/components/AI/useAIChat.tsx
SamyPesse, line 1: I don't think you should manage errors here, it's not the best place as it'll cover only a certain type of errors (between the server and the API), but not between the browser and the server.

Instead, we should handle errors on the client side when streaming (the server will emit any error, and the client will also emit error for disconnection, etc).

SamyPesse: Also your current approach per message means a lot of error messages can be displayed, instead of managing it as one global `error: boolean` on the chat

nolannbiron: I think we should have both a global error if the service is down, and a per-message error in case of individual failures
For example, if one prompt fails but the next one succeeds, I still want to keep a record that the first one failed in the chat history, you see what I mean?

Like this:
- Question
- failure message
- Question
- success

Instead of:
- Question
- Question
- success

SamyPesse: I think this is over-engineering the problem; managing multiple error states is complicated and not needed.

If the chat fails, the whole thing should be read-only with a message at the bottom (like we do for Docs Agents in the app, like LLM chats do, etc).

nolannbiron: So if you reload the page, the failed question isn’t saved and the global error disappears ?

SamyPesse: At the moment we don't even persist the state between reloads.<br><br>But considering that we do: no, it'll show the same thing as before:

* List of all messages
* Alert at the bottom saying that there was an error
* Text input disabled, maybe a hint to clear the conversation before starting again

nolannbiron: Ok so just to make sure, the only way to reset the error would be to clear the conversation ?

nolannbiron:
@SamyPesse I might be wrong, but to me the try/catch inside `useAIChat` makes sense because it wraps the calls inside `postMessage`. So if anything goes wrong during the `postMessage` flow (browser -> server action -> API), we display the error banner

Is this the behavior you were talking about ?

https://github.com/user-attachments/assets/d8238acc-9dd2-4660-a0e2-bd15e9924e55

zenoachtig (approved):
UI and copy looks basic but fine for now!

nolannbiron:
@zenoachtig I agree the current copy is quite basic, we can improve it quickly if you have any idea

nolannbiron:
About the UI/UX, I agree it's basic too, and it’s a chat, so I think we should match the standard AI chat error handling:

* per-message error/retry (we need to do it I think) 
* global error/retry (covered in this PR)

packages/gitbook/src/components/AIChat/AIChatInput.tsx
```diff
@@ -7,11 +7,15 @@ import { Tooltip } from '../primitives/Tooltip';
 
 export function AIChatInput(props: {
     value: string;
-    disabled: boolean;
+    disabled?: boolean;
+    /**
+     * When true, the input is disabled
+     */
+    isLoading: boolean;
```
SamyPesse: It could be just `loading` to match the property name in the chat state

PR #3416, opened 2025-07-01
Inline search modal

zenoachtig:


packages/gitbook/src/components/AIChat/AIChatButton.tsx
```diff
@@ -18,8 +19,8 @@ export function AIChatButton() {
             icon={<AIChatIcon />}
             iconOnly
             size="default"
-            variant="secondary"
-            className="!px-3 bg-tint-base py-2.5"
+            variant="header"
+            className={tcls('h-9 px-2.5', props.className)}
```
SamyPesse: I find it weird that we have to hard-code size on buttons

packages/gitbook/src/components/RootLayout/CustomizationRootLayout.tsx
```diff
@@ -100,6 +100,11 @@ export async function CustomizationRootLayout(props: {
                     <link rel="privacy-policy" href={customization.privacyPolicy.url} />
                 ) : null}
 
+                <meta
```
SamyPesse: We should not have to do this and instead use https://nextjs.org/docs/app/api-reference/functions/generate-viewport

You should add this to `generateSiteLayoutViewport`

packages/gitbook/src/components/Search/SearchContainer.tsx
```diff
@@ -0,0 +1,186 @@
...
+interface SearchContainerProps {
+    spaceTitle: string;
+    isMultiVariants: boolean;
+    withAsk: boolean;
+    withAIChat: boolean;
+    className?: string;
+}
+
+/**
+ * Client component to render the search modal when the url contains a search query.
```
SamyPesse: This comment should be revisited

packages/gitbook/src/components/Search/SearchContainer.tsx
```diff
@@ -0,0 +1,186 @@
...
+import { SearchInput } from './SearchInput';
+import { SearchResults, type SearchResultsRef } from './SearchResults';
+import { SearchScopeToggle } from './SearchScopeToggle';
+import { useSearch } from './useSearch';
+
+interface SearchContainerProps {
+    spaceTitle: string;
+    isMultiVariants: boolean;
+    withAsk: boolean;
+    withAIChat: boolean;
```
SamyPesse: Maybe now you can pass `aiMode: CustomizationAIMode`

packages/gitbook/src/components/Search/SearchContainer.tsx
```diff
@@ -0,0 +1,186 @@
...
+/**
+ * Client component to render the search modal when the url contains a search query.
+ */
+export function SearchContainer(props: SearchContainerProps) {
+    const { spaceTitle, isMultiVariants, withAsk, withAIChat, className } = props;
+
+    const [state, setSearchState] = useSearch();
+    const [open, setOpen] = useState(false);
+    const searchAsk = useSearchAskState();
+    // const [askState] = searchAsk;
```
SamyPesse:
```suggestion
```

packages/gitbook/src/components/Search/SearchContainer.tsx
```diff
@@ -0,0 +1,186 @@
...
+}
+
+/**
+ * Client component to render the search modal when the url contains a search query.
+ */
+export function SearchContainer(props: SearchContainerProps) {
+    const { spaceTitle, isMultiVariants, withAsk, withAIChat, className } = props;
+
+    const [state, setSearchState] = useSearch();
+    const [open, setOpen] = useState(false);
```
SamyPesse: Why do we need this state? since we have `useSearch` above which should manage this

packages/gitbook/src/components/Search/SearchContainer.tsx
```diff
@@ -0,0 +1,186 @@
...
+        }
+    };
+
+    React.useEffect(() => {
+        if (state === null) {
+            setOpen(false);
+        } else {
+            setOpen(true);
+        }
+    }, [state]);
```
SamyPesse: This smells bad

packages/gitbook/src/components/Search/SearchContainer.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+    return (
+        <SearchAskProvider value={searchAsk}>
+            <Popover
+                content={
+                    // Only show content if there's a query or Ask is enabled
+                    state?.query || withAsk ? (
+                        <React.Suspense fallback={null}>
+                            {isMultiVariants && !state?.ask ? (
+                                <SearchScopeToggle spaceTitle={spaceTitle} />
```
SamyPesse, line 128: We should look at filtering and tabs for sections in another PR :)

packages/gitbook/src/components/Search/SearchInput.tsx
```diff
@@ -0,0 +1,148 @@
...
+import React from 'react';
+import { useEffect, useRef, useState } from 'react';
+
+import { tString, useLanguage } from '@/intl/client';
+import { tcls } from '@/lib/tailwind';
+import { Icon } from '@gitbook/icons';
+import { Button, variantClasses } from '../primitives';
+import { useClassnames } from '../primitives/StyleProvider';
+
+interface SearchButtonProps {
```
SamyPesse:
```suggestion
interface SearchInputProps {
```

packages/gitbook/src/components/Search/SearchInput.tsx
```diff
@@ -0,0 +1,148 @@
...
+export const SearchInput = React.forwardRef<HTMLDivElement, SearchButtonProps>(
+    function SearchInput(props, ref) {
+        const { onChange, onKeyDown, onFocus, value, withAsk = false, isOpen, className } = props;
+        const inputRef = useRef<HTMLInputElement>(null);
+
+        const language = useLanguage();
+        const buttonStyles = useClassnames(['ButtonStyles']);
+
+        // Size classes for medium size button
+        const sizeClasses = ['text-sm', 'px-3.5', 'py-1.5', 'circular-corners:px-4'];
```
SamyPesse: Very minor, but no need to redefine this array at every render if it's constant

packages/gitbook/src/components/Search/SearchInput.tsx
```diff
@@ -0,0 +1,148 @@
...
+            }
+        }, [isOpen]);
+
+        return (
+            <div className="relative flex size-9 grow">
+                <div
+                    ref={ref}
+                    onClick={onFocus}
+                    onKeyDown={(e) => {
+                        if (e.key === 'Enter' || e.key === ' ') {
```
SamyPesse: Hum what?

packages/gitbook/src/components/Search/SearchInput.tsx
```diff
@@ -0,0 +1,148 @@
...
+                    {value ? (
+                        <Button
+                            variant="blank"
+                            label="Clear"
+                            size="medium"
+                            iconOnly
+                            icon="circle-xmark"
+                            className="-mx-1.5 animate-scaleIn px-1.5 theme-bold:text-header-link theme-bold:hover:bg-header-link/3"
+                            onClick={() => {
+                                onChange({
```
SamyPesse: This looks bad

Either you change the value of the input with `inputRef.current?.value = '';`

or you change the `onChange` to be `onChange: (value: string) => void`

packages/gitbook/src/components/Search/SearchInput.tsx
```diff
@@ -0,0 +1,148 @@
...
+                        <Icon icon="magnifying-glass" className="size-4 animate-scaleIn" />
+                    )}
+
+                    <input
+                        type="text"
+                        onFocus={onFocus}
+                        onKeyDown={onKeyDown}
+                        onChange={onChange}
+                        value={value ?? ''}
+                        size={1} // Determines the width of the input (in characters). It's inconsistent between browsers and limits the min-width, so we set it to a sensible minimum and control width ourselves.
```
SamyPesse: Why do we need it if we can just use min-width?

packages/gitbook/src/components/Search/SearchInput.tsx
```diff
@@ -0,0 +1,148 @@
...
+import { tcls } from '@/lib/tailwind';
+import { Icon } from '@gitbook/icons';
+import { Button, variantClasses } from '../primitives';
+import { useClassnames } from '../primitives/StyleProvider';
+
+interface SearchButtonProps {
+    onChange: (event: React.ChangeEvent<HTMLInputElement>) => void;
+    onKeyDown: (event: React.KeyboardEvent<HTMLInputElement>) => void;
+    onFocus: () => void;
+    value?: string;
```
SamyPesse: I think `value` should always be a string

packages/gitbook/src/components/Search/SearchInput.tsx
```diff
@@ -0,0 +1,148 @@
...
+            </div>
+        );
+    }
+);
+
+function Shortcut() {
+    const [operatingSystem, setOperatingSystem] = useState<string | null>(null);
+
+    useEffect(() => {
+        function getOperatingSystem() {
```
SamyPesse: Why do you have this function here? if it's global define it outside the component

Otherwise it's often redefined by the JS engine

packages/gitbook/src/components/Search/SearchPageResultItem.tsx
```diff
@@ -54,69 +41,139 @@ export const SearchPageResultItem = React.forwardRef(function SearchPageResultIt
...
-            >
-                <Icon
-                    icon={active ? 'arrow-turn-down-left' : 'chevron-right'}
-                    className={tcls('size-4')}
-                />
-            </div>
-        </Link>
+            </p>
+        </SearchResultItem>
+        // <Link
```
SamyPesse: ?

packages/gitbook/src/components/Search/SearchResults.tsx
```diff
@@ -78,8 +78,7 @@ export const SearchResults = React.forwardRef(function SearchResults(
 
             let cancelled = false;
 
-            // Silently fetch the recommended questions, instead of showing a spinner
-            setResultsState({ results: [], fetching: false });
+            // setResultsState({ results: [], fetching: true });
```
SamyPesse: ?

packages/gitbook/src/components/Search/SearchResults.tsx
```diff
@@ -320,3 +328,70 @@ function withQuestionResult(results: ResultType[], query: string): ResultType[]
 
     return [{ type: 'question', id: 'question', query }, ...(without ?? [])];
 }
+
+export const SearchResultItem = React.forwardRef(function SearchResultItem(
```
SamyPesse: It looks like you have a circular import. This should be in a different file if it's being imported by packages/gitbook/src/components/Search/SearchPageResultItem.tsx

packages/gitbook/src/components/Search/useSearch.ts
```diff
@@ -78,8 +78,8 @@ export function useSearchLink(): (query: Partial<SearchState>) => LinkProps {
         (query) => {
             const searchParams = new URLSearchParams();
             searchParams.set('q', query.query ?? '');
-            query.ask ? searchParams.set('ask', 'on') : searchParams.delete('ask');
-            query.global ? searchParams.set('global', 'on') : searchParams.delete('global');
```
SamyPesse, line 82: Why?

packages/gitbook/src/components/SpaceLayout/SpaceLayout.tsx
```diff
@@ -4,19 +4,18 @@ import React from 'react';
...
 import { TableOfContents } from '@/components/TableOfContents';
 import { CONTAINER_STYLE } from '@/components/layout';
-import { getSpaceLanguage } from '@/intl/server';
-import { t } from '@/intl/translate';
 import { tcls } from '@/lib/tailwind';
 
 import type { VisitorAuthClaims } from '@/lib/adaptive';
 import { GITBOOK_API_PUBLIC_URL, GITBOOK_APP_URL } from '@/lib/env';
 import { AIChat } from '../AIChat';
+import { AIChatButton } from '../AIChat/AIChatButton';
```
SamyPesse: It'd be great to avoid these types of imports and instead use an `index.ts`

packages/gitbook/src/components/SpaceLayout/SpaceLayout.tsx
```diff
@@ -52,6 +51,20 @@ export function SpaceLayout(props: {
 
     const withAIChat = context.customization.aiSearch.enabled && context.site.id === 'site_p4Xo4';
 
+    const searchAndAI = (
+        <div className="flex grow items-center gap-2">
+            <React.Suspense fallback={null}>
+                <SearchContainer
+                    withAsk={customization.aiSearch.enabled}
```
SamyPesse: When updating from `main`, this needs to be updated to the properr condition

packages/gitbook/src/components/TableOfContents/TableOfContents.tsx
```diff
@@ -67,7 +67,7 @@ export async function TableOfContents(props: {
                     'lg:flex',
                     'page-no-toc:lg:hidden',
                     'page-no-toc:xl:flex',
-                    'site-header-none:page-no-toc:lg:flex',
+                    '[body:has(.page-no-toc):not(:has(#site-header:not(.mobile-only)))_&]:lg:flex',
```
SamyPesse: Hum what? can we define this case a bit better?

packages/gitbook/src/components/primitives/Button.tsx
```diff
@@ -1,6 +1,7 @@
 'use client';
 
 import type { ButtonHTMLAttributes, HTMLAttributeAnchorTarget, HTMLAttributes } from 'react';
+import { forwardRef } from 'react';
```
SamyPesse: You have 2 import from `react`, let's do a clean `import * as React from 'react';`

packages/gitbook/src/components/primitives/styles.ts
```diff
@@ -13,21 +13,25 @@ export const ButtonStyles = [
...
     'contrast-more:hover:ring-2',
     'contrast-more:hover:ring-tint-12',
 
-    'hover:scale-104',
-    'depth-flat:hover:scale-100',
-    'active:scale-100',
+    'depth-subtle:hover:-translate-y-px',
+    'depth-subtle:focus-visible:-translate-y-px',
+    // 'depth-flat:hover:scale-100',
+    // 'active:scale-100',
```
SamyPesse: ?

packages/gitbook/src/components/Search/SearchContainer.tsx
zenoachtig: Yeah you're right, fixed this by deriving `open` from the state instead of having to manually set state based on it 😳

packages/gitbook/src/components/Search/SearchInput.tsx
zenoachtig: Agreed. It's a hack anyway because I'm visually replicating the div to look like a button. Making it a constant outside the render for now, but we need to look at buttons in general.

packages/gitbook/src/components/AIChat/AIChatButton.tsx
zenoachtig: Normally buttons should size themselves to their contents, but in these cases we want to guarantee the same height between the search button and chat button, even if their content changes (especially the search input has a few elements that are shown/hidden conditionally).

As said I want to take a look at buttons sizing and stuff soon

packages/gitbook/src/components/Search/SearchInput.tsx
zenoachtig: This div needs an onClick to show the input on mobile, where it's normally hidden.

Normally you'd also need to add a keyboard trigger to do the same without a pointer, so Biome complained and I added the `onKeyDown` prop.

But in this case the input already be focused on its own, so the keyboard trigger is really not needed. Removing this `onKeyDown` and adding a `biome-ignore` instead.

packages/gitbook/src/components/Search/SearchInput.tsx
zenoachtig: Good one, not my code but changed anyway

packages/gitbook/src/components/Search/SearchInput.tsx
zenoachtig: If we don't have it, some browsers (Safari…) will use the default `size=20` and it overrides `min-width`. 

Relevant stack overflow (adding to the comment): https://stackoverflow.com/a/29990524

packages/gitbook/src/components/Search/useSearch.ts
zenoachtig: We use `true` in other places (particularly `SearchAskAnswer.tsx`), and had to rewrite `on` to `true`, which I couldn't get to work with the new components I used. I might be doing something wrong, but this fixed it and seems more consistent anyway… happy to change though

packages/gitbook/src/components/TableOfContents/TableOfContents.tsx
zenoachtig: Yeah we should, but I'll leave this change out of this PR as it's not directly related. I want to improve the header/sidebar display logic in a follow-up.

What I was trying to do here: it's a bit of a long story but because of the way Tailwind's custom variants work we can't actually chain `site-header-none:` with `page-no-toc:` like this, because the generated CSS selector doesn't actually select the proper elements. 

So the original rule doesn't actually work here. Instead I'm chaining the two selectors in a custom selector, but it's very unreadable. I'll need to rethink that system, but there's a ticket in Linear for it already so I'll do it after this PR.

packages/gitbook/src/components/Search/SearchInput.tsx
zenoachtig: Actually I did something else that can leave this out

packages/gitbook/src/components/Search/SearchContainer.tsx
```diff
@@ -0,0 +1,180 @@
...
+            query: value,
+            global: prev?.global ?? false,
+            isOpen: true,
+        }));
+    };
+
+    // We trim the query to avoid invalidating the search when the user is typing between words.
+    const normalizedQuery = state?.query.trim() ?? '';
+
+    const isMobile = typeof window !== 'undefined' ? window.innerWidth < 640 : false;
```
SamyPesse: This will cause hydration issue, it should be a state and a useEffect. You can create a `useIsMobile` hooks in `@/components/hooks`:

```ts
export function useIsMobile(breakpoint: number = 640): boolean {
  const [isMobile, setIsMobile] = useState(false);

  useEffect(() => {
    // Use matchMedia for a single, efficient listener.
    // 0.02px fudge keeps it safely below the breakpoint (mirrors Tailwind’s 639.98px).
    const query = `(max-width: ${breakpoint - 0.02}px)`;
    const media = window.matchMedia(query);

    // Sync initial state if it differs (e.g., on hydration).
    setIsMobile(media.matches);

    const handleChange = (event: MediaQueryListEvent | MediaQueryList) => {
      setIsMobile(event.matches);
    }

    media.addEventListener('change', handleChange);

    return () => {
      media.removeEventListener('change', handleChange);
    };
  }, [breakpoint]);

  return isMobile;
}
```

packages/gitbook/src/components/Search/useSearch.ts
```diff
@@ -28,13 +29,24 @@ export function useSearch(): [SearchState | null, UpdateSearchState] {
         history: 'replace',
     });
 
+    // Separate local state for isOpen (not synchronized with URL)
+    // Default to true if there's already a query in the URL
```
SamyPesse: Why do you need a separate state?

packages/gitbook/src/components/Search/useSearch.ts
SamyPesse: Let's change this in case some used this URL; let's avoid useless breaking chnages

packages/gitbook/src/components/SpaceLayout/SpaceLayout.tsx
```diff
@@ -51,9 +50,26 @@ export function SpaceLayout(props: {
         customization.footer.groups?.length;
 
     // TODO: remove aiSearch and optional chain once the cache has been fully updated (after 11/07/2025)
-    const withAIChat =
-        context.customization.ai?.mode === CustomizationAIMode.Assistant &&
-        (context.site.id === 'site_p4Xo4' || context.site.id === 'site_JOVzv');
+    const aiMode =
+        context.customization.ai?.mode ??
+        (context.site.id === 'site_p4Xo4' || context.site.id === 'site_JOVzv'
```
SamyPesse: We can remove this condition now

SamyPesse (commented):
Not approving yet, we need to QA even more as it's a critical change

packages/gitbook/src/components/Search/useSearch.ts
zenoachtig: Because having this `open` prop in the URL looks messy and feels unnecessary have in there all the time. It's a very transient state that shouldn't be explicit, I believe at least.

Since we're using nuqs all of the state in `rawState` will be synced to query parameters. There _is_ a way to have state be managed by nuqs but exclude the query param, but it's hacky: you have to set an infinite throttle timeout on the `open` parameter so that it's being delayed forever.

```
open: parseAsBoolean.withOptions({ throttleMs: Number.POSITIVE_INFINITY })
```

This is why it feels cleaner to me to manage this state separately as normal React state.

packages/gitbook/src/components/Search/useSearch.ts
zenoachtig: Existing URLs also use `true`, the `on` prop was never actually used.
https://run-ai-docs.nvidia.com/?q=What+permissions+does+the+Template+administrator+have?&ask=true

Before this patch, each search result was a `<Link>` with the `href` generated by `useSearchLink()`, which included the `&ask=on` param. You can see it by right-clicking a search result and copying its URL. It will have the `&ask=on` included. But if you enter that URL in a new tab you'll see it won't actually launch Ask — it strips it because it's not parsed as a boolean and shows the default results list instead, so you have to click "ask" again yourself.

The reason it normally worked is because each result _also_ had an `onClick` handler that intercepted the `Link`'s navigation and triggered [`onSwitchToAsk()`](https://github.com/GitbookIO/gitbook/pull/3416/files/4e5590a4b695415f326eaa4984c1f4ed7e5b72fd#diff-4ce98836a3b5445401d773610cee7366e211cdce978528cf1a6def5491596787L190), which triggered a `setSearchState` that included `ask: true`, which propagated to the URL via nuqs as `ask=true`.

I've removed that `setSearchState` and perform a navigation, same behaviour as other search results. But because I'm now navigating to the faulty `ask=on` URL, it didn't work anymore and I have to update these two lines to set the URL params correctly.

If you think it's a mistake to perform a navigation and we should do a state update instead I can re-introduce that `onClick`/`setSearchState`, but either way the `on` doesn't actually work here and should be switched to `true`.

SamyPesse (approved):

PR #3423, opened 2025-07-02
Adjust cache life for error responses

conico974:
Modify caching behavior to differentiate between 404 errors and other error codes, allowing longer cache durations for 404 errors while reducing cache time for other errors.

I'm still a bit worried about the API load that this will create, we could do it on a per function basis instead of like that all at once

### DIFF
--- a/packages/gitbook/src/lib/data/errors.ts
+++ b/packages/gitbook/src/lib/data/errors.ts
@@ -2,6 +2,8 @@ import { GitBookAPIError } from '@gitbook/api';
 import { unstable_cacheLife as cacheLife } from 'next/cache';
 import type { DataFetcherErrorData, DataFetcherResponse } from './types';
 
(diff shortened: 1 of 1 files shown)


packages/gitbook/src/lib/data/errors.ts
```diff
@@ -134,6 +146,39 @@ export function ignoreDataFetcherErrors<T>(
...
+        if (!error.response) {
+            return undefined;
+        }
+
+        const cacheControl = error.response.headers.get('cache-control');
+        if (!cacheControl) {
+            return undefined;
+        }
+
+        const maxAgeMatch = cacheControl.match(/max-age=(\d+)/i);
```
SamyPesse: We already have `parse-cache-control` as a package, we should use this

SamyPesse (approved):
Looks good!

PR #3444, opened 2025-07-07
Styling improvements to AI Chat Followup questions

zenoachtig:
- Follow up questions are now shown at the bottom of the message view instead of fixed above the message view.
- Restyled follow up questions
- Rounded the message bubbles more on the `circular` corner style.

Screenshot 2025-07-07 at 16 04 18

### DIFF
--- a/packages/gitbook/src/components/AIChat/AIChatMessages.tsx
+++ b/packages/gitbook/src/components/AIChat/AIChatMessages.tsx
@@ -1,17 +1,20 @@
 import { tcls } from '@/lib/tailwind';
 import { AIMessageRole } from '@gitbook/api';
 import type React from 'react';
-import type { AIChatState } from '../AI/useAIChat';
+import type { AIChatController, AIChatState } from '../AI/useAIChat';
+import { AIChatFollowupSuggestions } from './AiChatFollowupSuggestions';
 
 export function AIChatMessages(props: {
     chat: AIChatState;
+    chatController: AIChatController;
     lastUserMessageRef?: React.RefObject<HTMLDivElement>;
 }) {
-    const { chat, lastUserMessageRef } = props;
+    const { chat, chatController, lastUserMessageRef } = props;
 
     return (
         <>
             {chat.messages.map((message, index) => {
+                const isLastMessage = index === chat.messages.length - 1;
                 const isLastUserMessage =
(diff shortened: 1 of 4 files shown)


packages/gitbook/src/components/AIChat/AiChatFollowupSuggestions.tsx
```diff
@@ -15,17 +15,17 @@ export function AIChatFollowupSuggestions(props: {
...
                 <Button
                     key={index}
                     onClick={() => {
                         chatController.postMessage({ message: suggestion });
                     }}
                     label={suggestion}
-                    className="max-w-full animate-[present_500ms_both] whitespace-normal text-right"
-                    size="small"
-                    variant="secondary"
+                    className="max-w-[80%] animate-[present_500ms_both] whitespace-normal text-left ring-1 ring-tint-subtle"
```
nolannbiron, line 26: Is there a reason we’re using ring instead of border? ring is meant for outlines, but we’re using it as a border everywhere

packages/gitbook/src/components/AIChat/AIChatMessages.tsx
```diff
@@ -48,6 +52,12 @@ export function AIChatMessages(props: {
                                 ))}
                             </div>
                         ) : null}
+                        {isLastMessage ? (
+                            <AIChatFollowupSuggestions
+                                chat={chat}
+                                chatController={chatController}
+                            />
+                        ) : null}
```
nolannbiron, line 60: We should also check that chat.error is false before rendering the suggestions, otherwise the alert and suggestions overlap

zenoachtig: The message view resizes to never overlap with the input container, it's dynamically sized. So we can render them together if we want

packages/gitbook/src/components/AIChat/AiChatFollowupSuggestions.tsx
zenoachtig: Ring doesn't add width/height (sometimes useful) and also overlaps shadows. Sometimes the borders look too close to the shadow and it can look smeared out. But I agree, we should probably look into refactoring / standardising eventually

PR #3445, opened 2025-07-07
Always open header logo link in the current tab

gregberge:


packages/gitbook/src/components/primitives/Link.tsx
```diff
@@ -89,7 +89,7 @@ export const Link = React.forwardRef(function Link(
                 {...domProps}
                 href={href}
                 onClick={onClick}
-                {...(externalLinksTarget === SiteExternalLinksTarget.Blank
+                {...(externalLinksTarget === SiteExternalLinksTarget.Blank && !domProps.target
```
SamyPesse: You also need to manage this in the `onClick`

packages/gitbook/src/components/Header/HeaderLogo.tsx
```diff
@@ -21,6 +21,7 @@ export async function HeaderLogo(props: HeaderLogoProps) {
     return (
         <Link
             href={linker.toAbsoluteURL(linker.toPathInSite(''))}
+            target="_self"
```
SamyPesse: I don't understand why there is a case where it was not opening in the same tab.

packages/gitbook/src/components/primitives/Link.tsx
gregberge: I fixed the whole logic in the link.

packages/gitbook/src/components/Header/HeaderLogo.tsx
gregberge: It's live on https://docs.ionos.com/cloud/, I don't know the underlying reason.

gregberge: That's because the link on the header logo is an absolute one.

packages/gitbook/src/components/Header/HeaderLogo.tsx
```diff
@@ -21,6 +21,9 @@ export async function HeaderLogo(props: HeaderLogoProps) {
     return (
         <Link
             href={linker.toAbsoluteURL(linker.toPathInSite(''))}
+            // This link should always open in the same tab
+            // so we set the target to _self.
+            target="_self"
```
SamyPesse:
```suggestion
href={linker.toPathInSite('')}
```

packages/gitbook/src/components/primitives/Link.tsx
```diff
@@ -52,46 +72,50 @@ export const Link = React.forwardRef(function Link(
...
     const forwardedClassNames = useClassnames(classNames || []);
-
-    // Use a real anchor tag for external links,s and a Next.js Link for internal links.
-    // If we use a NextLink for external links, Nextjs won't rerender the top-level layouts.
-    const isExternal = URL.canParse ? URL.canParse(props.href) : props.href.startsWith('http');
+    const isExternal = isExternalLink(href);
+    const { target, rel } = getTargetProps(props, { externalLinksTarget, isExternal });
 
     const onClick = (event: React.MouseEvent<HTMLAnchorElement>) => {
+        const isExternal = isExternalLink(href, window.location.origin);
```
SamyPesse: We should avoid re-using same var name in a scope

packages/gitbook/src/components/Header/HeaderLogo.tsx
SamyPesse: I don't think we should hack it with `target="_self"` here; we should have a proper relative link and be fine with it

SamyPesse (commented):
GLobally it looks like it refactors the code in a good way; but I think we should remove the `target="_self"` from HeaderLogo

PR #3454, opened 2025-07-08
Fix site redirect resolution with length check for source paths

conico974:
Implement a length check for source paths to ensure proper resolution of site redirects, preventing potential errors with overly long paths.

### DIFF
--- a/packages/gitbook/src/components/SitePage/fetch.ts
+++ b/packages/gitbook/src/components/SitePage/fetch.ts
@@ -71,14 +71,17 @@ async function resolvePage(context: GitBookSiteContext, params: PagePathParams |
                 redirectPathname,
             ]);
             for (const source of redirectSources) {
-                const resolvedSiteRedirect = await getDataOrNull(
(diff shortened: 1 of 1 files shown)


packages/gitbook/src/components/SitePage/fetch.ts
```diff
@@ -71,14 +71,17 @@ async function resolvePage(context: GitBookSiteContext, params: PagePathParams |
...
-                    context.dataFetcher.getSiteRedirectBySource({
-                        organizationId,
-                        siteId: site.id,
-                        source,
-                        siteShareKey: shareKey,
-                    })
-                );
+                // We try to resolve the site redirect
+                const resolvedSiteRedirect =
+                    source.length < 512 &&
```
gregberge, line 76:
```suggestion
                    source.length <= 512 &&
```

conico974: It's actually on purpose, the `rawPathname` used to fail exactly because of an `=`

PR #3455, opened 2025-07-08
Gracefully handle OpenAPI fetch errors

conico974:
Implement error handling for OpenAPI fetching.

### DIFF
--- a/packages/gitbook/src/lib/openapi/fetch.ts
+++ b/packages/gitbook/src/lib/openapi/fetch.ts
@@ -53,15 +53,29 @@ const fetchFilesystem = async (url: string) => {
     try {
         return await fetchFilesystemUncached(url);
     } catch (error) {
+        // To avoid hammering the file with requests, we cache the error for around a minute.
+        cacheLife('minutes');
         // Throwing an error inside a "use cache" function obfuscates the error,
         // so we need to handle it here and recreates the error outside the cache function.
         if (error instanceof OpenAPIParseError) {
-            cacheLife('seconds');
             return { error: { code: error.code, message: error.message } };
         }
-        throw error;
+        if (error instanceof OpenAPIFetchError) {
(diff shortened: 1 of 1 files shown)


packages/gitbook/src/lib/openapi/fetch.ts
```diff
@@ -53,16 +53,30 @@ const fetchFilesystem = async (url: string) => {
...
+        cacheLife('minutes');
         // Throwing an error inside a "use cache" function obfuscates the error,
         // so we need to handle it here and recreates the error outside the cache function.
         if (error instanceof OpenAPIParseError) {
-            cacheLife('seconds');
             return { error: { code: error.code, message: error.message } };
         }
-        throw error;
+        if (error instanceof OpenAPIFetchError) {
+            return { error: { code: 'invalid' as const, message: 'Failed to fetch OpenAPI file' } };
```
SamyPesse, line 79: Should you type this somewhere?

Is it the same as the API error stored in the cache?

If yes, the `code` should be a HTTP status code:


```suggestion
            return { error: { code: 500, message: 'Failed to fetch OpenAPI file' } };
```

packages/gitbook/src/lib/openapi/fetch.ts
```diff
@@ -78,7 +92,9 @@ async function fetchFilesystemUncached(
     });
 
     if (!response.ok) {
-        throw new Error(`Failed to fetch OpenAPI file: ${response.status} ${response.statusText}`);
+        throw new OpenAPIFetchError(
```
SamyPesse: Why not just use `DataFetcherError` and pass the status code?

conico974: Yeah i could have used the DataFetcherError, i'll update it to use that

packages/gitbook/src/lib/openapi/fetch.ts
conico974: The idea was to reuse the existing error handling on parse error.

PR #3459, opened 2025-07-09
Move AI Actions markdown fetching to client-side

nolannbiron:


packages/gitbook/src/components/AIActions/AIActionsDropdown.tsx
```diff
@@ -25,10 +25,33 @@ export function AIActionsDropdown(props: {
     trademark: boolean;
 }) {
     const ref = useRef<HTMLDivElement>(null);
+    const { markdownPageUrl } = props;
+
+    const [markdown, setMarkdown] = useState<string>();
+
+    // Fetch the markdown page when the URL changes
+    useEffect(() => {
```
SamyPesse: You want to fetch the markdown everytime we render the page?

Maybe less than 0.1% of users will click this button; I'm not sure it's worth doing 100m requests per month to generate the markdown for these 100k people.

The logic to fetch the markdown should be handled in the `onClick`

and potentially in the `onMouseEnter` by fetching the value into a ref

ex:

```
const markdownRef = React.useRef<Promise<string> | null>(null);

const fetchMarkdown = async () => {
  if (!markdownRef.current) {
    markdownRef.current = fetch(url).then(res => res.text());
  }
  return markdownRef.current;
}

const onMouseEnter = () => {
  fetchMarkdown(); // Start prefetching as early as possible
}

const onClick = async () => {
  const markdown = await fetchMarkdown();
}

```

nolannbiron: @SamyPesse Updated, a bit more complex, but it covers everything: prefetch on mouse enter, loading state if clicked directly, and cancellation of the prefetch if it’s still running

packages/gitbook/src/components/AIActions/AIActions.tsx
```diff
@@ -109,7 +111,8 @@ export function CopyMarkdown(props: {
             clearTimeout(timeoutRef.current);
         }
 
-        navigator.clipboard.writeText(markdown);
+        // Copy the markdown to the clipboard
+        copyMarkdown();
 
         setCopied(true);
```
gregberge: You said it's copied but it's that's async. Your state is not clear, "copied", "isLoading", etc..

packages/gitbook/src/components/AIActions/assets/useCopyMarkdown.tsx
```diff
@@ -0,0 +1,78 @@
+import { useRef, useState } from 'react';
+
+const markdownCache = new Map<string, string>();
+
+/**
+ * Hook to fetch and copy the markdown of a page.
+ * Result is cached to avoid fetching the same page multiple times.
+ */
+export function useCopyMarkdown(markdownPageUrl: string) {
```
gregberge: This should not be a hook, you can just expose a `fetchMarkdown` and handle the loading in the parent component.

packages/gitbook/src/components/AIActions/assets/useCopyMarkdown.tsx
gregberge, line 1: The logic is not good here. If you want to have a cache you should store the promise in it. So when the user click multiple times you handle it. Currently you will spawn lot of requests (abort does nothing at network level).

nolannbiron: @gregberge Even without storing the promise if the user clicks multiple times it doesn’t re-fetch once the result has been cached. And the initial fetch (triggered on hover) is canceled if the user clicks before it completes

https://github.com/user-attachments/assets/5164d75a-631d-4c48-a820-c5532f22ed65

packages/gitbook/src/components/AIActions/AIActions.tsx
nolannbiron: yes forgot the await, but why isn't it clear ? I can replace "isLoading" by "loading" since we don't really use the isXXXX for boolean but I think it's clear that copied is when the markdown is saved in clipboard and loading is when we're fetching the markdown

packages/gitbook/src/components/AIActions/assets/useCopyMarkdown.tsx
nolannbiron: Using a hook lets us separate the logic from the componen, it’s a large component with a lot to handle, so splitting some of the logic makes it clearer I think

packages/gitbook/src/components/AIActions/AIActions.tsx
gregberge: No I meant it was weird because you had loading: true and copied: true at the same time.

packages/gitbook/src/components/AIActions/assets/useCopyMarkdown.tsx
gregberge: So if you click very quickly or if we experience latency you throw potentially 50 requests.

nolannbiron: The button is disabled while loading, so you can’t trigger a new request.

The flow is: 
- onMouseEnter: triggers a prefetch
- Click: cancels the prefetch or uses the cached result if it resolved before the click
- The button is disabled during the fetch triggered by the click (if no cache) due to the loading state

tl;dr: If the data was prefetched, multiple clicks just use the cached result and no request is sent. If it wasn’t prefetched, the loading state prevents additional requests during the fetch

nolannbiron: I can't find a way to trigger 50 requests during testing 😅

nolannbiron: @gregberge I’ve completely simplified the flow, removed the prefetch since in 99% of cases, a click follows the hover, making it useless. 

Now it fetches on click, with a loading indicator and a disabled state to prevent multiple clicks

gregberge (approved):

SamyPesse (approved):
Looks good!

PR #3460, opened 2025-07-09
Reposition AI Actions dropdown

zenoachtig:
- Introduce `xsmall` button size
- Introduce `xs` screen breakpoint
- AI actions dropdown floats right of title/breadcrumbs, and one a separate line on mobile.
- Also passes `trademark` so AI Assistant action correctly uses the (un)branded version

### DIFF
--- a/packages/gitbook/src/components/AIActions/AIActions.tsx
+++ b/packages/gitbook/src/components/AIActions/AIActions.tsx
@@ -5,6 +5,7 @@ import { useAIChatState } from '@/components/AI/useAIChat';
 import { ChatGPTIcon } from '@/components/AIActions/assets/ChatGPTIcon';
 import { ClaudeIcon } from '@/components/AIActions/assets/ClaudeIcon';
 import { MarkdownIcon } from '@/components/AIActions/assets/MarkdownIcon';
+import { getAIChatName } from '@/components/AIChat';
 import AIChatIcon from '@/components/AIChat/AIChatIcon';
 import { Button } from '@/components/primitives/Button';
 import { DropdownMenuItem } from '@/components/primitives/DropdownMenu';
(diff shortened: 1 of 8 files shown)


packages/gitbook/src/components/AIActions/AIActions.tsx
```diff
@@ -21,22 +21,32 @@ type AIActionType = 'button' | 'dropdown-menu-item';
...
-            label={tString(language, 'ai_chat_ask', tString(language, 'ai_chat_assistant_name'))}
+            icon={
+                <AIChatIcon state={chat.loading ? 'thinking' : 'default'} trademark={trademark} />
+            }
+            label={tString(
+                language,
+                'ai_chat_ask',
+                trademark
+                    ? tString(language, 'ai_chat_assistant_name')
+                    : tString(language, 'ai_chat_assistant_name_unbranded')
```
nolannbiron: We can probably move this to a `getAIChatName` function

packages/gitbook/tailwind.config.ts
```diff
@@ -474,6 +474,7 @@ const config: Config = {
         },
         opacity: opacity(),
         screens: {
+            xs: '480px',
```
nolannbiron, line 477: Is there a specific use case for this ? I think 0-640px is already good to trigger any mobile

zenoachtig: I'm using it to trigger the float-right-of-title behaviour. In my experience 640px is very "early" to be switching to a full mobile view and some things look comically oversized if you make them single-column etc.

In fact, while we don't have this `xs` breakpoint officially, GBO has custom breakpoints sprinkled throughout, like for example in [the Cards layout](https://github.com/GitbookIO/gitbook/blob/9cc5a787e95a2e126ffd887ce72d8300b74c471e/packages/gitbook/src/components/DocumentView/Table/ViewCards.tsx#L18): `'min-[432px]:grid-cols-2'`.

So we can start using this breakpoint in place of those one-offs.

PR #3465, opened 2025-07-11
Improvement for CopyMarkdown action

nolannbiron:
This pull request introduces improvements to the CopyMarkdown action in the GitBook project. The main goal of these enhancements is to optimize the functionality, improve performance, and ensure better usability for end-users.

### DIFF
--- a/packages/gitbook/src/components/AIActions/AIActions.tsx
+++ b/packages/gitbook/src/components/AIActions/AIActions.tsx
@@ -61,7 +61,7 @@ type CopiedStore = {
 // We need to store everything in a store to share the state between every instance of the component.
 const useCopiedStore = create<
     CopiedStore & {
-        setState: (partial: Partial<CopiedStore>) => void;
+        setLoading: (loading: boolean) => void;
         copyWithTimeout: (props: { markdown: string }, opts?: { onSuccess?: () => void }) => void;
     }
 >((set) => {
@@ -70,7 +70,7 @@ const useCopiedStore = create<
     return {
         copied: false,
         loading: false,
-        setState: (partial: Partial<CopiedStore>) => set((state) => ({ ...state, ...partial })),
+        setLoading: (loading: boolean) => set({ loading }),
         copyWithTimeout: async (props, opts) => {
             const { markdown } = props;
             const { onSuccess } = opts || {};
@@ -112,16 +112,16 @@ export function CopyMarkdown(props: {
 
     const closeDropdown = useDropdownMenuClose();
 
-    const { copied, loading, setState, copyWithTimeout } = useCopiedStore();
+    const { copied, loading, setLoading, copyWithTimeout } = useCopiedStore();
 
     // Fetch the markdown from the page
     const fetchMarkdown = async () => {
-        setState({ loading: true });
+        setLoading(true);
 
         const result = await fetch(markdownPageUrl).then((res) => res.text());
         markdownCache.set(markdownPageUrl, result);
 
-        setState({ loading: false });
+        setLoading(false);
 
         return result;
     };
--- a/packages/gitbook/src/components/AIActions/AIActions.tsx
+++ b/packages/gitbook/src/components/AIActions/AIActions.tsx
@@ -8,7 +8,7 @@ import { MarkdownIcon } from '@/components/AIActions/assets/MarkdownIcon';
 import { getAIChatName } from '@/components/AIChat';
 import { AIChatIcon } from '@/components/AIChat';
 import { Button } from '@/components/primitives/Button';
-import { DropdownMenuItem } from '@/components/primitives/DropdownMenu';
+import { DropdownMenuItem, useDropdownMenuClose } from '@/components/primitives/DropdownMenu';
 import { tString, useLanguage } from '@/intl/client';
 import type { TranslationLanguage } from '@/intl/translations';
 import { Icon, type IconName, IconStyle } from '@gitbook/icons';
@@ -62,7 +62,7 @@ type CopiedStore = {
 const useCopiedStore = create<
     CopiedStore & {
         setState: (partial: Partial<CopiedStore>) => void;
-        copyWithTimeout: (props: { markdown: string; shouldCloseDropdown: boolean }) => void;
+        copyWithTimeout: (props: { markdown: string }, opts?: { onSuccess?: () => void }) => void;
     }
 >((set) => {
     let timeoutRef: ReturnType<typeof setTimeout> | null = null;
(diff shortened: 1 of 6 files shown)


packages/gitbook/src/components/AIActions/AIActions.tsx
```diff
@@ -53,19 +54,47 @@ export function OpenDocsAssistant(props: { type: AIActionType; trademark: boolea
...
+        copyWithTimeout: async (markdown: string) => {
+            // Clear any existing timeout
+            if (timeoutRef) {
+                clearTimeout(timeoutRef);
+            }
+
+            await navigator.clipboard.writeText(markdown);
+
+            // Set copied to true
+            set({ copied: true, markdown });
```
gregberge: So that's not partial, we remove loading?

packages/gitbook/src/components/AIActions/AIActions.tsx
```diff
@@ -53,19 +54,47 @@ export function OpenDocsAssistant(props: { type: AIActionType; trademark: boolea
...
+
+            // Set timeout to reset copied state
+            timeoutRef = setTimeout(() => {
+                set({ copied: false });
+                timeoutRef = null;
+            }, 1500);
```
gregberge: So you call a `set` outside the rendering of a component, is it OK?

packages/gitbook/src/components/AIActions/AIActions.tsx
```diff
@@ -77,61 +106,50 @@ export function CopyMarkdown(props: {
...
 
-        const markdown = await fetchMarkdown();
+        // Copy the markdown to the clipboard
+        copyWithTimeout(markdown || (await fetchMarkdown()));
 
-        navigator.clipboard.writeText(markdown);
-        setCopied(true);
+        // Close dropdown after a short delay to allow the copied state to show
+        if (type === 'dropdown-menu-item' && !isDefaultAction) {
+            setTimeout(() => {
```
gregberge: What is this hack... And the timeout isn't even shared.

gregberge (approved):
I will approve but this code is quite messy to be honest.

packages/gitbook/src/components/AIActions/AIActions.tsx
nolannbiron: That's Zustand's internal set so yeah I think it's ok

packages/gitbook/src/components/AIActions/AIActions.tsx
nolannbiron: that's different, the `setState` is partial and used to update from the component, `set` is zustand's internal setter

nolannbiron:
@gregberge I've updated the timeout, I know it's quite messy but I need to share the whole state between multiple instances (default action button + dropdown-menu item), so the goal of this PR is to fetch the markdown only once if you click multiple times on the copy button, with better loading state for default action and dropdown menu, etc

packages/gitbook/src/components/AIActions/AIActions.tsx
```diff
@@ -77,61 +121,36 @@ export function CopyMarkdown(props: {
...
-        // Dispatch on `document` so that the event is captured by Radix's
-        // dismissable-layer listener regardless of focus location.
-        document.dispatchEvent(new KeyboardEvent('keydown', { key: 'Escape', bubbles: true }));
-    };
+    // Clear cached markdown when navigating to a new page
+    useEffect(() => {
+        if (basePathname && pathname !== basePathname) {
+            setState({ markdown: null, pathname: basePathname });
+        }
+    }, [basePathname, pathname, setState]);
```
SamyPesse: This is not a good pattern; the better approach for this is to have your state be a map (page -> state).

It's better to think of state in a functional programming way and have purity; instead of imperative calls to sync things

packages/gitbook/src/components/AIActions/AIActions.tsx
```diff
@@ -53,19 +54,62 @@ export function OpenDocsAssistant(props: { type: AIActionType; trademark: boolea
...
+/**
+ * Function to manually close the dropdown
+ */
+function closeDropdown() {
+    const dropdownMenu = document.querySelector('div[data-radix-popper-content-wrapper]');
+    if (!dropdownMenu) return;
+
+    // Dispatch on `document` so that the event is captured by Radix's
+    // dismissable-layer listener regardless of focus location.
+    document.dispatchEvent(new KeyboardEvent('keydown', { key: 'Escape', bubbles: true }));
```
SamyPesse: is there no other way? this looks terrible

packages/gitbook/src/components/AIActions/AIActions.tsx
```diff
@@ -53,19 +54,62 @@ export function OpenDocsAssistant(props: { type: AIActionType; trademark: boolea
...
+    }
+>((set) => {
+    let timeoutRef: ReturnType<typeof setTimeout> | null = null;
+
+    return {
+        markdown: null,
+        copied: false,
+        loading: false,
+        pathname: '',
+        setState: (partial: Partial<CopiedStore>) => set((state) => ({ ...state, ...partial })),
```
SamyPesse: If you properly design your state, this one is not needed

packages/gitbook/src/components/AIActions/AIActions.tsx
nolannbiron: I did that first, but if you navigate to multiple pages, we need to clean the Map at some point, no ?

SamyPesse: We don't really have to, the user is not going to navigate between 10k pages in the site. But if you are concerned about this, you can use a LRU map (like [https://www.npmjs.com/package/quick-lru](https://www.npmjs.com/package/quick-lru))

nolannbiron:
@SamyPesse @gregberge I’ve updated the code to use an LRU cache for markdown content (max 10 paths), simplified `useCopiedStore` to only store `copied`, `loading` and the `copy` function (which is what we share between instances), wrapped `DropdownMenu` in a context, and created a `useDropdownMenuClose` hook (similar to what we use for Dialogs on GBX)

packages/gitbook/src/components/primitives/DropdownMenu.tsx
```diff
@@ -193,3 +202,12 @@ export function DropdownSubMenu(props: { children: React.ReactNode; label: React
...
     );
 }
+
+/**
+ * Hook to close the dropdown menu.
+ */
+export function useDropdownMenuClose() {
+    const context = useContext(DropdownMenuContext);
+    assert(context, 'DropdownMenuContext not found');
+    return () => context.setOpen(false);
```
gregberge: Wrap it in a useCallback, always better to have it stable. Can be passed in a useEffect

gregberge (approved):

PR #3471, opened 2025-07-14
Add browserslist and improve CSS masks for compatibility

conico974:
Introduce a browserslist configuration to support a wider range of browsers and enhance CSS mask properties for better compatibility with older browsers.

This configuration supports 93.3% of the browser worldwide

### DIFF
--- a/packages/gitbook/package.json
+++ b/packages/gitbook/package.json
@@ -110,5 +110,8 @@
         "e2e-customers": "playwright test e2e/customers.spec.ts",
         "unit": "bun test {src,packages} --preload ./tests/preload-bun.ts",
         "typecheck": "tsc --noEmit"
-    }
+    },
+    "browserslist": [
+        ">0.3%, chrome >= 64, edge >= 79, firefox >= 67, opera >= 51, safari >= 12 and not dead"
+    ]
 }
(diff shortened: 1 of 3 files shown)


packages/gitbook/package.json
```diff
@@ -110,5 +110,8 @@
         "e2e-customers": "playwright test e2e/customers.spec.ts",
         "unit": "bun test {src,packages} --preload ./tests/preload-bun.ts",
         "typecheck": "tsc --noEmit"
-    }
+    },
+    "browserslist": [
+        ">0.3%, chrome >= 64, edge >= 79, firefox >= 67, opera >= 51, safari >= 12 and not dead"
+    ]
```
gregberge, line 116: It will not evolve automatically over time you should use that instead:

```
last 2 versions
> 0.3%
not dead
not op_mini all
```

and use a `.browserlistrc`.

conico974: I took the one from next docs and customized it a bit.
https://nextjs.org/docs/architecture/supported-browsers
> If you would like to target specific browsers or features, Next.js supports [Browserslist](https://browsersl.ist/)
configuration in your package.json file

gregberge: OK let's go with this.

conico974: @gregberge Could you approve it then ?

PR #3472, opened 2025-07-14
Ensure schema.security is always an array

conico974:
Modify the handling of schema.security to guarantee it is treated as an array.

### DIFF
--- a/packages/react-openapi/src/resolveOpenAPIOperation.ts
+++ b/packages/react-openapi/src/resolveOpenAPIOperation.ts
@@ -40,11 +40,12 @@ export async function resolveOpenAPIOperation(
     }
 
     const servers = 'servers' in schema ? (schema.servers ?? []) : [];
-    const schemaSecurity = Array.isArray(schema.security ?? [])
+    const schemaSecurity = Array.isArray(schema.security)
         ? schema.security
-        : [schema.security];
-    const security: OpenAPIV3_1.SecurityRequirementObject[] =
-        operation.security ?? schemaSecurity ?? [];
+        : schema.security
+          ? [schema.security]
+          : [];
+    const security: OpenAPIV3_1.SecurityRequirementObject[] = operation.security ?? schemaSecurity;
 
     // If security includes an empty object, it means that the security is optional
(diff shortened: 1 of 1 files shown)


packages/react-openapi/src/resolveOpenAPIOperation.ts
```diff
@@ -40,8 +40,11 @@ export async function resolveOpenAPIOperation(
     }
 
     const servers = 'servers' in schema ? (schema.servers ?? []) : [];
+    const schemaSecurity = Array.isArray(schema.security ?? [])
```
valentin0h: Is this correct? we want something like this, no?

```typescript
 const schemaSecurity = Array.isArray(schema.security)
      ? schema.security
      : schema.security ? [schema.security] : [];
```

packages/react-openapi/src/resolveOpenAPIOperation.ts
```diff
@@ -40,8 +40,11 @@ export async function resolveOpenAPIOperation(
...
+        ? schema.security
+        : [schema.security];
     const security: OpenAPIV3_1.SecurityRequirementObject[] =
-        operation.security ?? schema.security ?? [];
+        operation.security ?? schemaSecurity ?? [];
```
valentin0h:
```suggestion
        operation.security ?? schemaSecurity;
```

packages/react-openapi/src/resolveOpenAPIOperation.ts
conico974: It should behave in the same way. If it's null/undefined it will return an empty array

valentin0h: Hm, maybe I'm misunderstanding, but `Array.isArray(schema.security ?? [])` will always evaluate to a truthy value. So `schemaSecurity` could end up being `null|undefined`

conico974: Ho no you're right actually, i'll use yours it's easier to understand at first glance anyway

valentin0h (approved):

PR #3479, opened 2025-07-16
Remove padding on first Heading in Columns

nolannbiron:


gregberge (approved):
Let's be sure there is some tests covering that, if not please add some in our columns block visual tests.

zenoachtig (approved):

PR #3483, opened 2025-07-18
Fix code block with PowerShell crashing published pages in Safari

clairechabas:
Fallback to plain highlighting on Safari for code blocks using PowerShell language. 

PowerShell commands can trigger complex regex that Safari JS engine doesn't support, causing the highlighter to crash.

### DIFF
--- a/packages/gitbook/src/components/DocumentView/CodeBlock/highlight.ts
+++ b/packages/gitbook/src/components/DocumentView/CodeBlock/highlight.ts
@@ -34,6 +34,9 @@ export type RenderedInline = {
     body: React.ReactNode;
 };
 
+const isSafari =
+    typeof navigator !== 'undefined' && /^((?!chrome|android).)*safari/i.test(navigator.userAgent);
+
 const theme = createCssVariablesTheme();
 
 const { getSingletonHighlighter } = createSingletonShorthands(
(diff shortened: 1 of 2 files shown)


packages/gitbook/src/components/DocumentView/CodeBlock/highlight.ts
```diff
@@ -65,8 +68,12 @@ export async function highlight(
     inlines: RenderedInline[]
 ): Promise<HighlightLine[]> {
     const langName = getBlockLang(block);
-    if (!langName) {
-        // Language not found, fallback to plain highlighting
+    if (!langName || (isSafari && langName === 'powershell')) {
```
emmerich, line 71: If powershell doesn't work, it's a bug in shiki. I saw our shiki is a few versions behind, did you try updating it and seeing if it's been fixed?

SamyPesse, line 71: Instead of hard-coding such condition; should we be more safe and wrap the whole thing in a try...catch and fallback to plainHighlight if it fails?

This looks like too much hard-coded to me and if it fails for this, it could fail for others.

clairechabas:
Closing in favor of a better fix : [https://github.com/GitbookIO/gitbook/pull/3484](https://github.com/GitbookIO/gitbook/pull/3484)

PR #3485, opened 2025-07-18
Page actions customisation

zenoachtig:


packages/gitbook/src/components/PageBody/PageHeader.tsx
```diff
@@ -45,7 +49,7 @@ export async function PageHeader(props: {
                         markdownPageUrl={`${context.linker.toAbsoluteURL(context.linker.toPathInSpace(page.path))}.md`}
                         withAIChat={withAIChat}
                         trademark={context.customization.trademark.enabled}
-                        withLLMActions={context.site.visibility === SiteVisibility.Public}
+                        actions={context.customization.pageActions}
```
nolannbiron: Is this false by default for private site in GBX ?

zenoachtig: I've added an override, forcing the externalAI actions off on non-public sites. We should handle this in the API instead, setting the default depending on site visibility instead.

@nolannbiron mentioned he could take a look later. Cc @SamyPesse

PR #3487, opened 2025-07-18
Make sure to use the correct version for adaptive content

conico974:
Use contextId from `PublishedContentSite` to force refresh RSC data based on it.
When the claims change, the whole RSC data would get refreshed.

### DIFF
--- a/packages/gitbook/next.config.mjs
+++ b/packages/gitbook/next.config.mjs
@@ -14,6 +14,9 @@ const nextConfig = {
             dynamic: 3600, // 1 hour
             static: 3600, // 1 hour
         },
+
+        // Since content is fully static, we don't want to fetch on hover again
+        optimisticClientCache: false,
     },
 
     env: {
--- a/packages/gitbook/src/components/SiteLayout/ClientContexts.tsx
+++ b/packages/gitbook/src/components/SiteLayout/ClientContexts.tsx
@@ -3,6 +3,7 @@
 import type { CustomizationThemeMode, SiteExternalLinksTarget } from '@gitbook/api';
 import { ThemeProvider } from 'next-themes';
 import type React from 'react';
+import { RouterCacheClearer } from '../hooks/useClearRouterCache';
 import { LinkSettingsContext } from '../primitives';
 
 export function ClientContexts(props: {
@@ -26,6 +27,7 @@ export function ClientContexts(props: {
     return (
         <ThemeProvider nonce={nonce} attribute="class" enableSystem forcedTheme={forcedTheme}>
             <LinkSettingsContext.Provider value={{ externalLinksTarget, adaptiveContentHash }}>
+                <RouterCacheClearer />
                 {children}
             </LinkSettingsContext.Provider>
         </ThemeProvider>
--- a/packages/gitbook/src/components/SiteLayout/ClientContexts.tsx
+++ b/packages/gitbook/src/components/SiteLayout/ClientContexts.tsx
@@ -9,9 +9,10 @@ export function ClientContexts(props: {
     nonce?: string;
     forcedTheme: CustomizationThemeMode | undefined;
     externalLinksTarget: SiteExternalLinksTarget;
+    adaptiveContentHash?: string;
     children: React.ReactNode;
 }) {
-    const { children, forcedTheme, externalLinksTarget } = props;
+    const { children, forcedTheme, externalLinksTarget, adaptiveContentHash } = props;
 
     /**
      * A bug in ThemeProvider is causing the nonce to be included incorrectly
@@ -24,7 +25,7 @@ export function ClientContexts(props: {
 
     return (
         <ThemeProvider nonce={nonce} attribute="class" enableSystem forcedTheme={forcedTheme}>
-            <LinkSettingsContext.Provider value={{ externalLinksTarget }}>
+            <LinkSettingsContext.Provider value={{ externalLinksTarget, adaptiveContentHash }}>
                 {children}
             </LinkSettingsContext.Provider>
         </ThemeProvider>
--- a/packages/gitbook/src/components/SiteLayout/ClientContexts.tsx
+++ b/packages/gitbook/src/components/SiteLayout/ClientContexts.tsx
@@ -4,7 +4,7 @@ import type { CustomizationThemeMode, SiteExternalLinksTarget } from '@gitbook/a
 import { ThemeProvider } from 'next-themes';
 import type React from 'react';
 import { RouterCacheClearer } from '../hooks/useClearRouterCache';
-import { Link, LinkSettingsContext } from '../primitives';
+import { LinkSettingsContext } from '../primitives';
 
 export function ClientContexts(props: {
     nonce?: string;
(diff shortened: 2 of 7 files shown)


conico974:
I've spent way too much time trying to understand what's going on here.
It seems that the `staleTimes` that we set for the client is not always respected. 
Sometimes before the 5 minutes where it should serve stale requests from the client cache, it still triggers a request.

It's not perfect, but this PR should allow us to be safe and have a good enough client cache.
The only downsides is the "ugly" URL that we get

packages/gitbook/next.config.mjs
```diff
@@ -14,6 +14,9 @@ const nextConfig = {
             dynamic: 3600, // 1 hour
             static: 3600, // 1 hour
         },
+
+        // Since content is fully static, we don't want to fetch on hover again
+        optimisticClientCache: false,
```
SamyPesse, line 19: This is exactly what helps make sure perfs are good, no?
Otherwise the links are not pre-fetched and you get a first-time render when clicking or just the time to download the page.

What is the reason for disabling this?

packages/gitbook/src/components/SiteLayout/SiteLayout.tsx
```diff
@@ -46,10 +47,17 @@ export async function SiteLayout(props: {
         });
     });
 
+    // Set the adaptive content hash to the current version of the content.
+    const adaptiveContentHash =
+        Object.keys(visitorAuthClaims).length > 0
+            ? hash('sha256', JSON.stringify(visitorAuthClaims))
+            : undefined;
```
SamyPesse: Instead of computing our own hash, we could use https://github.com/GitbookIO/gitbook-x/blob/main/packages/api-client/spec/schemas/content/PublishedSiteContent.yaml#L43-L45 which is provided by the API

@spastorelli was thinking about removing it as it's no longer used, but if it's served here, let's keep it.

packages/gitbook/src/components/primitives/Link.tsx
```diff
@@ -123,11 +127,19 @@ export const Link = React.forwardRef(function Link(
         );
     }
 
+    const newHref = adaptiveContentHash
+        ? `${href}${href.includes('?') ? '&' : '?'}ach=${adaptiveContentHash}`
```
SamyPesse: So all URLs will include this hash?

That's not great.

SamyPesse (commented):
While the approach makes sense, it forces us to have ugly URls and also to handle ALL urls everywhere.


Can we instead have an effect like:

```
React.useEffect(() => {
   next.clearRouterCacheSomehow();
}, [cacheContextKey]);
```

?

packages/gitbook/next.config.mjs
conico974: This one seems counter intuitive, but it's actually what we want.
https://github.com/vercel/next.js/discussions/40268?sort=top#discussioncomment-3572642
It's just a terrible name

packages/gitbook/src/components/primitives/Link.tsx
conico974: The other option is to force a hard refresh (and we'll need to change the cache-control to never cache on the client, especially on cloudflare)

conico974:
One thing with it right now, for unsigned claims using a `<Link>` with prefetch will cause trouble because the cookies would be set before actually clicking on it.
It could be interesting to provide a custom button for that so that we can more easily control that

packages/gitbook/src/middleware.ts
```diff
@@ -328,6 +329,11 @@ async function serveSiteRoutes(requestURL: URL, request: NextRequest) {
         response.headers.set('x-gitbook-route-type', routeType);
         response.headers.set('x-gitbook-route-site', siteURLWithoutProtocol);
 
+        // When we use adaptive content, we want to ensure that the cache is not used at all on the client side.
+        if (siteURLData.contextId) {
+            response.headers.set('cache-control', 'public, max-age=0, must-revalidate');
```
SamyPesse, line 335: I'm already getting this header in prod. Why do we need this? only for Cloudflare? If yes, let's add a comment to explain

Image

packages/gitbook/src/components/SiteLayout/ClientContexts.tsx
```diff
@@ -24,7 +26,8 @@ export function ClientContexts(props: {
 
     return (
         <ThemeProvider nonce={nonce} attribute="class" enableSystem forcedTheme={forcedTheme}>
-            <LinkSettingsContext.Provider value={{ externalLinksTarget }}>
+            <LinkSettingsContext.Provider value={{ externalLinksTarget, contextId }}>
```
SamyPesse: You probably don't need it in the context if we can pass it directly as a prop to `RouterCacheClearer`

packages/gitbook/src/components/SiteLayout/ClientContexts.tsx
```diff
@@ -3,15 +3,17 @@
...
 import { ThemeProvider } from 'next-themes';
 import type React from 'react';
+import { RouterCacheClearer } from '../hooks/useClearRouterCache';
 import { LinkSettingsContext } from '../primitives';
 
 export function ClientContexts(props: {
     nonce?: string;
     forcedTheme: CustomizationThemeMode | undefined;
     externalLinksTarget: SiteExternalLinksTarget;
+    contextId?: string;
```
SamyPesse: I think we should avoid optional prop, `string | undefined` is better

packages/gitbook/src/components/hooks/useClearRouterCache.tsx
```diff
@@ -0,0 +1,37 @@
...
+        // This prevents unnecessary cache clearing on the first render.
+        if (contextId !== previousContextId && previousContextId !== undefined) {
+            previousContextId = contextId;
+            // Trigger a full reload to clear the in-memory cache
+            router.refresh(); // This will clear the cache and re-fetch the data
+        }
+    }, [contextId, router]);
+}
+
+export const RouterCacheClearer = () => {
```
SamyPesse: Why expose a component instead of just using the hook in `ClientContexts`?

packages/gitbook/src/components/primitives/Link.tsx
```diff
@@ -37,6 +37,10 @@ export type LinkProps = Omit<BaseLinkProps, 'href'> &
  */
 export const LinkSettingsContext = React.createContext<{
     externalLinksTarget: SiteExternalLinksTarget;
+    /**
+     * Context ID used by adaptive content. It is used to clear the router cache.
+     */
+    contextId?: string;
```
SamyPesse: I'd recommend removing it if we don't actually use it for links

packages/gitbook/src/lib/context.ts
```diff
@@ -123,6 +124,9 @@ export type GitBookSiteContext = GitBookSpaceContext & {
 
     /** Scripts to load for the site. */
     scripts: SiteIntegrationScript[];
+
+    /** Context ID used by adaptive content. It is used to clear the router cache */
```
SamyPesse:
```suggestion
    /** Context ID used by adaptive content. It represents an unique identifier for the authentication context */
```

SamyPesse (approved):

packages/gitbook/src/middleware.ts
conico974: Yeah it's only for cloudflare, i'll add a comment

packages/gitbook/src/components/hooks/useClearRouterCache.tsx
conico974: I was using the context, but yeah it's not useful

PR #3495, opened 2025-07-22
Tweaks to AI Chat and AI Actions Dropdown

zenoachtig:
Resolves multiple small issues

- AI Actions dropdown was overlapping with OpenAPI "on this page"
Screenshot 2025-07-22 at 18 25 16
Screenshot 2025-07-22 at 18 25 13

- Styling improvements to AIChatToolsSummary and the messages placeholder.
- Added `select-none` to parts of AIChat so selection works better
- Renamed "More" menu to "Actions"
- Improved "based on your context" tooltip behaviour, now also accessible on mobile
- Improved animation naming and consistency

### DIFF
--- a/packages/gitbook/src/components/primitives/Tooltip.tsx
+++ b/packages/gitbook/src/components/primitives/Tooltip.tsx
@@ -1,5 +1,9 @@
+'use client';
+
 import { tcls } from '@/lib/tailwind';
 import * as RadixTooltip from '@radix-ui/react-tooltip';
+import { useState } from 'react';
+import { useIsMobile } from '../hooks/useIsMobile';
 
 export function Tooltip(props: {
     children: React.ReactNode;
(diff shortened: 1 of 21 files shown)


packages/gitbook/src/components/primitives/Tooltip.tsx
```diff
@@ -22,10 +26,24 @@ export function Tooltip(props: {
...
     } = props;
 
+    const isMobile = useIsMobile();
+    const [open, setOpen] = useState(false);
+
     return (
         <RadixTooltip.Provider delayDuration={300} {...providerProps}>
-            <RadixTooltip.Root {...rootProps}>
-                <RadixTooltip.Trigger asChild {...triggerProps}>
+            <RadixTooltip.Root open={open} onOpenChange={(e) => setOpen(e)} {...rootProps}>
```
clairechabas, line 34: Are you sure this is the order you want here? If the `open` and `onOpenChange` props are present in the `rootProps` then the ones you added there will be overridden.

If you want to enforce those props then you'll want to spread the `rootProps` first:
```suggestion
            <RadixTooltip.Root {...rootProps} open={open} onOpenChange={(e) => setOpen(e)}>
```

packages/gitbook/src/components/primitives/Tooltip.tsx
```diff
@@ -22,10 +26,24 @@ export function Tooltip(props: {
...
+                    asChild
+                    onClick={(event) => {
+                        event.preventDefault();
+                        if (isMobile) {
+                            setOpen(true);
+                        }
+                        setOpen(true);
+                    }}
+                    onPointerDown={(event) => event.preventDefault()}
+                    {...triggerProps}
```
clairechabas, line 45: Same comment for the props order on the trigger here, just in case.

packages/gitbook/src/components/primitives/Tooltip.tsx
zenoachtig: No I deliberately put them in this order: this way the default is to allow this behaviour, but if one so desires they can override it by setting a custom onClick.

clairechabas: Sounds good just wanted to be sure 👍

PR #3497, opened 2025-07-23
Fix headings overflowing in hint block

clairechabas:
The `break-words` property was set on the hint content container which works for paragraphs but doesn't work for headings because headings include 2 children: the heading text and their hash link button.

### DIFF
--- a/packages/gitbook/src/components/DocumentView/Heading.tsx
+++ b/packages/gitbook/src/components/DocumentView/Heading.tsx
@@ -46,6 +46,8 @@ export function Heading(props: BlockProps<DocumentBlockHeading>) {
                     'flex-1',
                     'z-[1]',
                     'justify-self-start',
+                    'max-w-full',
+                    'break-words',
                     getTextAlignment(block.data.align),
                     textStyle.lineHeight
                 )}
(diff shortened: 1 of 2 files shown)


gregberge:
@clairechabas ask @zenoachtig if the heading should be full-width or not.

clairechabas:
@gregberge I checked our docs for this because I wasn't sure either, and it's not full-width there (see [here](https://gitbook.com/docs/creating-content/blocks/hint#this-is-a-h2-heading)). So it looks to be by design.
But @zenoachtig if you'd rather change it let me know.

zenoachtig:
> @gregberge I checked our docs for this because I wasn't sure either, and it's not full-width there (see [here](https://gitbook.com/docs/creating-content/blocks/hint#this-is-a-h2-heading)). So it looks to be by design. But @zenoachtig if you'd rather change it let me know.

Oh no that looks like a bug – the heading "bar" should span the complete width!

gregberge:
@zenoachtig @clairechabas That's what I thought, @claire can you fix it at the same time?

clairechabas:
Yep it's fixed 👍  Thanks for confirming @zenoachtig!

zenoachtig (approved):

PR #3499, opened 2025-07-23
Improve AI Actions dropdown

zenoachtig:
- Add `ButtonGroup` component
- Add logic to not show the dropdown button if there's only 1 action
- Add menu separator component

Screenshot 2025-07-23 at 18 52 33

### DIFF
--- a/packages/gitbook/src/components/primitives/DropdownMenu.tsx
+++ b/packages/gitbook/src/components/primitives/DropdownMenu.tsx
@@ -203,6 +203,12 @@ export function DropdownSubMenu(props: { children: React.ReactNode; label: React
     );
 }
 
+export function DropdownMenuSeparator() {
+    return (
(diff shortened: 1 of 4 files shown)


packages/gitbook/src/components/primitives/DropdownMenu.tsx
```diff
@@ -203,6 +203,12 @@ export function DropdownSubMenu(props: { children: React.ReactNode; label: React
     );
 }
 
+export function DropdownMenuSeparator() {
+    return (
+        <RadixDropdownMenu.Separator className="h-px w-full border-tint border-t first:hidden" />
```
nolannbiron: not sure we should handle the `first:hidden` it can be confusing

nolannbiron (approved):
Make sure the gap between the items and the separators is the same as the padding

Image

PR #3501, opened 2025-07-24
Improve french translations

gregberge:


packages/gitbook/src/intl/translations/fr.ts
```diff
@@ -1,71 +1,68 @@
...
+    sponsored_via_gitbook: 'Sponsorisé via GitBook',
+    switch_to_dark_theme: 'Activer le thème sombre',
+    switch_to_light_theme: 'Activer le thème clair',
+    switch_to_system_theme: 'Utiliser le thème du système',
     search: 'Rechercher',
-    search_or_ask: 'Demander ou rechercher',
-    search_input_placeholder: 'Rechercher le contenu',
-    search_ask_input_placeholder: 'Rechercher du contenu ou poser une question',
-    search_no_results_for: 'Aucun résultat pour "${1}".',
+    search_or_ask: 'Poser une question ou rechercher',
```
zenoachtig: Not saying we shouldn't do this, but this is a very long string for the search input so it will be cut off on the smaller input size

gregberge: I will just put "Rechercher" that's fine.

PR #3503, opened 2025-07-24
Implement basic URL scheme for Assistant & AI Search

zenoachtig:


packages/gitbook/src/components/AI/useAIChat.tsx
```diff
@@ -5,6 +5,7 @@ import * as zustand from 'zustand';
 import { AIMessageRole } from '@gitbook/api';
 import * as React from 'react';
 import { useTrackEvent } from '../Insights';
+import { useSearch } from '../Search/useSearch';
```
taranvohra: We should avoid nested imports. Export it from an `index.ts`

packages/gitbook/src/components/Search/SearchContainer.tsx
```diff
@@ -40,17 +41,27 @@ export function SearchContainer(props: SearchContainerProps) {
...
 
-    const onClose = async (to?: string) => {
-        if (state?.query === '') {
-            await setSearchState(null);
-        } else if (state) {
-            await setSearchState({ ...state, open: false });
+    // Close the pop-over when we switch to “ask” mode
+    React.useEffect(() => {
+        if (state?.ask && open && aiMode === CustomizationAIMode.Assistant) {
+            setSearchState((prev) => (prev ? { ...prev, open: false } : prev));
```
taranvohra: Why do you need a conditional? Why can't it be just `setSearchState(prev => ({ ...prev, open: false })`

zenoachtig: The conditional is needed because the closing of the search box only needs to happen when `aiMode` is set to Assistant, where it "hands off" to the GitBook Assistant chat window. When it's in Search mode, the search modal itself stays open and handles the question. So while you're right, the conditional is not necessary to set it depending on `ask`, (you could just use `setSearchState((prev) => (prev ? { ...prev, open: !state.ask } : prev));`), we still need to gate that behaviour based on `aiMode`.

I've changed the expression slightly to not do an `open` check anymore and instead just do an inversion of `state.ask`, but kept the conditional for `aiMode`.

zenoachtig:
> Code looks good to me. Maybe we should add couple more E2E tests for the behavior. for ex - type in search bar -> click Assistant icon -> should open the AI chat etc

Thanks Taran, I added a bunch of tests now 😄

PR #3510, opened 2025-07-30
Fix AI Search Query

zenoachtig:


packages/gitbook/src/components/Search/SearchContainer.tsx
```diff
@@ -147,7 +148,7 @@ export function SearchContainer(props: SearchContainerProps) {
                                     aiMode={aiMode}
                                 />
                             ) : null}
-                            {state?.ask ? <SearchAskAnswer query={normalizedQuery} /> : null}
+                            {state?.ask ? <SearchAskAnswer query={normalizedAsk} /> : null}
```
taranvohra: Probably want `normalizedAsk ?` in the ternary no?

PR #3511, opened 2025-07-30
Don't track events if a special middleware header is set.

emmerich:
In order to prevent inflated values on tracking whilst users view sites through previews, we will disable all tracking for dynamic routes when a specific middleware is set.

### DIFF
--- a/packages/gitbook/src/middleware.ts
+++ b/packages/gitbook/src/middleware.ts
@@ -141,14 +141,12 @@ async function serveSiteRoutes(requestURL: URL, request: NextRequest) {
         url: siteRequestURL,
     });
 
-    const withAPIToken = async (
-        apiToken: string | null,
+    //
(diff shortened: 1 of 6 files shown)


packages/gitbook/src/middleware.ts
```diff
@@ -319,6 +326,10 @@ async function serveSiteRoutes(requestURL: URL, request: NextRequest) {
             },
         });
 
+        Object.entries(responseHeaders ?? {}).forEach(([key, value]) => {
+            response.headers.set(key, value);
```
conico974: I don't think this should be added to the response headers, but rather to the request headers. It's not a big deal, it works this way, but it feels wrong to add this to the response

conico974 (approved):
Another small things, we should filter this header from the request itself, only us should set this header. Someone behind a proxy could trick us into disabling tracking

PR #3514, opened 2025-07-30
Suggest questions in the current space context

taranvohra:
This PR ensures we show the recommended questions based off the specific space the user is in as context. It avoids questions coming from other spaces in other sections which might not be relevant in the current context the user is in

### DIFF
--- a/packages/gitbook/src/components/Search/SearchResults.tsx
+++ b/packages/gitbook/src/components/Search/SearchResults.tsx
@@ -36,6 +36,7 @@ type ResultType =
  * We cache the recommended questions globally to avoid calling the API multiple times
  * when re-opening the search modal. The cache is per space, so that we can
  * have different recommended questions for different spaces of the same site.
+ * It should not be used outside of an useEffect.
  */
 const cachedRecommendedQuestions: Map<string, ResultType[]> = new Map();
 


packages/gitbook/src/components/Search/SearchResults.tsx
```diff
@@ -33,9 +34,10 @@ type ResultType =
 
 /**
  * We cache the recommended questions globally to avoid calling the API multiple times
- * when re-opening the search modal.
+ * when re-opening the search modal. The cache is per space, so that we can
+ * have different recommended questions for different spaces of the same site.
  */
-let cachedRecommendedQuestions: null | ResultType[] = null;
+const cachedRecommendedQuestions: Map<string, ResultType[]> = new Map();
```
conico974, line 41: Not sure how useful it would be given that this is deployed in a serverless env.
It could actually have the side effects of growing way too much

taranvohra: It's a client component though. We only cache it if user is navigating between sections or variants

conico974: Ho yeah my bad i missed that. 
Could you add a comment that it should not be used outside of a `useEffect`, client component are still executed on the server

PR #3516, opened 2025-07-30
Show a HoverCard when an internal link is broken

gregberge:
Also refactor the code to abstract HoverCard primitive and fix the TooltipProvider.

CleanShot 2025-07-30 at 16 50 19@2x

### DIFF
--- a/packages/gitbook/src/components/DocumentView/InlineLink/InlineLink.tsx
+++ b/packages/gitbook/src/components/DocumentView/InlineLink/InlineLink.tsx
@@ -1,8 +1,7 @@
 import { type DocumentInlineLink, SiteInsightsLinkPosition } from '@gitbook/api';
 
 import { getSpaceLanguage, tString } from '@/intl/server';
-import { languages } from '@/intl/translations';
-import type { GitBookAnyContext } from '@/lib/context';
+import { type TranslationLanguage, languages } from '@/intl/translations';
 import { type ResolvedContentRef, resolveContentRef } from '@/lib/references';
 import { Icon } from '@gitbook/icons';
 import { HoverCard, HoverCardRoot, HoverCardTrigger, StyledLink } from '../../primitives';
@@ -19,8 +18,14 @@ export async function InlineLink(props: InlineProps<DocumentInlineLink>) {
               resolveAnchorText: false,
           })
         : null;
+    const { contentContext } = context;
 
-    if (!context.contentContext || !resolved) {
+    const language =
+        contentContext && 'customization' in contentContext
+            ? getSpaceLanguage(contentContext.customization)
+            : languages.en;
+
+    if (!contentContext || !resolved) {
         return (
             <HoverCardRoot>
                 <HoverCardTrigger>
@@ -36,11 +41,9 @@ export async function InlineLink(props: InlineProps<DocumentInlineLink>) {
                 <HoverCard className="flex flex-col gap-1 p-4">
                     <div className="flex items-center gap-2">
                         <Icon icon="ban" className="size-4 text-tint-subtle" />
-                        <h5 className="font-semibold">Page not found</h5>
+                        <h5 className="font-semibold">{tString(language, 'notfound_title')}</h5>
                     </div>
-                    <p className="text-sm text-tint">
-                        This link points to a page that has been removed or no longer exists.
-                    </p>
+                    <p className="text-sm text-tint">{tString(language, 'notfound_link')}</p>
                 </HoverCard>
             </HoverCardRoot>
         );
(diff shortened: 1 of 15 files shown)


packages/gitbook/src/components/DocumentView/InlineLink/InlineLink.tsx
```diff
@@ -22,14 +22,27 @@ export async function InlineLink(props: InlineProps<DocumentInlineLink>) {
...
+                            document={document}
+                            nodes={inline.nodes}
+                            ancestorInlines={[...ancestorInlines, inline]}
+                        />
+                    </span>
+                </HoverCardTrigger>
+                <HoverCard className="flex flex-col gap-1 p-4">
+                    <div className="flex items-center gap-2">
+                        <Icon icon="ban" className="size-4 text-tint-subtle" />
+                        <h5 className="font-semibold">Page not found</h5>
```
zenoachtig: Should be a translateable string

packages/gitbook/src/components/DocumentView/InlineLink/InlineLink.tsx
```diff
@@ -22,14 +22,27 @@ export async function InlineLink(props: InlineProps<DocumentInlineLink>) {
...
+                    </div>
+                    <p className="text-sm text-tint">
+                        This link points to a page that has been removed or no longer exists.
```
zenoachtig: Should be a translateable string

packages/gitbook/src/components/DocumentView/InlineLink/InlineLinkTooltipImpl.tsx
```diff
@@ -21,89 +20,76 @@ export function InlineLinkTooltipImpl(props: {
...
-                    <Tooltip.Content className="z-40 w-screen max-w-md animate-present px-4 sm:w-auto">
-                        <div className="overflow-hidden rounded-md straight-corners:rounded-none shadow-lg shadow-tint-12/4 ring-1 ring-tint-subtle dark:shadow-tint-1 ">
-                            <div className="bg-tint-base p-4">
-                                <div className="flex items-start gap-4">
-                                    <div className="flex flex-col">
-                                        {breadcrumbs && breadcrumbs.length > 0 ? (
-                                            <div className="mb-1 flex grow flex-wrap items-center gap-x-2 gap-y-0.5 font-semibold text-tint text-xs uppercase leading-tight tracking-wide">
-                                                {breadcrumbs.map((crumb, index) => {
-                                                    const Tag = crumb.href ? StyledLink : 'div';
+        <HoverCardRoot>
```
zenoachtig, line 23: Why make this a component separate from Popover or Tooltip?

gregberge: - A Tooltip is a tiny text displayed on elements it has no pointer events
- A Popover is something displayed when we click on an trigger
- An HoverCard is like a Popover but triggered by an hover event

https://www.radix-ui.com/primitives/docs/components/hover-card
https://www.radix-ui.com/primitives/docs/components/popover
https://www.radix-ui.com/primitives/docs/components/tooltip

We should actually use the HoverCard for our component but it was like that I don't want to refactor everything.

packages/gitbook/src/components/DocumentView/InlineLink/InlineLink.tsx
gregberge: Good catch, thanks!

packages/gitbook/src/components/DocumentView/InlineLink/InlineLink.tsx
```diff
@@ -19,17 +18,34 @@ export async function InlineLink(props: InlineProps<DocumentInlineLink>) {
...
         return (
-            <span title="Broken link" className="underline">
-                <Inlines
-                    context={context}
-                    document={document}
-                    nodes={inline.nodes}
-                    ancestorInlines={[...ancestorInlines, inline]}
-                />
-            </span>
+            <HoverCardRoot>
```
conico974, line 30: This is likely not a big deal given that it's only for broken link, but all these HoverCard could be in a separate client component and only take a children as props. 
This would reduce the size of the RSC

packages/gitbook/src/components/DocumentView/InlineLink/InlineLinkTooltipImpl.tsx
zenoachtig: Alright, makes a lot of sense. Let's keep the refactor for later, we should indeed.

zenoachtig (approved):

packages/gitbook/src/components/DocumentView/InlineLink/InlineLinkTooltipImpl.tsx
gregberge: Not so important, it works well honestly I don't even know what is the difference with the HoverCard from Radix.

PR #3527, opened 2025-08-04
Fix image format handling and enhance PageCoverImage component

conico974:
Make sure to only use a correct DPR for avif/webp component.
PageCover now also use a srcSet


emmerich (approved):
To be tested on card cover images, as @nolannbiron also made changes here to support dark cover images

PR #3539, opened 2025-08-06
Tweaks to AIChatButton and AIChatInput

zenoachtig:
- Switch to Cmd+I to open Assistant
- AIChatButton now only ever opens the assistant
- Clean up AIChatInput to better account for long translated strings and add prettier focus state
- Improve some color inconsistencies in `Button` that were showing up in Assistant

### DIFF
--- a/packages/gitbook/src/components/AIChat/AIChatButton.tsx
+++ b/packages/gitbook/src/components/AIChat/AIChatButton.tsx
@@ -2,7 +2,7 @@
 import { useLanguage } from '@/intl/client';
 import { t } from '@/intl/translate';
 import { tcls } from '@/lib/tailwind';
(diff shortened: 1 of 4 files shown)


packages/gitbook/src/components/AIChat/AIChatButton.tsx
```diff
@@ -2,7 +2,7 @@
 import { useLanguage } from '@/intl/client';
 import { t } from '@/intl/translate';
 import { tcls } from '@/lib/tailwind';
-import { useAIChatController, useAIChatState } from '../AI/useAIChat';
+import { useAIChatController } from '../AI/useAIChat';
```
taranvohra: Let's fix the nested import while we're at it ;)

taranvohra (approved):
Missing a changeset

PR #3541, opened 2025-08-06
Add HashContext and HashProvider for managing URL hash state

conico974:
Link with an hash on the same page were not triggering change from the `useHash` hook.

### DIFF
--- a/packages/gitbook/src/components/hooks/useHash.tsx
+++ b/packages/gitbook/src/components/hooks/useHash.tsx
@@ -3,6 +3,11 @@ import React from 'react';
 
 export const HashContext = React.createContext<{
     hash: string | null;
+    /**
+     * Updates the hash value from the URL provided here.
+     * It will then be used by the `useHash` hook.
+     * URL can be relative or absolute.
+     */
     updateHashFromUrl: (href: string) => void;
 }>({
     hash: null,


packages/gitbook/src/components/hooks/useHash.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+function getHash(): string | null {
+    if (typeof window === 'undefined') {
+        return null;
+    }
+    return window.location.hash.slice(1);
+}
+
+export const HashProvider: React.FC<React.PropsWithChildren<{}>> = ({ children }) => {
+    const [hash, setHash] = React.useState<string | null>(getHash);
```
emmerich, line 25:
```suggestion
    const [hash, setHash] = React.useState<string | null>(getHash());
```

not sure how typescript didn't catch this one 👀

packages/gitbook/src/components/hooks/useHash.tsx
```diff
@@ -0,0 +1,47 @@
+'use client';
+import React from 'react';
+
+export const HashContext = React.createContext<{
+    hash: string | null;
+    updateHashFromUrl: (href: string) => void;
```
emmerich, line 11: Can you add a comment as to what the function does?

emmerich (approved):

packages/gitbook/src/components/hooks/useHash.tsx
conico974: Ho no that's normal, you can just pass a function to `useState` instead of a value

PR #3547, opened 2025-08-08
Support new coverDefinitionDark for cards & image type

nolannbiron:


packages/gitbook/src/components/DocumentView/Table/utils.ts
```diff
@@ -12,6 +17,23 @@ export function getRecordValue<T extends number | string | boolean | string[] |
     return record.values[definitionId];
 }
 
+/**
+ * Get the covers for a record card.
+ * Returns both the light and dark covers.
+ * The light cover is a string or a content ref (image or files column type).
+ * The dark cover is a content ref (image column type).
+ */
+export function getRecordCardCovers(record: DocumentTableRecord, view: DocumentTableViewCards) {
```
emmerich:
```suggestion
export function getRecordCardCovers(record: DocumentTableRecord, view: DocumentTableViewCards): { light: ContentRefFile | ContentRefURL | null; dark: ContentRefFile | ContentRefURL | null } {
```

Better to be explicit in the typing

packages/gitbook/src/components/DocumentView/Table/utils.ts
```diff
@@ -12,6 +17,23 @@ export function getRecordValue<T extends number | string | boolean | string[] |
...
+    return {
+        light: view.coverDefinition
```
emmerich:
```suggestion
        light: (() => {
          if (!view.coverDefinition) {
            return null;
          }
          
          const value = getRecordValue(record, view.coverDefinition) as ContentRefFile | ContentRefURL | string[];
          if(Array.isArray(value) && typeof value[0] === 'string') {
            return { kind: 'file', file: value[0] };
          }
          
          return value;
        })()
```

Do all your legacy content ref checking here. Keep the return type of this util really clean - just return content refs.

packages/gitbook/src/components/DocumentView/Table/RecordCard.tsx
```diff
@@ -21,25 +19,29 @@ export async function RecordCard(
...
+    const { dark, light } = getRecordCardCovers(record[1], view);
     const targetRef = view.targetDefinition
         ? (record[1].values[view.targetDefinition] as ContentRef)
         : null;
 
-    const [cover, target] = await Promise.all([
-        coverFile && context.contentContext
-            ? resolveContentRef({ kind: 'file', file: coverFile }, context.contentContext)
+    // Support for files coverDefinition (legacy)
+    const lightContentRef: ContentRef =
```
emmerich: See my other comment, you won't need this anymore - light/dark will both be of type `ContentRefFile | ContentRefURL | null`

packages/gitbook/src/components/DocumentView/Table/RecordCard.tsx
```diff
@@ -63,23 +57,32 @@ export async function RecordCard(
...
+                              : '',
+                          darkCoverIsSquareOrPortrait
+                              ? 'dark:grid-cols-[40%__1fr] dark:min-[432px]:grid-cols-none dark:min-[432px]:grid-rows-[auto_1fr]'
+                              : '',
+                      ].filter(Boolean)
                     : 'grid-rows-[auto_1fr]'
             )}
         >
-            {cover ? (
+            {lightCover ? (
```
emmerich, line 72: So if you have a `darkCover` but no `lightCover`, the image won't render? Does our table schema protect against that?

packages/gitbook/src/components/DocumentView/Table/RecordColumnValue.tsx
```diff
@@ -329,6 +329,51 @@ export async function RecordColumnValue<Tag extends React.ElementType = 'div'>(
...
+
+            if (!image) {
+                return null;
+            }
+
+            return (
+                <Tag className={tcls('text-base')} aria-labelledby={ariaLabelledBy}>
+                    <StyledLink
+                        href={image.href}
+                        target="_blank"
```
emmerich: I don't think you should define this, let the `StyledLink` decide it based on the href. Sometimes we open in a new tab, sometimes now.

```suggestion
```

PR #3552, opened 2025-08-11
Never follow redirect in the middleware

conico974:
Configure the middleware to not follow redirects, ensuring that redirects are handled by the client instead.

This was causing 404 for most redirects on the first call after an invalidation in 2C.


conico974:
> What does it fix?

On 2c for redirects from within a page (i.e. not from the middleware), the first ever call could return a 404. This was not necessarily noticeable, as most of them were for background request that would never be accessed. And subsequent request worked just fine.
This caused flakyness for these tests as well 
image

PR #3555, opened 2025-08-12
Improve OpenAPI parsing errors

gregberge:
We make the error more explicit, saying that we didn't find any version.

CleanShot 2025-08-12 at 14 29 52@2x

### DIFF
--- a/packages/gitbook/src/lib/openapi/fetch.ts
+++ b/packages/gitbook/src/lib/openapi/fetch.ts
@@ -36,25 +36,28 @@ export async function fetchOpenAPIFilesystem(
         return { filesystem: null, specUrl: null };
     }
 
-    const filesystem = await (() => {
+    const result = await (() => {
+        // If the reference is a new OpenAPI reference, we return it.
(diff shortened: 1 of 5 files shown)


packages/gitbook/src/lib/openapi/fetch.ts
```diff
@@ -86,21 +89,16 @@ const fetchFilesystem = async (
...
-};
+}
 
-async function fetchFilesystemUncached(
-    url: string,
-    options?: {
-        signal?: AbortSignal;
-    }
-) {
+async function fetchFilesystemNoCache(url: string) {
```
conico974, line 94: Is there a reason to change this function besides the naming itself ? Doing so will remove all the existing data cache for `fetchFilesystem`

gregberge: No there is not, but honestly it does not represent lot of them, so no problem.

conico974: Ok I'll approve then

PR #3566, opened 2025-08-15
Fix null contentRef handling in RecordColumnValue

conico974:
Improve handling of contentRef to prevent null values and ensure proper resolution in the RecordColumnValue function. Introduce a utility function to validate contentRef objects.

### DIFF
--- a/packages/gitbook/src/components/DocumentView/Table/RecordColumnValue.tsx
+++ b/packages/gitbook/src/components/DocumentView/Table/RecordColumnValue.tsx
@@ -19,7 +19,13 @@ import type { BlockProps } from '../Block';
 import { Blocks } from '../Blocks';
 import { FileIcon } from '../FileIcon';
 import type { TableRecordKV } from './Table';
-import { type VerticalAlignment, getColumnAlignment, isContentRef, isStringArray } from './utils';
+import {
+    type VerticalAlignment,
+    getColumnAlignment,
+    isContentRef,
+    isDocumentTableImageRecord,
+    isStringArray,
+} from './utils';
 
 const alignmentMap: Record<'text-left' | 'text-center' | 'text-right', string> = {
     'text-left': '**:text-left text-left',
@@ -355,12 +361,15 @@ export async function RecordColumnValue<Tag extends React.ElementType = 'div'>(
             );
         }
         case 'image': {
-            if (!isContentRef(value)) {
+            if (!isDocumentTableImageRecord(value)) {
                 return null;
             }
 
             const image = context.contentContext
-                ? await resolveContentRef(value, context.contentContext)
+                ? await resolveContentRef(
+                      'ref' in value ? value.ref : value,
+                      context.contentContext
+                  )
                 : null;
 
             if (!image) {
--- a/packages/gitbook/src/components/DocumentView/Table/RecordColumnValue.tsx
+++ b/packages/gitbook/src/components/DocumentView/Table/RecordColumnValue.tsx
@@ -19,7 +19,7 @@ import type { BlockProps } from '../Block';
 import { Blocks } from '../Blocks';
 import { FileIcon } from '../FileIcon';
 import type { TableRecordKV } from './Table';
-import { type VerticalAlignment, getColumnAlignment, isContentRef } from './utils';
+import { type VerticalAlignment, getColumnAlignment } from './utils';
 
 const alignmentMap: Record<'text-left' | 'text-center' | 'text-right', string> = {
     'text-left': '**:text-left text-left',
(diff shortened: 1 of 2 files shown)


packages/gitbook/src/components/DocumentView/Table/utils.ts
```diff
@@ -110,3 +110,7 @@ export function getColumnVerticalAlignment(column: DocumentTableDefinition): Ver
 
     return 'self-center';
 }
+
+export function isContentRef(ref: DocumentTableRecord['values'][string]): ref is ContentRef {
+    return Boolean(ref && typeof ref === 'object' && 'kind' in ref && 'file' in ref);
```
emmerich:
```suggestion
    return Boolean(ref && typeof ref === 'object' && 'kind' in ref);
```

your current check only allows file refs to pass

emmerich (changes requested):

packages/gitbook/src/components/DocumentView/Table/utils.ts
conico974: Yeah just realized that when i looked at the argos changes. I shouldn't have trusted copilot :smile:

packages/gitbook/src/components/DocumentView/Table/utils.ts
```diff
@@ -110,3 +110,7 @@ export function getColumnVerticalAlignment(column: DocumentTableDefinition): Ver
 
     return 'self-center';
 }
+
+export function isContentRef(ref: DocumentTableRecord['values'][string]): ref is ContentRef {
```
SamyPesse: Hum, why do we need this?

We should not have to check types dynamically; the API should return what it says. **Otherwise it's a problem on GBX**.

SamyPesse (changes requested):
I don't think we are fixing at the right place; I always oppose all changes that go into fixing data problems in places where we should trust the types.
The real problem here is that the API outputs something it should not; let's fix this.

packages/gitbook/src/components/DocumentView/Table/utils.ts
conico974: Steven already made the change in GBX to the API https://github.com/GitbookIO/gitbook-x/pull/19236.
`isContentRef` is not necessarily needed, `ref` can be `null | ContentRef`

conico974: Actually I think the `isContentRef` is the good solution here, by nature we can't easily type a Table as it is a user created type.
The type of `DocumentTableRecord['values'][string]` is `string | number | boolean | string[] | ContentRef | null`
Not sure if we can type this one that much better without overcomplicating things

packages/gitbook/src/components/DocumentView/Table/RecordColumnValue.tsx
```diff
@@ -329,8 +328,13 @@ export async function RecordColumnValue<Tag extends React.ElementType = 'div'>(
             );
         }
         case 'image': {
+            const contentRef = value ? (value as ContentRef) : null;
```
SamyPesse: So it can actually be null in GBX?

emmerich: the `as` here is dangerous and hides the problem. We should change line 55 to be:

```typescript
const value = record[1].values[column] as api.DocumentTableRecord['values'][String];
```

once you have that, you don't need any `as` in this file and you can handle the types properly.

conico974: Yeah that's expected, You could have a field image but not provide an image for this specific card/table cell.
Steven suggested some change so that we remove the as on every switch. I'll update that and merge it after

packages/gitbook/src/components/DocumentView/Table/RecordColumnValue.tsx
```diff
@@ -329,8 +355,12 @@ export async function RecordColumnValue<Tag extends React.ElementType = 'div'>(
             );
         }
         case 'image': {
+            if (!isContentRef(value)) {
+                return null;
+            }
+
             const image = context.contentContext
-                ? await resolveContentRef(value as ContentRef, context.contentContext)
+                ? await resolveContentRef(value, context.contentContext)
```
emmerich: Can't images also be of type `api.DocumentTableImageRecord` ? meaning you'd need to extract the ref from `value.ref`?

conico974: You're right, let me add a check for that

emmerich: How come typescript didn't pick that up though?

conico974: There is actually no error typescript wise here.
`DocumentTableImageRecord` can actually be a contentRef. When it was an object with a `ref` prop, we were just returning `null`

emmerich (approved):

PR #3569, opened 2025-08-15
Register custom assistant

taranvohra:
This PR adds the ability for integrations to register custom AI assistants. Custom AIs are surfaced everywhere the internal AI is:
- Button in the header*
- Result item in search
- Page actions menu/dropdown*
- Trigger on page load with `?ask=`**

_* Only if the integration is initialised with `ui=true` which is the case by default. If you want a "headless" AI mode, you can manually set it to `false`._ 
_** When multiple AIs are defined, the first one is triggered to handle the `?ask=` query on page load._

AI Search and GitBook Assistant are now registered via the same framework. This also allowed for a refactor of the Search and AIChat components with simplified and more agnostic state management.

https://github.com/user-attachments/assets/cdb2c28e-7ac0-4acc-884d-199861d211b7

Example registration:
```
window.GitBook.registerCustomAssistant({
    label: 'Custom AI',
    icon: 'sparkle',
    onOpen: (query) => {
        alert(`Test assistant opened: ${query ?? 'no query provided'}`);
    },
});
```

### DIFF
--- a/packages/gitbook/src/components/Search/useSearch.tsx
+++ b/packages/gitbook/src/components/Search/useSearch.tsx
@@ -1,4 +1,4 @@
-// 'use client';
+'use client';
 
 import { parseAsBoolean, parseAsString, useQueryStates } from 'nuqs';
 import React from 'react';


packages/browser-types/src/index.ts
```diff
@@ -21,6 +22,15 @@ export type GitBookIntegrationTool = AIToolDefinition & {
     execute: (input: object) => Promise<Pick<AIToolCallResult, 'output' | 'summary'>>;
 };
 
+export type Assistant = {
+    id: string;
+    label: string;
+    icon: ReactNode | string;
```
SamyPesse: I don't think it's a good idea to accept a `ReactNode` here.

Why do we need this now?

packages/browser-types/src/index.ts
```diff
@@ -21,6 +22,15 @@ export type GitBookIntegrationTool = AIToolDefinition & {
...
+    onOpen: (query?: string) => void;
+    button?: boolean;
+    mode?: 'overlay' | 'sidebar' | 'search';
```
SamyPesse: What does it mean?

We should document these values. These types are also documentation for integrations.

packages/browser-types/src/index.ts
```diff
@@ -21,6 +22,15 @@ export type GitBookIntegrationTool = AIToolDefinition & {
     execute: (input: object) => Promise<Pick<AIToolCallResult, 'output' | 'summary'>>;
 };
 
+export type Assistant = {
+    id: string;
+    label: string;
+    icon: ReactNode | string;
+    onOpen: (query?: string) => void;
+    button?: boolean;
```
SamyPesse: What does it mean?

packages/browser-types/src/index.ts
```diff
@@ -21,6 +22,15 @@ export type GitBookIntegrationTool = AIToolDefinition & {
     execute: (input: object) => Promise<Pick<AIToolCallResult, 'output' | 'summary'>>;
 };
 
+export type Assistant = {
```
SamyPesse: Prefix with `GitBook` as these are global types

packages/browser-types/src/index.ts
SamyPesse: It looks like it's because we want to use this type as an internal one for our own assistant.

We should not. This type is API-level, exposed to integrations.

We should not use it to type the assistant data; instead do:

```
type Assistant = GitBookAssistant & {
  id:
  icon: ReactNode | IconName
}
```

SamyPesse:
```suggestion
    icon: IconName;
```

packages/gitbook/src/components/AI/useAI.tsx
```diff
@@ -0,0 +1,117 @@
...
+function useAIConfig(): AIConfig {
+    const ctx = React.useContext(AIContext);
+    if (!ctx) {
+        throw new Error('useAI must be used within AIContextProvider');
+    }
+    return ctx;
+}
+
+type AIContext = AIConfig & {
+    assistants: Assistant[];
```
SamyPesse, line 61: Use a type defined here, instead of mixing everything in one type

packages/gitbook/src/components/AIActions/AIActionsDropdown.tsx
```diff
@@ -97,16 +102,17 @@ function AIActionsDropdownMenuContent(props: AIActionsDropdownProps) {
  * A default action shown as a quick-access button beside the dropdown menu
  */
 function DefaultAction(props: AIActionsDropdownProps) {
-    const { markdownPageUrl, withAIChat, trademark, actions } = props;
+    const { markdownPageUrl, actions } = props;
+    const assistants = useAI().assistants.filter((assistant) => assistant.button === true);
```
SamyPesse: I still don't understand what `button === true` means

packages/gitbook/src/components/Integrations/LoadIntegrations.tsx
```diff
@@ -42,10 +65,20 @@ if (typeof window !== 'undefined') {
                 tools: [...state.tools, tool],
             }));
         },
+        registerCustomAssistant: (assistant) => {
```
SamyPesse: I think it should be `registerAssistant` instead of `registerCustomAssistant`

packages/gitbook/src/components/Integrations/LoadIntegrations.tsx
```diff
@@ -23,6 +24,28 @@ export const integrationsAssistantTools = zustand.createStore<{
     };
 });
 
+export const customAssistants = zustand.createStore<{
+    assistants: Array<Assistant>;
+    registerAssistant: (assistant: Omit<Assistant, 'id' | 'mode'>) => () => void;
```
SamyPesse: Actually you can only store `Array<Assistant>` and then call `integrationAssistants.setState()` instead

packages/gitbook/src/components/Integrations/LoadIntegrations.tsx
```diff
@@ -23,6 +24,28 @@ export const integrationsAssistantTools = zustand.createStore<{
     };
 });
 
+export const customAssistants = zustand.createStore<{
```
SamyPesse: Le'ts rename to `integrationAssistants`

packages/browser-types/src/index.ts
```diff
@@ -21,6 +22,15 @@ export type GitBookIntegrationTool = AIToolDefinition & {
...
+    id: string;
+    label: string;
+    icon: ReactNode | string;
+    onOpen: (query?: string) => void;
```
SamyPesse: maybe it should be `open` instead of `onOpen`

packages/browser-types/src/index.ts
zenoachtig: The `mode` specifies the way the site treats the assistant, e.g. `search` assistants appear in the search box and `sidebar` assistants appear like the GitBook assistant does. This way we can abstract the internal assistants to use the exact same system we use for integration assistants. 

In the future we _could_ let integrations render within these two places in the UI if they specify this mode, but for now integration assistants are always set to `mode: overlay` and we don't have hooks to render in place. So I've taken the `mode` out of the public type and it's only used within the internal `Assistant` type.

packages/browser-types/src/index.ts
zenoachtig: Renamed to `ui: boolean;` and added a comment explaining it. This basically decides if we render UI triggers for an AI integration or not — useful if the AI wants to render its own trigger but still be recognised as an AI assistant and benefit from e.g. `?ask=` URL triggering. It's optional and defaults to true, so if an integration doesn't specify it it will render with UI by default.

packages/browser-types/src/index.ts
```diff
@@ -21,6 +21,31 @@ export type GitBookIntegrationTool = AIToolDefinition & {
...
+    /**
+     * Callback when the assistant is opened.
+     */
+    open: (query?: string) => void;
+
+    /**
+     * Whether to display the triggers for this assistant in the UI.
+     * @default true
+     */
+    ui?: boolean;
```
SamyPesse, line 46: I'm wondering if we should allow this at this stage. Why would an integration want to display an assistant with a UI?

or is it the use-case of "showing their own custom button" ?

packages/gitbook/src/components/AI/useAI.tsx
```diff
@@ -0,0 +1,141 @@
...
+        });
+    }
+
+    const integrationAssistants = useIntegrationAssistants();
+    if (integrationAssistants.length > 0) {
+        assistants.push(
+            ...integrationAssistants.map((assistant) => ({
+                ...assistant,
+                icon:
+                    typeof assistant.icon === 'string' ? (
```
SamyPesse: It's the only case possible here

packages/gitbook/src/components/AI/useAI.tsx
```diff
@@ -0,0 +1,141 @@
...
+     * - `sidebar`: Display the assistant in a sidebar next to the page. Only supported for the GitBook Assistant.
+     * - `search`: Display the assistant inside the search container. Only supported for GitBook AI Search.
+     * @default 'overlay'
+     */
+    mode?: 'overlay' | 'sidebar' | 'search';
+
+    /**
+     * Icon of the assistant displayed in the UI.
+     */
+    icon: ReactNode | IconName;
```
SamyPesse:
```suggestion
    icon: ReactNode;
```

You are already converting IconName to react node below so let's keep it a react node

packages/gitbook/src/components/AI/useAIChat.tsx
```diff
@@ -414,45 +415,6 @@ export function useAIChatController(): AIChatController {
         }));
     }, [setSearchState]);
 
-    // Auto-trigger AI chat when ?ask= parameter appears in URL (only once)
```
SamyPesse, line 417: So this whole logic moved somewhere else?

packages/gitbook/src/components/Search/SearchContainer.tsx
```diff
@@ -39,34 +42,40 @@ export function SearchContainer(props: SearchContainerProps) {
...
     const isMobile = useIsMobile();
 
-    // Derive open state from search state
-    const open = state?.open ?? false;
+    const withAI = assistants.length > 0;
+    const withSearchAI = assistants.filter((assistant) => assistant.mode === 'search').length > 0;
+    // const withAssistantAI = withAI && !withSearchAI;
 
+    // Handle initial ask state (trigger once per unique ask when assistants are ready)
+    const initial = React.useRef(state?.ask === undefined || state?.ask === null); // If ask is not set on page load, we will never trigger
```
SamyPesse: What is `initial`?

Usually we suffix refs variables with `Ref`

packages/gitbook/src/components/Search/SearchContainer.tsx
SamyPesse, line 1: Lot of changes in this file; it's hard to review but I hope we are not breaking anything

packages/gitbook/src/components/Search/SearchQuestionResultItem.tsx
```diff
@@ -11,11 +11,11 @@ export const SearchQuestionResultItem = React.forwardRef(function SearchQuestion
         question: string;
         active: boolean;
         recommended?: boolean;
-        withAIChat: boolean;
+        assistant?: Assistant;
```
SamyPesse: I'd recommend against optional prop as it's easy to miss them:


```suggestion
        assistant: Assistant | undefined;
```

packages/gitbook/src/components/Search/useSearch.ts
```diff
@@ -1,6 +1,7 @@
+'use client';
```
SamyPesse, line 1: You should not need this since the hook can only be used from a client component anyway that will have it there

packages/gitbook/src/components/Search/useSearch.ts
```diff
@@ -97,37 +98,51 @@ export function useSearch(withAIChat = false): [SearchState | null, UpdateSearch
         [setRawState]
     );
 
-    return [state, setState];
+    return React.createElement(SearchContext.Provider, { value: { state, setState } }, children);
```
SamyPesse: Wtf, why not use a tsx file and the proper JSX syntax?

packages/gitbook/types/gitbook-integrations-global.d.ts
SamyPesse, line 1: This file should be removed now

packages/browser-types/src/index.ts
zenoachtig: Discussed in chat: it's indeed mostly to register integrations with custom trigger UI.

packages/gitbook/src/components/AI/useAIChat.tsx
zenoachtig: Yep – all of this was necessary to check if an incoming ask was present on page load and handle it only under the right conditions... besides it feeling hacky to me it's also not good to hijack the ask if the Assistant might not be the only AI provider around.

Instead we're now having the component mostly in charge of the search state, `SearchContainer`, [check the list of assistants upon load](https://github.com/GitbookIO/gitbook/pull/3569/files/9fc0519a30aa8894ab4c87eac5ccf0c0a3ab57de#diff-564d77954f45276596d00e9be5a17a7950a63d40e8bfdc6122364e8f176a2bfbL45) and hand off to the first one with a simple `.open()`. Feels a lot more agnostic and clean.

packages/gitbook/src/components/Search/useSearch.ts
zenoachtig: It's actually not because of the `useSearch()` but because of the `React.useContext()` call here. The `ContextProvider` is added in `SpaceLayout` (which is a server component). If I remove it React complaisn about only using `createContext` in Client Components. Open to better ways of doing that of course.

packages/gitbook/src/components/Search/SearchContainer.tsx
zenoachtig: I thoroughly reviewed all possible `?q=` and `?ask=` cases on all different AI experience levels, but will have a last check on the deployment.

packages/gitbook/src/components/Search/useSearch.ts
zenoachtig: Sorry about that one, missed it :(

PR #3570, opened 2025-08-15
Proxy analytics events

conico974:
Middleware now handles forwarding insights events instead of going through the api directly.


packages/gitbook/src/components/Insights/InsightsProvider.tsx
```diff
@@ -216,14 +216,16 @@ export function useTrackEvent(): TrackEventCallback {
...
+    basePath: string;
     organizationId: string;
     siteId: string;
     events: api.SiteInsightsEvent[];
 }) {
-    const { apiHost, organizationId, siteId, events } = args;
-    const url = new URL(apiHost);
-    url.pathname = `/v1/orgs/${organizationId}/sites/${siteId}/insights/events`;
+    const { basePath, organizationId, siteId, events } = args;
+    const url = new URL(window.location.href);
```
SamyPesse: Maybe we should give the entire URL string as a prop; so that the server action can choose between the API one or the proxied one (if in the future we want to have it as an option)

packages/gitbook/src/middleware.ts
```diff
@@ -140,6 +140,26 @@ async function serveSiteRoutes(requestURL: URL, request: NextRequest) {
         });
     }
 
+    //Forwards analytics events
+    if (siteRequestURL.pathname.endsWith('/~gitbook/__evt')) {
+        const orgs = requestURL.searchParams.get('o');
```
SamyPesse: Can you move the inside as a function here, so that we can clearly easily the proxying

packages/gitbook/src/middleware.ts
```diff
@@ -140,6 +140,26 @@ async function serveSiteRoutes(requestURL: URL, request: NextRequest) {
...
+        const sites = requestURL.searchParams.get('s');
+        if (!orgs || !sites) {
+            return new Response('Missing required query parameters: o (orgs) and s (sites)', {
+                status: 400,
+                headers: { 'content-type': 'text/plain' },
+            });
+        }
+        const url = new URL(`${GITBOOK_API_URL}/v1/orgs/${orgs}/sites/${sites}/insights/events`);
```
SamyPesse: Why org**s** and site**s**? it's only one, it should not be plural

SamyPesse (approved):

PR #3576, opened 2025-08-20
Add `ready` handshake to integration assistants

zenoachtig:
Signal the integration is ready by calling `.ready()` on it. This way it only gets called upon when initialised.

```
const assistant = window.GitBook!.registerAssistant({
  label: 'My Assistant',
  icon: 'sparkle',
  open: (query?: string) => { /* ... */ },
});

// after init completes
assistant.ready();

// later, to unregister
assistant.dispose();
```


zenoachtig:
Closing (for now) as we're relying on the integration self-registering once it's ready

PR #3577, opened 2025-08-20
Fixes to search modal

zenoachtig:
Two small fixes after inspecting staging:
1. Recommended questions did not show up properly when an AI Assistant was defined
2. The `onClose` did not properly fire when closing the modal, causing `?q=` to remain in the URL.

### DIFF
--- a/packages/gitbook/src/components/Search/SearchContainer.tsx
+++ b/packages/gitbook/src/components/Search/SearchContainer.tsx
@@ -170,7 +170,7 @@ export function SearchContainer(props: SearchContainerProps) {
                 rootProps={{
                     open: state?.open ?? false,
                     onOpenChange: (open) => {
-                        open === true ? onOpen() : onClose();
+                        open ? onOpen() : onClose();
                     },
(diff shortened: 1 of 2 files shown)


packages/gitbook/src/components/Search/SearchContainer.tsx
```diff
@@ -170,7 +170,7 @@ export function SearchContainer(props: SearchContainerProps) {
                 rootProps={{
                     open: state?.open ?? false,
                     onOpenChange: (open) => {
-                        setSearchState((prev) => (prev ? { ...prev, open } : null));
+                        open === true ? onOpen() : onClose();
```
taranvohra: why `open === true` instead of just `open ? `

taranvohra (approved):
Let's add a changeset

packages/gitbook/src/components/Search/SearchContainer.tsx
zenoachtig: Good one

PR #3581, opened 2025-08-20
Fix pageaside height on pages with hero cover

zenoachtig:
This PR adds a check on the page cover to the TOC's height calculation script, which the pageaside also uses, to make sure the aside does not stretch beyond the screen when the page cover is present.

### Before

Screenshot 2025-08-21 at 10 25 39

### After

Screenshot 2025-08-21 at 10 26 01

### DIFF
--- a/packages/gitbook/src/components/PageBody/PageCover.tsx
+++ b/packages/gitbook/src/components/PageBody/PageCover.tsx
@@ -76,6 +76,8 @@ export async function PageCover(props: {
 
     return (
         <div
+            id="page-cover"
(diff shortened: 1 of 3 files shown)


packages/gitbook/src/components/PageBody/PageCover.tsx
```diff
@@ -76,6 +76,8 @@ export async function PageCover(props: {
 
     return (
         <div
+            id="page-cover"
+            data-full={as === 'full' ? 'true' : 'false'}
```
clairechabas: could do:
```suggestion
            data-full={String(as === 'full')}
```

PR #3584, opened 2025-08-22
Custom assistants followup

zenoachtig:
- Add `.close()` and `.dispose()` methods on assistant registration
- Clean up `?ask=` URL query param on `.close()`.
- Add `pageAction` attributes to internal Assistant object, dictating if and how the page action displays. Only enabled for the GitBook Assistant.

### DIFF
--- a/packages/browser-types/src/index.ts
+++ b/packages/browser-types/src/index.ts
@@ -71,7 +71,7 @@ export type GitBookGlobal = {
     /**
      * Register a custom assistant to be available on the site.
      */
-    registerAssistant: (assistant: GitBookAssistant) => () => void;
+    registerAssistant: (assistant: GitBookAssistant) => { close: () => void; dispose: () => void };
 };
 
 declare global {
--- a/packages/gitbook/src/components/AI/useAI.tsx
+++ b/packages/gitbook/src/components/AI/useAI.tsx
@@ -35,6 +35,16 @@ export type Assistant = Omit<GitBookAssistant, 'icon'> & {
     mode?: 'overlay' | 'sidebar' | 'search';
 
     /**
+     * Configuration for the in-page action to open the assistant.
+     * The button is only rendered if `ui` and `pageAction` are both set.
+     * Currently not available for custom assistants.
+     */
+    pageAction?: {
+        description: string;
+        open: (query?: string) => void;
+    };
+
+    /**
      * Icon of the assistant displayed in the UI.
      */
     icon: ReactNode;
@@ -89,6 +99,19 @@ export function useAI(): AIContext {
                 if (query) {
                     chatController.postMessage({ message: query });
                 }
+            },
+            pageAction: {
+                description: tString(
+                    language,
+                    'ai_chat_ask_about_page',
+                    getAIChatName(language, config.trademark)
+                ),
+                open: (query?: string) => {
+                    chatController.open();
+                    if (query) {
+                        chatController.postMessage({ message: query });
+                    }
+                },
             },
             ui: true,
             mode: 'sidebar',
--- a/packages/gitbook/src/components/AIActions/AIActions.tsx
+++ b/packages/gitbook/src/components/AIActions/AIActions.tsx
@@ -31,7 +31,7 @@ export function OpenAIAssistant(props: { assistant: Assistant; type: AIActionTyp
             icon={assistant.icon}
             label={assistant.label}
             shortLabel={tString(language, 'ask')}
-            description={tString(language, 'ai_chat_ask_about_page', assistant.label)}
+            description={assistant.pageAction?.description}
             disabled={chat.loading}
             onClick={() => {
                 assistant.open(tString(language, 'ai_chat_suggested_questions_about_this_page'));
--- a/.changeset/weak-kiwis-type.md
+++ b/.changeset/weak-kiwis-type.md
@@ -1,0 +1,6 @@
+---
+"@gitbook/browser-types": patch
+"gitbook": patch
+---
+
+Custom assistants followup
--- a/packages/gitbook/src/components/AIActions/AIActionsDropdown.tsx
+++ b/packages/gitbook/src/components/AIActions/AIActionsDropdown.tsx
@@ -25,7 +25,9 @@ interface AIActionsDropdownProps {
  */
 export function AIActionsDropdown(props: AIActionsDropdownProps) {
     const ref = useRef<HTMLDivElement>(null);
-    const assistants = useAI().assistants.filter((assistant) => assistant.ui === true);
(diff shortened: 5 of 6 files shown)


packages/browser-types/src/index.ts
```diff
@@ -71,7 +71,7 @@ export type GitBookGlobal = {
     /**
      * Register a custom assistant to be available on the site.
      */
-    registerAssistant: (assistant: GitBookAssistant) => () => void;
+    registerAssistant: (assistant: GitBookAssistant) => { close: () => void; dispose: () => void };
```
SamyPesse: Why is `close` defined in the result?

it should be a property on `GitBookAssistant` if the goal is to have a function to close the assistant

packages/gitbook/src/components/AI/useAI.tsx
```diff
@@ -34,6 +34,16 @@ export type Assistant = Omit<GitBookAssistant, 'icon'> & {
...
     mode?: 'overlay' | 'sidebar' | 'search';
 
+    /**
+     * Configuration for the in-page action to open the assistant.
+     * The button is only rendered if `ui` and `pageAction` are both set.
+     * Currently not available for custom assistants.
+     */
+    pageAction?: {
+        description: string;
+        open: (query?: string) => void;
```
SamyPesse: We already have an `open` on the assisant itself; why duplicate it?

packages/gitbook/src/components/AIActions/AIActions.tsx
```diff
@@ -31,7 +31,7 @@ export function OpenAIAssistant(props: { assistant: Assistant; type: AIActionTyp
             icon={assistant.icon}
             label={assistant.label}
             shortLabel={tString(language, 'ask')}
-            description={tString(language, 'ai_chat_ask_about_page', assistant.label)}
+            description={assistant.pageAction?.description}
```
SamyPesse: What happens if `assistant.pageAction` is not defined?

packages/gitbook/src/components/AIActions/AIActions.tsx
```diff
@@ -31,7 +31,7 @@ export function OpenAIAssistant(props: { assistant: Assistant; type: AIActionTyp
...
             disabled={chat.loading}
             onClick={() => {
                 assistant.open(tString(language, 'ai_chat_suggested_questions_about_this_page'));
```
SamyPesse: You are not using the `assistant.pageAction.open` here?

packages/gitbook/src/components/AIChat/AIChatButton.tsx
```diff
@@ -30,7 +32,15 @@ export function AIChatButton(props: {
                     <KeyboardShortcut keys={['mod', 'i']} className="border-tint-11 text-tint-1" />
                 </div>
             }
-            onClick={() => assistant.open()}
+            onClick={() => {
+                setSearchState((prev) => ({
```
SamyPesse: This looks messy; I think we are in a mess with state management and create more mess to get out of the mess.

If you explain what we want with the URL / state; I might find a solution

packages/gitbook/src/components/Search/useSearch.tsx
```diff
@@ -25,6 +25,21 @@ export type UpdateSearchState = (
     update: React.SetStateAction<SearchState | null>
 ) => Promise<URLSearchParams>;
 
+// Imperative setter reference to allow non-React code to update search state
+let externalSetSearchState: UpdateSearchState | null = null;
```
SamyPesse: Then just zustand .... we have a global state solution that e use in a bunch of places; why would we want to hardcode something like this?

packages/browser-types/src/index.ts
zenoachtig: It's not meant to close the assistant but the other way around — it's for the assistant back to the app that it closed, and we need to clean up the `?ask=` in the URL to keep it from opening again on page load.

SamyPesse: I don't understand what you are describing

> it's for the assistant back to the app that it closed

what does it mean?

packages/gitbook/src/components/AIChat/AIChatButton.tsx
zenoachtig: Sure. I think it's already much better than before, but very open to a refactor of this. Broadly speaking:

- `query`/`?q` reflects the state of the search input.
  - `?q=`: search input is open without a query. Displays recommended questions if there are assistants defined.
  - `?q=query`: search input has a query. Displays search results.
  - If present on load, the search box is opened with that query
- `ask`/`?ask` reflects the state of the (first) assistant.
  - `?ask=`: assistant is open without a query. Displays the empty chat or recommended questions in case of an assistant with `mode: search`.
  - `?ask=query`: assistant is open with `query` as its initial message/prompt.
  - If present on load, the assistant is opened with that query
  - `ask` will override `query` in terms of what gets opened by default.
- `global`/`?global` reflects the search scope, either 1 space or all spaces on the site.
- `open` reflects the state of search container. Not present in the URL since its state will change a lot as you interact with search/assistant. 

The reason it's complicated (I think) is because there are 3 different AI modes all using the same state, but they have different display logic in different places.

packages/gitbook/src/components/AI/useAI.tsx
zenoachtig: Following your [example in Slack](https://gitbook.slack.com/archives/C08RHFH8BMG/p1755849658186869?thread_ts=1755843029.785569&cid=C08RHFH8BMG), the AI might want to do something different if it gets passed a query with context. But could keep this just a simple boolean for now. 

The reason we need `pageAction` now is because all assistants use the "external" definition so we need a way to differentiate which assistants get added to the page menu and which don't. Adding this felt like a better thing to do than hardcoding a check on the GitBook Assistant's `id` or something. 

But you're right, currently the `open` does the exact same thing so it's probably useless. I'll change it to a boolean for now instead.

packages/gitbook/src/components/Search/useSearch.tsx
zenoachtig: That's a much better approach. Does zustand easily combine with nuqs?

Before AI Chat we used the URL as the state store, now we should transition to zustand and use the URL to save it across page loads. It's all so messy.

packages/browser-types/src/index.ts
zenoachtig: When the assistant closes itself, it's supposed to call `.close()` which signals back to us that we can discard the `ask` param from the state, since it's handled. This is the same behaviour we have on the AI chat, where `ask` is only set when the chat is open and removed when the chat is closed.

So `close` is returned from `registerAssistant` so that the integration can call it at a later time.

PR #3587, opened 2025-08-22
Turn on the "noUncheckedIndexedAccess" flag in all TS config files

utkuufuk:
`noUncheckedIndexedAccess` flag ensures that accessing elements in arrays or objects doesn't result in `undefined` or `null`, preventing a certain class of runtime errors. Without it, developers might overlook potential `null` or `undefined` values, leading to unexpected behaviors.

I _properly_ fixed existing errors that were obvious and easy to fix. I used `!` or the `@ts-expect-error` directive for others in order to avoid spending too much time on this.

The main idea is to avoid creating _more_ problems in the future as the compiler will now start warning us.


packages/gitbook/src/components/AI/server-actions/api.tsx
```diff
@@ -39,7 +39,9 @@ export async function streamRenderAIMessage(
 
             if (message.steps[stepIndex]) {
                 message.steps = [...message.steps];
+                // @ts-expect-error
```
SamyPesse, line 42: We can properly fix these ones

PR #3594, opened 2025-08-26
Support new object-fit property on individual card cover image

viktorrenkema:
This PR implements the GBO-side of allowing the object-fit property on cards. Table record fields for cover image now not only accept `api.ContentRefFile | api.ContentRefURL`, but also an object that contains record-level configurations like object-fit where then the content ref will be _nested_:

```
export type TableDefinitionImageRecord =
    | api.ContentRefFile
    | api.ContentRefURL
    | {
          ref: api.ContentRefFile | api.ContentRefURL;
          objectFit?: api.CardsImageObjectFit;
      };
```

If an object-fit is used that results in the background of the card being visible, we apply a subtle bg tint-color of the configured customization styles.

see https://github.com/GitbookIO/gitbook-x/pull/19221

## Preview

https://github.com/user-attachments/assets/3b2ab7a8-bd7b-4948-90bf-ba88a63d6b17

### DIFF
--- a/packages/gitbook/src/components/DocumentView/Table/RecordCard.tsx
+++ b/packages/gitbook/src/components/DocumentView/Table/RecordCard.tsx
@@ -3,7 +3,7 @@ import { Image } from '@/components/utils';
 import { type ResolvedContentRef, resolveContentRef } from '@/lib/references';
 import { tcls } from '@/lib/tailwind';
 import {
-    type CardsImageObjectFit,
+    CardsImageObjectFit,
     type ContentRef,
     type DocumentTableViewCards,
     SiteInsightsLinkPosition,
@@ -40,12 +40,8 @@ export async function RecordCard(
     const darkCoverIsSquareOrPortrait = isSquareOrPortrait(darkCover);
     const lightCoverIsSquareOrPortrait = isSquareOrPortrait(lightCover);
 
-    const darkObjectFit = dark.objectFit
-        ? `dark:${getObjectFitClass(dark.objectFit as CardsImageObjectFit)}`
-        : '';
-    const lightObjectFit = light.objectFit
-        ? getObjectFitClass(light.objectFit as CardsImageObjectFit)
-        : '';
+    const darkObjectFit = dark.objectFit ? `dark:${getObjectFitClass(dark.objectFit)}` : '';
+    const lightObjectFit = light.objectFit ? getObjectFitClass(light.objectFit) : '';
     const objectFits = `${lightObjectFit} ${darkObjectFit}`;
 
     const body = (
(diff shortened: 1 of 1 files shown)


packages/gitbook/src/components/DocumentView/Table/RecordCard.tsx
```diff
@@ -94,7 +98,12 @@ export async function RecordCard(
                         'min-w-0',
                         'w-full',
                         'h-full',
-                        'object-cover',
+                        'bg-tint-subtle',
+                        getObjectFitClass(light.objectFit),
+                        // Apply dark mode object-fit if different from light
+                        dark.objectFit && dark.objectFit !== light.objectFit
+                            ? `dark:${getObjectFitClass(dark.objectFit)}`
```
gregberge: You can't do that in TW, use plain names.

packages/gitbook/src/components/DocumentView/Table/RecordCard.tsx
```diff
@@ -190,3 +199,17 @@ function isSquareOrPortrait(contentRef: ResolvedContentRef | null) {
 
     return file.dimensions?.width / file.dimensions?.height <= 1;
 }
+
+/**
+ * Get the CSS class for object-fit based on the objectFit value.
+ */
+function getObjectFitClass(objectFit?: string): string {
```
gregberge: Is it not typed?

gregberge (changes requested):

packages/gitbook/src/components/DocumentView/Table/RecordCard.tsx
```diff
@@ -35,6 +40,14 @@ export async function RecordCard(
     const darkCoverIsSquareOrPortrait = isSquareOrPortrait(darkCover);
     const lightCoverIsSquareOrPortrait = isSquareOrPortrait(lightCover);
 
+    const darkObjectFit = dark.objectFit
+        ? `dark:${getObjectFitClass(dark.objectFit as CardsImageObjectFit)}`
```
gregberge: Still not OK for Tailwind class.
If you use as, then it's untyped, typing is useless here. It should be typed at API level, it it's not then just pass a string here.

packages/gitbook/src/components/DocumentView/Table/RecordCard.tsx
```diff
@@ -190,3 +204,19 @@ function isSquareOrPortrait(contentRef: ResolvedContentRef | null) {
...
+        case 'contain':
+            return 'object-contain';
+        case 'fill':
+            return 'object-fill';
+        case 'cover':
+            return 'object-cover';
+        default:
+            throw new Error(`Unsupported object fit: ${objectFit}`);
+    }
+}
```
gregberge: Safer like you did before, if we pass an unsupported thing here we don't want to crash the app.

viktorrenkema: Fixed, now using the api values

PR #3595, opened 2025-08-26
Fix Search results are not clickable

zenoachtig:
The root cause for this was that we instantiated two search popovers on sites without a header, since these sites still have a mobile header with its own search popover in addition to the sidebar search popover.

Add a `viewport` check to only render the popover when appropriate.

Also cleans up the mobile sidebar to not repeat the logo & search button unncessarily.

### DIFF
--- a/packages/gitbook/src/components/Search/SearchContainer.tsx
+++ b/packages/gitbook/src/components/Search/SearchContainer.tsx
@@ -23,14 +23,14 @@ interface SearchContainerProps {
     isMultiVariants: boolean;
     style: CustomizationSearchStyle;
     className?: string;
-    visibility?: 'desktop' | 'mobile';
+    viewport?: 'desktop' | 'mobile';
 }
 
 /**
  * Client component to render the search input and results.
  */
 export function SearchContainer(props: SearchContainerProps) {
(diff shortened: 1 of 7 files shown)


packages/gitbook/src/components/Search/SearchContainer.tsx
```diff
@@ -23,13 +23,14 @@ interface SearchContainerProps {
     isMultiVariants: boolean;
     style: CustomizationSearchStyle;
     className?: string;
+    visibility?: 'desktop' | 'mobile';
```
taranvohra: Should we choose a different name for this prop? Maybe `viewport`? Visibility I'd related with something that can be shown/hidden

packages/gitbook/src/components/Search/SearchContainer.tsx
```diff
@@ -145,6 +146,12 @@ export function SearchContainer(props: SearchContainerProps) {
 
     const showAsk = withSearchAI && normalizedAsk; // withSearchAI && normalizedAsk;
 
+    const visible = React.useMemo(() => {
```
taranvohra: `useMemo` is unnecessary in this case

taranvohra (approved):

PR #3597, opened 2025-08-26
Fix content min-height with sections

zenoachtig:
Previously we always set the content to have a `min-height` of `calc(100vh-64px)` to subtract the header. With sections the header can be larger, so we need to subtract more.

### DIFF
--- a/packages/gitbook/src/components/SpaceLayout/SpaceLayout.tsx
+++ b/packages/gitbook/src/components/SpaceLayout/SpaceLayout.tsx
@@ -109,7 +109,10 @@ export function SpaceLayout(props: {
                                     'site-width-wide:max-w-full',
 
                                     // Ensure the footer is display below the viewport even if the content is not enough
-                                    withFooter && 'min-h-[calc(100vh-64px)]',
+                                    withFooter && [
+                                        'site-header:min-h-[calc(100vh-64px)]',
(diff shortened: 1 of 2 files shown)


packages/gitbook/src/components/SpaceLayout/SpaceLayout.tsx
```diff
@@ -109,7 +109,10 @@ export function SpaceLayout(props: {
                                     'site-width-wide:max-w-full',
 
                                     // Ensure the footer is display below the viewport even if the content is not enough
-                                    withFooter && 'min-h-[calc(100vh-64px)]',
+                                    withFooter && [
+                                        'site-header:min-h-[calc(100vh-64px)]',
```
gregberge, line 113: site-header represents the "without sections" right?

gregberge (approved):

packages/gitbook/src/components/SpaceLayout/SpaceLayout.tsx
zenoachtig: Yep!

gregberge: Not very intuitive because I expect to have it even if there is sections. We should replace it by `site-header-no-sections` probably but not the scope of this PR.

PR #3601, opened 2025-08-27
Fix corner radius of active section on "Line" sidebar style

zenoachtig:
Before

Screenshot 2025-08-27 at 10 10 53

After

Screenshot 2025-08-27 at 10 11 07

### DIFF
--- a/packages/gitbook/src/components/PageAside/AsideSectionHighlight.tsx
+++ b/packages/gitbook/src/components/PageAside/AsideSectionHighlight.tsx
@@ -25,7 +25,7 @@ export const AsideSectionHighlight = React.memo(
                     'rounded-md',
                     'straight-corners:rounded-none',
                     'circular-corners:rounded-2xl',
-                    'sidebar-list-line:rounded-l-none',
+                    'sidebar-list-line:rounded-l-none!',
 
                     'sidebar-list-pill:bg-primary',
                     '[html.theme-muted.sidebar-list-pill_&]:bg-primary-hover',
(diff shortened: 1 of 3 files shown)


packages/gitbook/src/components/PageAside/AsideSectionHighlight.tsx
```diff
@@ -25,7 +25,7 @@ export const AsideSectionHighlight = React.memo(
                     'rounded-md',
                     'straight-corners:rounded-none',
                     'circular-corners:rounded-2xl',
-                    'sidebar-list-line:rounded-l-none',
+                    'sidebar-list-line:rounded-l-none!',
```
gregberge, line 28: We already use tw-merge, using important also is maybe a bit too much. Could we fix it in another way?

zenoachtig: The problem is that even with tw-merge, `sidebar-list-line` is overridden by `circular-corners`. The only way is to explicitly add `circular-corners:sidebar-list-line:rounded-l-none` or something like that. Instead it felt cleaner to basically say "no matter what, `sidebar-list-line` should ALWAYS have these corners flat"..

But happy to change if the explicit declarations seem more clear!

gregberge: OK It's just that it's becoming a bit messy but I see the problem. Let's go with that.

Note that I think the variant order is important, maybe there is something to do at TW config level: https://github.com/GitbookIO/gitbook/blob/729921f33803441e6b8f84d758a29ee433c134b4/packages/gitbook/tailwind.config.ts#L549-L570

PR #3602, opened 2025-08-27
Catch OpenAPI validation errors

nolannbiron:
This pull request improves error handling in the OpenAPI parser by ensuring validation errors are properly caught and reported. The main change is to wrap the OpenAPI validation step in a try-catch block, so that any errors are thrown as a custom `OpenAPIParseError` with relevant details.

### DIFF
--- a/packages/openapi-parser/src/v3.ts
+++ b/packages/openapi-parser/src/v3.ts
@@ -10,7 +10,14 @@ import type { ParseOpenAPIInput, ParseOpenAPIResult } from './parse';
  */
 export async function parseOpenAPIV3(input: ParseOpenAPIInput): Promise<ParseOpenAPIResult> {
     const { value, rootURL, options = {} } = input;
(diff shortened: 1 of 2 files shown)


packages/openapi-parser/src/v3.ts
```diff
@@ -10,7 +10,16 @@ import type { ParseOpenAPIInput, ParseOpenAPIResult } from './parse';
...
+
+    const result = await validate(value).catch((error) => {
+        throw new OpenAPIParseError(
+            error instanceof Error ? error.message : 'Invalid OpenAPI document',
+            {
+                code: 'invalid',
+                rootURL,
+            }
+        );
+    });
```
gregberge: Can you please use a proper error message (and not pick the one from the parser) and use the "cause" property to chain the error please?

nolannbiron: done

PR #3603, opened 2025-08-27
Assistant hold message

zenoachtig:
Show a client-side message (so it can be easily translated and show up even in poor network conditions) after the AI thinks for >5 seconds. A second message appears after 10 seconds.

https://github.com/user-attachments/assets/4f04a385-4ac4-4e22-8e8e-5033eed313a4

### DIFF
--- a/packages/gitbook/src/components/AIChat/AIChatMessages.tsx
+++ b/packages/gitbook/src/components/AIChat/AIChatMessages.tsx
@@ -14,7 +14,6 @@ export function AIChatMessages(props: {
     lastUserMessageRef?: React.RefObject<HTMLDivElement>;
 }) {
     const { chat, chatController, lastUserMessageRef } = props;
-    const language = useLanguage();
 
     return (
         <>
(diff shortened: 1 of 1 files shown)


packages/gitbook/src/components/AIChat/AIChatMessages.tsx
```diff
@@ -46,17 +49,49 @@ export function AIChatMessages(props: {
...
-                                        className="h-4 animate-[fadeIn_500ms_ease_both,pulse_1.5s_infinite] circular-corners:rounded-2xl rounded-corners:rounded-md bg-tint-4"
-                                        style={{
-                                            width: `calc(${(index % 4) * 20 + 10}% - 4px)`,
-                                            animationDelay: `${index * 0.1}s`,
-                                        }}
-                                    />
-                                ))}
+                            <div className="flex w-full animate-fade-in-slow flex-col gap-2">
+                                {!message.content ? (
+                                    <div className="animate-[heightIn_500ms_4500ms_ease_both] py-2 text-tint-subtle">
```
SamyPesse: We should split a bit more into components; it'll make things easier to read compared to one big component.

For example this whole thing could be `{message.content ? null : <HoldMessage />}`

PR #3607, opened 2025-08-27
Adapt OpenAPI block rendering to evaluate adaptive exprs and prefill TryIt configuration

spastorelli:


packages/gitbook/src/components/Adaptive/AdaptiveVisitorContextProvider.tsx
```diff
@@ -0,0 +1,86 @@
+'use client';
+
+import type { GitBookSiteContext } from '@/lib/context';
+import { ExpressionRuntime, parseTemplate } from '@gitbook/expr';
+import { OpenAPITryItPrefillContextProvider } from '@gitbook/react-openapi';
+import type React from 'react';
+import { createContext, useCallback, useContext, useEffect, useRef, useState } from 'react';
```
SamyPesse: I usually prefer to prefix with React (`createContext` -> `React.createContext`) for clarity

packages/gitbook/src/components/Adaptive/AdaptiveVisitorContextProvider.tsx
```diff
@@ -0,0 +1,86 @@
...
+};
+
+const AdaptiveVisitorContext = createContext<{ visitorClaims: VisitorClaimsData } | null>(null);
+
+/**
+ * Provide context to adapt site based on visitor claims.
+ */
+export function AdaptiveVisitorContextProvider(
+    props: React.PropsWithChildren<{
+        getVisitorClaimsUrl: string;
```
SamyPesse: The name seems to indicate that it's a function:


```suggestion
        visitorClaimsURL: string;
```

packages/gitbook/src/components/Adaptive/AdaptiveVisitorContextProvider.tsx
```diff
@@ -0,0 +1,86 @@
...
+        contextId: GitBookSiteContext['contextId'];
+    }>
+) {
+    const { getVisitorClaimsUrl, contextId, children } = props;
+
+    const runtime = useRef<ExpressionRuntime>(new ExpressionRuntime());
+    const [visitorClaimsData, setVisitorClaimsData] = useState<VisitorClaimsData | null>(null);
```
SamyPesse: Instead of using a state and a context; you could also use `zustand` for this

packages/gitbook/src/components/Adaptive/AdaptiveVisitorContextProvider.tsx
```diff
@@ -0,0 +1,86 @@
...
+                    return;
+                }
+                const data = await res.json<VisitorClaimsData>();
+                setVisitorClaimsData(data);
+            } catch (err) {
+                console.error('Error fetching visitor claims data', err);
+                setVisitorClaimsData(null);
+            }
+        }
+        fetchVisitorData();
```
SamyPesse: When doing an async effect like this, you should handle the cancellation; otherwise you might have bug where you'll set the state with a previous value

```
let cancelled = false;

return () => {
  cancelled = true;
}

```

You can even use an AbortSignal for the fetch

packages/gitbook/src/components/SiteLayout/SiteLayout.tsx
```diff
@@ -45,6 +46,10 @@ export async function SiteLayout(props: {
         });
     });
 
+    const getVisitorClaimsUrl = new URL(
```
SamyPesse: Why are you doing a `new URL` instead of just `toAbsoluteURL`; as you are doing toString after

packages/gitbook/src/components/Adaptive/AdaptiveVisitorContextProvider.tsx
```diff
@@ -0,0 +1,86 @@
...
+    const [visitorClaimsData, setVisitorClaimsData] = useState<VisitorClaimsData | null>(null);
+
+    useEffect(() => {
+        // Only fetch visitor claims if contextId is defined (adaptive content site).
+        if (!contextId) {
+            setVisitorClaimsData(null);
+            return;
+        }
+
+        async function fetchVisitorData() {
```
SamyPesse: This approach means that EVERY page view is going to create one more request to our middleware. At our scale, this is meaningful.

We might want to revisit this strategy with a smarter hook `useVisitorClaims(): Claims | null` which use a global zustand state and does the fetch only when the hook is used 

Then we need to find a way so that the OpenAPI try it can lazily use it.

99% of the content are not going to use try-it / adaptive content / dynamic expression.

packages/gitbook/src/components/Adaptive/AdaptiveVisitorContextProvider.tsx
```diff
@@ -0,0 +1,86 @@
...
+            <OpenAPITryItPrefillContextProvider
+                resolveTryItPrefillExpression={resolveTryItPrefillExpression}
+            >
+                {children}
+            </OpenAPITryItPrefillContextProvider>
+        </AdaptiveVisitorContext.Provider>
+    );
+}
+
+export function useAdaptiveVisitorContext() {
```
SamyPesse: I think it's better to name with something a bit direct; it's not because it uses a context that it should be suffixed with `Context`


```suggestion
export function useAdaptiveVisitor() {
```

packages/react-openapi/src/util/tryit-prefill.test.ts
```diff
@@ -0,0 +1,358 @@
...
+                method: 'GET',
+                operation: { summary: 'Get space by ID' },
+                servers: [{ url: 'https://api.gitbook.com/v1/' }],
+                securities: [
+                    [
+                        'apiTokenScheme',
+                        {
+                            type: 'http',
+                            scheme: 'bearer',
+                            'x-gitbook-tryit-prefill': '{{ visitor.claims.apiToken }}',
```
SamyPesse: This will be used for more than just try it (ex: code sample).


```suggestion
                            'x-gitbook-prefill': '{{ visitor.claims.apiToken }}',
```

packages/react-openapi/src/OpenAPITryItPrefillContext.tsx
```diff
@@ -0,0 +1,41 @@
+'use client';
+
+import { createContext, useContext, useMemo } from 'react';
+import type { TryItPrefillExpressionResolver } from './util/tryit-prefill';
+
+interface OpenAPITryItPrefillContextValue {
+    resolveTryItPrefillExpression: TryItPrefillExpressionResolver;
```
SamyPesse: Why not just take a `@gitbook/expr` context as prop?

packages/react-openapi/src/util/tryit-prefill.test.ts
```diff
@@ -0,0 +1,358 @@
...
+function createTryItPrefillResolverWithVisitorData(
+    visitorData: Record<string, unknown>
+): TryItPrefillExpressionResolver {
+    return (expr: string) => {
+        const parts = parseTemplate(expr);
+        if (!parts.length) {
+            return undefined;
+        }
+
+        return expressionRuntime.evaluateTemplate(expr, visitorData);
```
SamyPesse: This seems more complex than needed; imo `react-openapi` should just accept a `expressionContext` (which in that case here equals `visitorData`) and take care of parsing/evaluating.

It's fine for `react-openapi` to depends on `@gitbook/expr`; it should just be generic on the data used to evaluate

packages/react-openapi/src/OpenAPITryItPrefillContext.tsx
SamyPesse, line 1: We already have `context.tsx` and `OpenAPIContext`; let's just use this and add `expressionContext` in it

packages/gitbook/src/components/SiteLayout/SiteLayout.tsx
```diff
@@ -54,18 +58,23 @@ export async function SiteLayout(props: {
             }
             externalLinksTarget={customization.externalLinks.target}
         >
-            <AIContextProvider
-                aiMode={customization.ai?.mode}
-                trademark={customization.trademark.enabled}
+            <AdaptiveVisitorContextProvider
```
SamyPesse: Maybe it should be moved into `SpaceLayout` so that it does also apply to the new embedded version

packages/gitbook/src/components/Adaptive/visitor-store.ts
SamyPesse, line 1: Why use a zustand store instead of a global variable?

You don't actually "listen" to the value anywhere, so we don't need a "state".

Also it looks like you store only one promise/data globally, but the context ID can change, you might want to use a unique promise/data per context ID.

packages/gitbook/src/lib/visitors.ts
```diff
@@ -443,3 +443,45 @@ export function getVisitorAuthCookieMaxAge(decoded: JwtPayload): number {
 
     return defaultMaxAge;
 }
+
+/**
+ * Handler for the /~gitbook/visitor middleware route to expose visitor data.
+ */
+export function serveVisitorClaimsDataRequest(request: NextRequest, siteRequestURL: URL) {
```
SamyPesse, line 450: What is your thinking on the CORS for this route?
We should probably restrict it to the same domain.

packages/openapi-parser/src/types.ts
```diff
@@ -70,6 +70,13 @@ export interface OpenAPICustomOperationProperties {
     'x-stability'?: OpenAPIStability;
 }
 
+/**
+ * Custom properties that can be defined to enable prefilling for the TryIt functionality.
+ */
+export interface OpenAPICustomTryItPrefillProperties {
+    'x-gitbook-prefill'?: string;
```
SamyPesse, line 77: @talagluck @addisonschultz what are your thoughts on this special property name to represent the expression to be evaluated to get the auth value?

@spastorelli let's not block on this to merge; but good to get feedback before we start documenting it, etc

packages/react-openapi/src/util/tryit-prefill.test.ts
```diff
@@ -0,0 +1,311 @@
...
+
+            const result = resolveTryItPrefillForOperation({
+                operation,
+                prefillInputContext,
+            });
+
+            expect(result).toEqual({
+                authentication: {
+                    securitySchemes: {
+                        apiTokenScheme: { token: 'gb_api_testToken' },
```
SamyPesse, line 38: Why is the key `token`? is it always a token?

Why not `value`?

packages/react-openapi/src/util/tryit-prefill.test.ts
```diff
@@ -0,0 +1,311 @@
...
+                            type: 'http',
+                            scheme: 'basic',
+                            'x-gitbook-prefill': '{{ visitor.claims.basicAuth }}',
+                        },
+                    ],
+                ],
+            };
+
+            const prefillInputContext: PrefillInputContextData = {
+                visitor: { claims: { basicAuth: 'testuser:testpassword' } },
```
SamyPesse, line 63: I'm confused by what is being done here 🤔  so we parse it?

why is it not:


```suggestion
                visitor: { claims: { basicAuth: { username: 'testuser', password: 'testpassword' } } },
```

packages/react-openapi/src/OpenAPIPrefillContextProvider.tsx
SamyPesse, line 1: I think names in this file are not very obvious; and there is 0 comments; making it quite hard to follow without good context

packages/react-openapi/src/OpenAPIPrefillContextProvider.tsx
```diff
@@ -0,0 +1,26 @@
+'use client';
+
+import * as React from 'react';
+
+export type PrefillInputContextData = Record<string, unknown>;
+
+type PrefillContextValue = () => Promise<PrefillInputContextData | null>;
```
SamyPesse: We should at least explain why it's a function returning a promise

and what it's supposed to return

packages/react-openapi/src/ScalarApiButton.tsx
```diff
@@ -6,30 +6,42 @@ import { createPortal } from 'react-dom';
...
+        null
+    );
     const controllerRef = useRef<ScalarModalControllerRef>(null);
+
+    // Fetch visitor data and open modal
+    const openModal = async () => {
+        const data = await getPrefillInputContextData();
+        setPrefillInputContext(data);
+        controllerRef.current?.openClient?.();
+        setIsOpen(true);
```
SamyPesse: I'm not a huge fan of this approach; why not do it in `ScalarModal` , actually suspends on the promise using `React.use()` (BTW to be check if we can use it in GBX) ?

packages/react-openapi/src/util/tryit-prefill.test.ts
spastorelli: Because that's what Scalar modal expect to prefill when the auth scheme is a of type `bearer`:
https://guides.scalar.com/scalar/scalar-api-references/configuration

The shape of the expected prefill config auth changes depending on the auth scheme.

packages/react-openapi/src/util/tryit-prefill.test.ts
spastorelli: Yes we parse the value after evaluating the `x-gitbook-prefill` so we can get the username and password. So in this case `{{ visitor.claims.basicAuth }}` would be after evaluation `testuser:testpassword` so in this case we need to extract the username and password as that's what Scalar modal prefill config expects for basic auth scheme:
https://guides.scalar.com/scalar/scalar-api-references/configuration

packages/gitbook/src/lib/visitors.ts
spastorelli: We could add CORS but even without it we would not be leaking info I think since the cookies sent with the request will be only the ones from the same domain.

packages/openapi-parser/src/types.ts
talagluck: Is this used anywhere other than the Test It button? I think it would be clearer to call the property `x-gitbook-testit-prefill` to indicate where this will be used. 

Also, since the button we use says Test it, we may want to consider updating this class name to `OpenAPICustomTestItPrefillProperties` for consistency.

packages/react-openapi/src/ScalarApiButton.tsx
spastorelli: I can't seem to be able to use `React.use` to suspend on the promise because it's a client component and getting this error if I do:
```
A component was suspended by an uncached promise. Creating promises inside a Client Component or hook is not yet supported, except via a Suspense-compatible library or framework.
```
The other options I see is to call the promise in an effect in ScalarModal or use a "Suspense-compatible" lib maybe something like tanstack query or `useSWR`.

packages/gitbook/src/components/Adaptive/AdaptiveVisitorContextProvider.tsx
```diff
@@ -0,0 +1,100 @@
...
+>();
+
+/**
+ * Factory that creates a function to fetch visitor claims
+ * for a given endpoint URL and contextId so it can be exposed as adaptive visitor context.
+ *
+ * The returned function caches results per context ID to avoids duplicate
+ * requests to fetch the claims when the context ID doesn't change.
+ */
+async function getAdaptiveVisitorClaimsFactory(
```
SamyPesse: The comment and naming is wrong; it's not a factorey, it just returns the value, not a function.

packages/gitbook/src/components/Adaptive/AdaptiveVisitorContextProvider.tsx
```diff
@@ -0,0 +1,100 @@
...
+    url: string,
+    contextId?: string
```
SamyPesse: You should avoid optional argument when these are not options.

It's easy to miss it and endup with considering it's `undefined`.


```suggestion
    contextId: string | undefined
```

packages/gitbook/src/components/Adaptive/AdaptiveVisitorContextProvider.tsx
```diff
@@ -0,0 +1,100 @@
...
+): Promise<AdaptiveVisitorClaimsData | null> {
+    // Only fetch visitor claims if contextId is defined (adaptive content site).
+    if (!contextId) {
+        return null;
+    }
```
SamyPesse: Skip this in the parent function then

packages/gitbook/src/components/Adaptive/AdaptiveVisitorContextProvider.tsx
```diff
@@ -0,0 +1,100 @@
...
+    return (
+        <AdaptiveVisitorContext.Provider value={getAdaptiveVisitorClaims}>
+            <OpenAPIPrefillContextProvider getPrefillInputContextData={getAdaptiveVisitorClaims}>
+                {children}
+            </OpenAPIPrefillContextProvider>
+        </AdaptiveVisitorContext.Provider>
+    );
+}
+
+export function useAdaptiveVisitor() {
```
SamyPesse: It'll be used in a bunch of place, it deserves a comment and a type

packages/react-openapi/src/ScalarApiButton.tsx
SamyPesse: Even if you can't use `React.use` (which is fine because not support in the react version we use in GBX).

We can do it in the component render with:

```
const data = getData()
if (data insteanceof Promise) {
  throw data;
}
```

But the current approach here is not solid at all; subect to race condition

packages/react-openapi/src/ScalarApiButton.tsx
```diff
@@ -57,12 +73,25 @@ export function ScalarApiButton(props: {
...
+    securities: OpenAPIOperationData['securities'];
+    servers: OpenAPIOperationData['servers'];
     specUrl: string;
     controllerRef: React.Ref<ScalarModalControllerRef>;
+    prefillInputContext: Record<string, unknown>;
 }) {
-    const { method, path, specUrl, controllerRef } = props;
+    const { method, path, securities, servers, specUrl, controllerRef, prefillInputContext } =
+        props;
+
```
SamyPesse, line 79: You need to do the fetch of claims and suspend here:

```tsx
const prefillInputContext = await getPrefillInputContextData();
if (prefillInputContext instanceof Promise) {
  throw prefillInputContext;
}
```

packages/react-openapi/src/types.ts
```diff
@@ -1,17 +1,28 @@
...
     OpenAPICustomSpecProperties,
     OpenAPIV3,
 } from '@gitbook/openapi-parser';
 
-export type OpenAPISecurityWithRequired = OpenAPIV3.SecuritySchemeObject & { required?: boolean };
+export type OpenAPIServerVariableWithCustomProperties = OpenAPIV3.ServerVariableObject &
+    OpenAPICustomPrefillProperties;
+
+export type OpenAPIServerWithCustomProperties = Omit<OpenAPIV3.ServerObject, 'variables'> & {
+    variables?: {
```
SamyPesse, line 15: What is this?

Can you comment to explain? to make sure future engs looking at this are not totally lost

PR #3609, opened 2025-08-28
Support bold in headings

gregberge:


SamyPesse (approved):

zenoachtig (changes requested):
- Headings should not use `900/Black`/`font-black` but instead use `700/Bold`/`font-bold`.
- We should add `600/SemiBold` to fonts that don't have it by default (Roboto, ABCFavorit).

gregberge:
@zenoachtig for ABC the variable font supports it, that's enough

PR #3623, opened 2025-09-04
Fix error on parameters being an object instead of an array

conico974:
Fix an error for OpenAPI where `operation.parameters` is incorrectly provided by the validate function from scalar.
Parameters are supposed to be an Array, but for some invalid schema, it will be an object causing the rendering to crash.

### DIFF
--- a/packages/react-openapi/src/OpenAPICodeSample.tsx
+++ b/packages/react-openapi/src/OpenAPICodeSample.tsx
@@ -66,7 +66,11 @@ function generateCodeSamples(props: {
     const searchParams = new URLSearchParams();
     const headersObject: { [k: string]: string } = {};
 
-    data.operation.parameters?.forEach((param) => {
+    // The parser can sometimes returns invalid parameters (an object instead of an array).
+    // It should get fixed in scalar, but in the meantime we just ignore the parameters in that case.
+    const params = Array.isArray(data.operation.parameters) ? data.operation.parameters : [];
+
+    params.forEach((param) => {
         if (!param) {
             return;
         }
--- a/packages/react-openapi/src/OpenAPISpec.tsx
+++ b/packages/react-openapi/src/OpenAPISpec.tsx
@@ -117,19 +117,22 @@ function getParameterGroupName(paramIn: string, context: OpenAPIClientContext): 
 /** Deduplicate parameters by name and in.
  * Some specs have both parameters define at path and operation level.
  * We only want to display one of them.
+ * Parameters can have the wrong type (object instead of array) sometimes, we just return an empty array in that case.
  */
 function deduplicateParameters(parameters: OpenAPI.Parameters): OpenAPI.Parameters {
     const seen = new Set();
 
-    return parameters.filter((param) => {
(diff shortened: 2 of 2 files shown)


packages/react-openapi/src/OpenAPICodeSample.tsx
```diff
@@ -66,7 +66,11 @@ function generateCodeSamples(props: {
     const searchParams = new URLSearchParams();
     const headersObject: { [k: string]: string } = {};
 
-    data.operation.parameters?.forEach((param) => {
+    // The parser can sometimes returns invalid parameters (an object instead of an array).
+    // It should get fixed in scalar, but in the meantime we just ignore the parameters in that case.
+    const params = Array.isArray(data.operation.parameters) ? data.operation.parameters : [];
```
emmerich, line 71: Any way of turning the object into an array? So the user doesn't see empty params

packages/react-openapi/src/OpenAPISpec.tsx
```diff
@@ -117,19 +117,22 @@ function getParameterGroupName(paramIn: string, context: OpenAPIClientContext):
...
+              }
 
-        seen.add(key);
+              seen.add(key);
 
-        return true;
-    });
+              return true;
+          })
+        : [];
```
emmerich, line 137: Same here

packages/react-openapi/src/OpenAPICodeSample.tsx
conico974: We could try to infer something, but this is invalid OpenAPI spec. Inferring something might be worse than just ignoring them.

gregberge, line 71:
```suggestion
    // Ignore wrong reference
    const params = checkIsReference(data.operation.parameters) ? [] : data.operation.parameters;
```

packages/react-openapi/src/OpenAPISpec.tsx
```diff
@@ -117,19 +117,22 @@ function getParameterGroupName(paramIn: string, context: OpenAPIClientContext):
...
  * Some specs have both parameters define at path and operation level.
  * We only want to display one of them.
+ * Parameters can have the wrong type (object instead of array) sometimes, we just return an empty array in that case.
  */
 function deduplicateParameters(parameters: OpenAPI.Parameters): OpenAPI.Parameters {
     const seen = new Set();
 
-    return parameters.filter((param) => {
-        const key = `${param.name}:${param.in}`;
+    return Array.isArray(parameters)
```
gregberge, line 125: Just do:

```ts
if (checkIsReference(parameters) {
  return []
}
```

packages/react-openapi/src/OpenAPICodeSample.tsx
gregberge: That's a reference, we don't want to convert it. I proposed the correct way of handling that.

conico974: Just checked that, it's not only for ref, we have to check if it's an array or not

PR #3630, opened 2025-09-08
Better handling for external link "mailto:" in Hovered Card in GBO

zenoachtig:
CleanShot 2025-09-08 at 12 06 22@2x

### DIFF
--- a/packages/gitbook/src/components/DocumentView/InlineLink/InlineLink.tsx
+++ b/packages/gitbook/src/components/DocumentView/InlineLink/InlineLink.tsx
@@ -46,6 +46,7 @@ export async function InlineLink(props: InlineProps<DocumentInlineLink>) {
         );
     }
     const isExternal = inline.data.ref.kind === 'url';
+    const isMailto = resolved.href.startsWith('mailto:');
     const content = (
         <StyledLink
             href={resolved.href}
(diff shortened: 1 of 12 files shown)


packages/gitbook/src/components/DocumentView/InlineLink/InlineLink.tsx
```diff
@@ -63,7 +64,13 @@ export async function InlineLink(props: InlineProps<DocumentInlineLink>) {
                 nodes={inline.nodes}
                 ancestorInlines={[...ancestorInlines, inline]}
             />
-            {isExternal ? (
+            {isMailto ? (
```
gregberge, line 67: `isMailto ? ... : isExternal ? ... : ...`

packages/gitbook/src/components/DocumentView/InlineLink/InlineLink.tsx
```diff
@@ -105,6 +113,14 @@ function InlineLinkTooltipWrapper(props: {
...
         ];
     }
+    if (isMailto) {
+        resolved.text = resolved.text.split('mailto:')[1] ?? resolved.text;
+        breadcrumbs = [
+            {
+                label: tString(language, 'link_tooltip_email'),
+            },
+        ];
+    }
```
gregberge: Same better to do:

```ts
if (isMailto) {
  ...
} else if (isExternal) {
  ...
}
```

It makes it clearer that the two options are exclusive. You could also do it for `isSamePage`.

PR #3635, opened 2025-09-09
Don't show toolbar when rendering within GitBook app preview

viktorrenkema:
When a preview link for a change-request or revision is opened, we show a toolbar with helpful info and actions. However, we don't want to show this toolbar if the site is rendered in an iframe in the GitBook web-app editor, under the preview tab.

This behaviour likely broke when the 'multi-id' functionality became obsolete with the move to gbo v2.
- this pr removes that old logic
- and replaces it with a new check using the existing `isPreviewRequest` util

## Examples
Toolbar no longer shows when previewing in-app:
CleanShot 2025-09-09 at 11 28 24@2x

Toolbar still shows when visiting a change-request or revision preview link:
CleanShot 2025-09-09 at 11 32 20@2x

### DIFF
--- a/packages/gitbook/src/components/AdminToolbar/IframeWrapper.tsx
+++ b/packages/gitbook/src/components/AdminToolbar/IframeWrapper.tsx
@@ -1,0 +1,27 @@
+'use client';
+
+import React from 'react';
+
+interface IframeWrapperProps {
+    children: React.ReactNode;
+}
+
+/**
+ * Client component that detects if we're in an iframe and conditionally renders children
+ */
+export function IframeWrapper({ children }: IframeWrapperProps) {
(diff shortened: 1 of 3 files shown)


packages/gitbook/src/components/AdminToolbar/IframeWrapper.tsx
```diff
@@ -0,0 +1,27 @@
...
+        const inIframe = window !== window.parent;
+        setIsInIframe(inIframe);
+    }, []);
+
+    // Don't render children if we're in an iframe (GitBook app preview)
+    if (isInIframe) {
+        return null;
+    }
+
+    return <>{children}</>;
```
gregberge: You don't need the fragment here

gregberge (approved):

PR #3637, opened 2025-09-10
RND-7985: accessible search

BrettJephson:
Fixes some issues around search accessibility:

1.  Search results did not make sense and were overly verbose, I’ve simplified to “Page with title ‘some title’” and “Section with title ‘some title’” and abbreviated the content in a result that returns a section body so it doesn’t read like War and Peace. With translations provided by **Claude**.
2. When search results are returned or no results are returned we actually notify the screen reader user
3. We let the screen reader know that this is a combobox with an associated list of results
4. We surface number of results and which is the selected result

### DIFF
--- a/.changeset/selfish-ravens-drop.md
+++ b/.changeset/selfish-ravens-drop.md
@@ -1,0 +1,5 @@
+---
+"gitbook": patch
+---
+
+Expose "Best match" when site has translations
--- a/packages/gitbook/src/components/Header/Header.tsx
+++ b/packages/gitbook/src/components/Header/Header.tsx
@@ -124,10 +124,7 @@ export function Header(props: {
                                 withSiteVariants={
                                     sections?.list.some(
                                         (s) =>
-                                            s.object === 'site-section' &&
-                                            s.siteSpaces.filter(
-                                                (s) => s.space.language === siteSpace.space.language
-                                            ).length > 1
+                                            s.object === 'site-section' && s.siteSpaces.length > 1
                                     ) ?? false
                                 }
                                 withSections={!!sections}
--- a/packages/gitbook/src/components/SpaceLayout/SpaceLayout.tsx
+++ b/packages/gitbook/src/components/SpaceLayout/SpaceLayout.tsx
@@ -185,11 +185,7 @@ export function SpaceLayout(props: SpaceLayoutProps) {
                                                 sections?.list.some(
                                                     (s) =>
                                                         s.object === 'site-section' &&
-                                                        s.siteSpaces.filter(
-                                                            (s) =>
-                                                                s.space.language ===
-                                                                siteSpace.space.language
-                                                        ).length > 1
+                                                        s.siteSpaces.length > 1
                                                 ) ?? false
                                             }
                                             withSections={withSections}


packages/gitbook/src/components/Search/SearchContainer.tsx
```diff
@@ -223,6 +237,10 @@ export function SearchContainer(props: SearchContainerProps) {
                     withAI={withSearchAI}
                     isOpen={state?.open ?? false}
                     className={className}
+                    resultsShowing={resultsState.showing}
+                    resultsCount={resultsState.count}
+                    cursor={resultsState.cursor}
+                    controlsId={searchResultsId}
```
clairechabas: Just a suggestion: I think it could be more readable to keep a `resultsState` object throughout instead of spreading it :
```suggestion
                    resultsState={resultsState}
                    controlsId={searchResultsId}
```

packages/gitbook/src/components/Search/SearchInput.tsx
```diff
@@ -16,6 +16,10 @@ interface SearchInputProps {
     withAI: boolean;
     isOpen: boolean;
     className?: string;
+    resultsCount: number;
+    cursor: number | null;
+    resultsShowing: boolean;
```
clairechabas: Following up my suggestion above:

```suggestion
    resultsState: ResultsState;
```

packages/gitbook/src/components/Search/SearchContainer.tsx
```diff
@@ -161,6 +161,16 @@ export function SearchContainer(props: SearchContainerProps) {
 
     const visible = viewport === 'desktop' ? !isMobile : viewport === 'mobile' ? isMobile : true;
 
+    const [resultsState, setResultsState] = React.useState<{
+        count: number;
+        showing: boolean;
+        cursor: number | null;
+    }>({
```
clairechabas: Following up my suggestion below:
```suggestion
    const [resultsState, setResultsState] = React.useState<ResultsState>({
```

And the type would live above:
```
type ResultsState = {
        count: number;
        showing: boolean;
        cursor: number | null;
    }
```

packages/gitbook/src/components/Search/SearchInput.tsx
```diff
@@ -84,9 +100,16 @@ export const SearchInput = React.forwardRef<HTMLDivElement, SearchInputProps>(
                             className="size-4 shrink-0 animate-scale-in"
                         />
                     )}
-
+                    <div className="sr-only" aria-live="assertive" role="alert" aria-relevant="all">
+                        {resultsShowing
+                            ? resultsCount > 0
```
clairechabas: and :

```suggestion
                        {resultsState.showing
                            ? resultsState.count > 0
```

packages/gitbook/src/components/Search/SearchInput.tsx
```diff
@@ -26,7 +30,19 @@ const sizeClasses = ['text-sm', 'px-3.5', 'py-1.5', 'md:circular-corners:px-4'];
...
 export const SearchInput = React.forwardRef<HTMLDivElement, SearchInputProps>(
     function SearchInput(props, ref) {
-        const { onChange, onKeyDown, onFocus, value, withAI, isOpen, className } = props;
+        const {
+            onChange,
+            onKeyDown,
+            onFocus,
+            resultsCount,
+            resultsShowing,
+            cursor,
```
clairechabas:
```suggestion
           resultsState
```

PR #3640, opened 2025-09-11
Scope search across sections and variants

zenoachtig:
## Components
- Add `data-active` to `Button` to make styling easier
- Improve active style for blank buttons in `contrast-more` mode
- Add leading icon to `DropdownMenuItem`, to be further standardised in other places in follow-up PR
- Create `SegmentedControl` component that styles buttons to look like one big selector.

## Search
- Rework `global` search parameter into two separate ones to determine search scope:
  - `scope: all | current`: how "wide" to search on the site. Determines section scope.
  - `depth: single | full`: how "deep" to search within the scope. This determines the variant scope within sections.
- Change `SearchScopeToggle` to use new scope & depth params and segmented control to show a UI to switch between scopes.

## Demo

https://github.com/user-attachments/assets/75c31c47-70be-47c3-aa22-7c45bc39fa77

CleanShot 2025-09-11 at 12 07 14@2x

CleanShot 2025-09-11 at 12 10 19@2x


packages/gitbook/src/components/AI/useAI.tsx
```diff
@@ -137,7 +137,8 @@ export function useAI(): AIContext {
                     setSearchState((prev) => ({
                         ask: null, // Reset ask as we assume the assistant will handle it
                         query: prev?.query ?? null,
-                        global: prev?.global ?? false,
+                        scope: prev?.scope ?? 'all',
+                        depth: prev?.depth ?? 'single',
```
taranvohra: I am usually not a fan of multiple query params because coordinating becomes a mess.

What are your thoughts about a single `scope`:
- `current` -> just the current variant
- `extended` -> current + rest of the variants
- `global`  -> everything

taranvohra: We might also need to be backwards compatible `global: true -> scope: global`

zenoachtig: Yeah I struggled with this, open to a better way. Two problems with that proposal
1. We need a fourth state for "all variants in current" (which is different from `extended`)
2. The default will change depending on the site's structure. Variant-only sites start on `current` and can go to `global`, section-only sites start on `global` and can go to `current`, variant+section sites start on `extended` and can go to either. If we want to have the default be omitted from the URL, that would mean having a dynamic default in `useSearch`... unless I'm overthinking it

This matrix is complex but does solve the two issues above in a way that is transparent to the visitor in the URL. It's semantic about the scope's intent instead of rigid in different modes _and_ it allows for expansion in either axis.

packages/gitbook/src/intl/translations/en.ts
```diff
@@ -12,8 +12,10 @@ export const en = {
     search_no_results_for: 'No results for "${1}".',
     search_no_results: 'No results',
     search_results_count: '${1} results',
-    search_scope_space: '${1}',
-    search_scope_all: 'All content',
+    search_scope_current: '${1}',
+    search_scope_extended: 'All content',
+    search_scope_default: 'Best match',
+    search_scope_all: 'Everything',
```
taranvohra, line 18: Should it be "Everything" or "Entire site" 🤔

PR #3642, opened 2025-09-11
Add navigation loader and rename context provider

conico974:
Introduce a navigation loader component to indicate loading state during navigation and rename the hash context provider to better reflect its purpose.

The loading state only appears after 400ms, pages that are fast enough or already loaded won't show the loading indicator.

### DIFF
--- a/packages/gitbook/tailwind.config.ts
+++ b/packages/gitbook/tailwind.config.ts
@@ -508,10 +508,6 @@ const config: Config = {
                         scale: '1 1',
                         translate: '100% 0',
                     },
-                    // '60%': {
-                    //     scale: '1 1',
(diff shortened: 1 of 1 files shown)


packages/gitbook/tailwind.config.ts
```diff
@@ -498,6 +499,24 @@ const config: Config = {
...
+                        translate: '0 0',
+                    },
+                    '40%': {
+                        scale: '1 1',
+                        translate: '100% 0',
+                    },
+                    // '60%': {
+                    //     scale: '1 1',
+                    //     translate: '0 0',
+                    // },
```
clairechabas:
```suggestion
```

PR #3657, opened 2025-09-16
Add visibility label and smoothen magnification

viktorrenkema:
Add a tooltip to the toolbar to convey that these are not shown to customers
-> closes https://linear.app/gitbook-x/issue/RND-8155/add-visibility-affordance-to-gbo-toolbar

Tweak the magnification logic so that when the cursor enters and exits the hoverable zone, we ease into the scaling instead of abruptly scaling up/down.
-> fixes https://linear.app/gitbook-x/issue/RND-8144/gitbook-toolbar-magnifying-effect-not-perfect 

Demo:

https://github.com/user-attachments/assets/4b9be851-ec05-4936-9381-dc53ed126d27

### DIFF
--- a/packages/gitbook/e2e/customers.spec.ts
+++ b/packages/gitbook/e2e/customers.spec.ts
@@ -62,11 +62,11 @@ const testCases: TestsCase[] = [
         contentBaseURL: 'https://docs.gmgn.ai',
         tests: [{ name: 'Home', url: '/' }],
     },
-    {
-        name: 'docs.spicychat.ai',
-        contentBaseURL: 'https://docs.spicychat.ai',
-        tests: [{ name: 'Home', url: '/' }],
-    },
+    // {
+    //     name: 'docs.spicychat.ai',
+    //     contentBaseURL: 'https://docs.spicychat.ai',
+    //     tests: [{ name: 'Home', url: '/' }],
+    // },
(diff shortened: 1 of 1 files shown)


gregberge (approved):

packages/gitbook/e2e/customers.spec.ts
```diff
@@ -62,11 +62,11 @@ const testCases: TestsCase[] = [
...
-    {
-        name: 'docs.spicychat.ai',
-        contentBaseURL: 'https://docs.spicychat.ai',
-        tests: [{ name: 'Home', url: '/' }],
-    },
+    // {
+    //     name: 'docs.spicychat.ai',
+    //     contentBaseURL: 'https://docs.spicychat.ai',
+    //     tests: [{ name: 'Home', url: '/' }],
+    // },
```
zenoachtig, line 69:
```suggestion
```

viktorrenkema: Intent is to turn ci tests for this org back on. We've also commented another org in this file to disable them temporarily, so imo commenting is better otherwise it may be forgotten to turn it back on if the org doesnt remain suspended: https://gitbook.slack.com/archives/C01NXGWJELS/p1758031171853909

PR #3659, opened 2025-09-17
Convert relative URL to absolute in Markdown (page and copy)

gregberge:
When accessing a page from ".md" or using "Copy in Markdown", we need to convert link to absolute because the context (URL) is lost.


packages/gitbook/src/lib/markdownPage.ts
```diff
@@ -62,39 +67,60 @@ export async function getMarkdownForPage(
         };
     }
 
+    const tree = fromPageMarkdown({
+        // @TODO not sure how to get the basePath here
+        basePath: '/',
```
gregberge: I am not sure of this part. In the llms, we compute it from the section + space.

When I test the output it works with "/" but I am probably missing something here.

PR #3660, opened 2025-09-17
Add support for inline expressions rendering with visitor data on GBO side

spastorelli:
This PR aims to add support for inline expressions rendering with visitor data on GBO side so we can move the evaluation of inline expression from the API to GBO and allow for injecting dynamic claims values (e.g API keys) in code blocks/Open API block and page content.

The evaluation of inline expressions will also be removed (kept only for search indexation) from the API in another GBX PR.

https://github.com/user-attachments/assets/9c4b54ef-22e2-4657-b5ab-6290c42a6970

### DIFF
--- a/.changeset/little-months-exercise.md
+++ b/.changeset/little-months-exercise.md
@@ -1,0 +1,5 @@
+---
+"gitbook": patch
+---
+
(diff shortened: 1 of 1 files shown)


sdirosa (approved):
Looks good to me

PR #3667, opened 2025-09-18
Enhance tracking with geolocation headers and continent info in analytics proxy

conico974:
Add geolocation headers to improve analytics tracking by including country, latitude, longitude, and continent information. This enhancement allows for better insights into user locations.

Should fix RND-8185

### DIFF
--- a/packages/gitbook/src/lib/tracking.ts
+++ b/packages/gitbook/src/lib/tracking.ts
@@ -54,10 +54,10 @@ export async function serveProxyAnalyticsEvent(req: Request) {
         method: 'POST',
         headers: {
             'Content-Type': 'application/json',
-            'x-location-country': country || '',
-            'x-location-latitude': latitude || '',
-            'x-location-longitude': longitude || '',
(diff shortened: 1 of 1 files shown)


packages/gitbook/src/lib/tracking.ts
```diff
@@ -43,6 +54,10 @@ export async function serveProxyAnalyticsEvent(req: Request) {
         method: 'POST',
         headers: {
             'Content-Type': 'application/json',
+            'x-location-country': country || '',
```
emmerich:
```suggestion
            ...(country ? { 'x-location-country': country } : {}),
```

would omitting the header be better than the empty string? Fine either way

emmerich (approved):

packages/gitbook/src/lib/tracking.ts
conico974: Yeah that's a better idea, in theory once deployed it should never happened, but it's cleaner this way

PR #3677, opened 2025-09-22
Use siteSpace title instead of space title in search

zenoachtig:
We were exposing the space's internal title in search instead of the variant/siteSpace title.

CleanShot 2025-09-22 at 16 29 36@2x
CleanShot 2025-09-22 at 16 29 19@2x

### DIFF
--- a/packages/gitbook/src/components/Search/server-actions.tsx
+++ b/packages/gitbook/src/components/Search/server-actions.tsx
@@ -14,6 +14,7 @@ import type {
     SearchSpaceResult,
     SiteSection,
     SiteSectionGroup,
+    SiteSpace,
     Space,
 } from '@gitbook/api';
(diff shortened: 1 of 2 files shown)


packages/gitbook/src/components/Search/server-actions.tsx
spastorelli, line 1: Can you quickly check that we are consistent (and as per our expectation) in other places where we use the title?
Looks like in the variant dropdown for example we may still be using the space title:
https://github.com/GitbookIO/gitbook/blob/main/packages/gitbook/src/components/Header/SpacesDropdown.tsx#L50

zenoachtig: This actually makes us consistent with that title, we use `siteSpace` consistently now, the object is just named `space` in that instance. 👌

PR #3679, opened 2025-09-23
Improve default site icon

zenoachtig:
- Improve design of icon, using primary colour and optional border
- Improve `apple-touch-icon` support, fixing RND-6431
- Move image font rendering to `lib` and use it across `ogimage` and `icon` routes

### DIFF
--- a/packages/gitbook/src/lib/imageFonts.ts
+++ b/packages/gitbook/src/lib/imageFonts.ts
@@ -104,12 +104,20 @@ async function loadCustomFont(input: { url: string; weight: 400 | 700 }) {
 const staticCache = new Map<string, any>();
 
 async function getWithCache<T>(key: string, fn: () => Promise<T>) {
-    const cached = staticCache.get(key) as T;
+    const cached = staticCache.get(key) as Promise<T>;
     if (cached) {
-        return Promise.resolve(cached);
+        return cached;
     }
 
-    const result = await fn();
-    staticCache.set(key, result);
-    return result;
+    const promise = fn();
+    staticCache.set(key, promise);
+
+    try {
+        const result = await promise;
+        return result;
+    } catch (error) {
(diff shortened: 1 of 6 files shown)


packages/gitbook/src/lib/imageFonts.ts
```diff
@@ -0,0 +1,115 @@
...
+    } as const;
+}
+
+// biome-ignore lint/suspicious/noExplicitAny: <explanation>
+const staticCache = new Map<string, any>();
+
+async function getWithCache<T>(key: string, fn: () => Promise<T>) {
+    const cached = staticCache.get(key) as T;
+    if (cached) {
+        return Promise.resolve(cached);
```
gregberge:
```suggestion
        return  cached
```

packages/gitbook/src/lib/imageFonts.ts
```diff
@@ -0,0 +1,115 @@
...
+        data,
+        style: 'normal' as const,
+        weight,
+    } as const;
+}
+
+// biome-ignore lint/suspicious/noExplicitAny: <explanation>
+const staticCache = new Map<string, any>();
+
+async function getWithCache<T>(key: string, fn: () => Promise<T>) {
```
gregberge, line 106: If you want to improve the cache here, you should cache the promise, not the result. So concurrent calls would be cached.

zenoachtig: I've updated it, but don't know if this does what you mean — can you have a look? Feel free to adapt if needed.

gregberge: Not exactly I will adapt it in another PR. Feel free to merge.

gregberge: If you want to work well your func must not be "async".

PR #3682, opened 2025-09-24
Improve fonts loading

gregberge:
- If a font is not found don't ignore and cache the result, throw
  instead
- Fix the cache to cache the promise instead of the result of it
- Fix the typings

### DIFF
--- a/packages/gitbook/src/lib/imageFonts.ts
+++ b/packages/gitbook/src/lib/imageFonts.ts
@@ -4,11 +4,18 @@ import { type FontWeight, getDefaultFont } from '@gitbook/fonts';
 import { getFontSourcesToPreload } from '@/fonts/custom';
 import type { GitBookSiteContext } from '@/lib/context';
 import { filterOutNullable } from '@/lib/typescript';
+import QuickLRU from 'quick-lru';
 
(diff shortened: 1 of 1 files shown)


packages/gitbook/src/lib/imageFonts.ts
```diff
@@ -85,39 +86,40 @@ async function loadGoogleFont(input: {
 
 async function loadCustomFont(input: { url: string; weight: 400 | 700 }) {
     const { url, weight } = input;
-    const response = await fetch(url);
-    if (!response.ok) {
-        return null;
-    }
+    return getWithCache(`custom-font-files:${url}:${weight}`, async () => {
```
conico974: Using the static cache is not ideal here, custom font are user defined but there is nothing that deletes the cache. Technically the cache could grow indefinitely causing OOM.
I don't think it will be an issue in practice, but we should at least add a comment about it somewhere

zenoachtig (approved):

packages/gitbook/src/lib/imageFonts.ts
gregberge: I switched to LRU

PR #3683, opened 2025-09-24
Fix crash for card cover defined without objectFit

SamyPesse:


packages/gitbook/src/components/DocumentView/Table/utils.ts
```diff
@@ -71,16 +71,18 @@ function processCoverValue(value: DocumentTableImageRecord | string[] | null | u
         }
     }
 
+    const imageValue = value as DocumentTableImageRecord | null | undefined;
+
     // Check if it's the new schema with objectFit
-    if (value && typeof value === 'object' && 'ref' in value && 'objectFit' in value) {
+    if (imageValue && typeof imageValue === 'object' && 'ref' in imageValue) {
```
SamyPesse, line 77: `DocumentTableImageRecord` is:

```ts
/** A table record value for image columns, supporting both direct ContentRefs and the additional format with record-level settings. */
export type DocumentTableImageRecord = ContentRefFile | ContentRefURL | {
    ref: ContentRefFile | ContentRefURL;
    /** Object fit for image display in card views */
    objectFit?: CardsImageObjectFit;
};
```

So this test was incorrect as it's not accepting `{ ref: ... }`

conico974 (approved):

PR #3690, opened 2025-09-25
Ensure operation security reqs inherits from spec-level when not defined

spastorelli:


packages/react-openapi/src/resolveOpenAPIOperation.ts
```diff
@@ -71,7 +71,7 @@ export async function resolveOpenAPIOperation(
 
     return {
         servers,
-        operation,
+        operation: { ...operation, security },
```
spastorelli, line 74: `security` is already properly computed as `operation.security ?? schemaSecurity` above:
https://github.com/GitbookIO/gitbook/pull/3690/files#diff-37c21cbbac81d50e26d53ac90ca70400a28c1d1abf92bafe08cde662686c56ffR48

gregberge (approved):

sdirosa (approved):

PR #3691, opened 2025-09-26
Highlight discriminator properties in oneOf, allOf, anyOf objects

spastorelli:
Screenshot 2025-09-26 at 15 35 24

### DIFF
--- a/.changeset/mean-starfishes-promise.md
+++ b/.changeset/mean-starfishes-promise.md
@@ -1,0 +1,5 @@
+---
+"gitbook": minor
+---
+
+Move inline expression evaluation from API to GBO
--- a/.changeset/mighty-bears-type.md
+++ b/.changeset/mighty-bears-type.md
@@ -1,0 +1,5 @@
+---
+"gitbook": patch
+---
+
+Fix rounded outline items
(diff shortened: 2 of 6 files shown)


taranvohra (commented):
I'm not sure I understand the change. What does `discriminator` represent?

spastorelli:
> I'm not sure I understand the change. What does `discriminator` represent?

@taranvohra when dealing with oneOf, allOf, anyOf schemas, the [discriminator](https://swagger.io/specification/v3/#discriminator-object) is what helps you identify which property in the alternatives objects (ie. schemas/objects listed in oneOf, allOf, anyOf) that identify the data/object type. Basically, it's the property that helps discriminate between each of the objects contained within the oneOf, allOf, anyOf list.

packages/react-openapi/src/OpenAPISchemaName.tsx
```diff
@@ -27,9 +28,18 @@ export function OpenAPISchemaName(props: OpenAPISchemaNameProps) {
                     {propertyName}
                 </span>
             ) : null}
+            {isDiscriminatorProperty ? (
+                <span className="openapi-schema-discriminator">
+                    {t(context.translation, 'discriminator')}
+                </span>
+            ) : null}
```
taranvohra, line 35: In the screenshot, it looks slightly misaligned

taranvohra (approved):

PR #3692, opened 2025-09-26
Improve wide page width

zenoachtig:
- Simplify HTML layout of main page layout (TOC, main, outline) into a single flex container
- On pages with `wide` width, make the TOC and outline float up against the page content instead of all the way to the sides.

Before

CleanShot 2025-09-26 at 17 48 11@2x

After

CleanShot 2025-09-26 at 17 48 16@2x

### DIFF
--- a/.changeset/dirty-mice-sleep.md
+++ b/.changeset/dirty-mice-sleep.md
@@ -1,0 +1,6 @@
+---
+"@gitbook/react-openapi": patch
+"gitbook": patch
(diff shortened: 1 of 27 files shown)


gregberge (commented):
Problem with cover, see Argos https://app.argos-ci.com/GitbookIO/gitbook/builds/15381/202161823

gregberge (approved):

PR #3698, opened 2025-10-03
Allow hiding toolbar for one session or persistent

viktorrenkema:
This PR adds a menu to the toolbar shown on either preview/CR links or to authenticated users to:
- minify toolbar
- close toolbar this session
- dont show toolbar anymore at all (stored in local storage, our docs will have [an article](https://app.gitbook.com/o/d8f63b60-89ae-11e7-8574-5927d48c4877/s/NkEGS7hzeqa35sMXQZ4X/~/changes/799/resources/gitbook-ui/toolbar-on-published-sites/~/overview) explaining how to get it back)

Also the styles are updated a bit & using motion values for a much smoother magnification effect.

Still coming up: dragging (needs some more time as the variable width makes it harder)

https://github.com/user-attachments/assets/2cca3f98-6710-43ee-934a-431d734b1068

### DIFF
--- a/packages/gitbook/src/components/AdminToolbar/AdminToolbarClient.tsx
+++ b/packages/gitbook/src/components/AdminToolbar/AdminToolbarClient.tsx
@@ -1,10 +1,12 @@
 'use client';
 import { Icon } from '@gitbook/icons';
 import { MotionConfig } from 'motion/react';
+import React from 'react';
 import { useCheckForContentUpdate } from '../AutoRefreshContent';
 import { useVisitorSession } from '../Insights';
 import { useCurrentPagePath } from '../hooks';
-import { DateRelative } from '../primitives';
+import { DateRelative, Tooltip } from '../primitives';
+import { HideToolbarButton } from './HideToolbarButton';
 import { IframeWrapper } from './IframeWrapper';
 import { RefreshContentButton } from './RefreshContentButton';
 import {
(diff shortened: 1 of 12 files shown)


packages/gitbook/src/components/AdminToolbar/Toolbar.module.css
```diff
@@ -0,0 +1,40 @@
+.arcMenu {
+    --arc-width: 505px;
+    --arc-height: 400px;
+    --arc-radius: 34%;
+    --start-distance: -240px;
+    --spread-distance: 45px;
+}
```
viktorrenkema, line 7: This controls the arc menu that pops open when the button for hiding options is clicked. 

We use a border-radius on an invisible parent combined with offset-distance on each child button to then have them 'follow' [in this file](https://github.com/GitbookIO/gitbook/pull/3698/files#diff-06c3e3656f455c6c47ff2f4a59e118ca5b7d4d08f30de6e7dbac739ce03e54b4). 

Better visualized with a dashed border below:

CleanShot 2025-10-03 at 12 17 50@2x

packages/gitbook/src/components/AdminToolbar/AdminToolbarClient.tsx
```diff
@@ -17,48 +19,130 @@ import {
...
 export function AdminToolbarClient(props: AdminToolbarClientProps) {
-    const { context } = props;
+    const { context, onPersistentClose, onSessionClose, onToggleMinify } = props;
+    const [minified, setMinified] = React.useState(true);
     const visitorSession = useVisitorSession();
+    const [sessionClosed, setSessionClosed] = React.useState(false);
+    const [shouldHide, setShouldHide] = React.useState(false);
+
+    React.useEffect(() => {
+        const STORAGE_KEY = 'gitbook_toolbar_closed';
```
nolannbiron: you can put this outside the component

packages/gitbook/src/components/AdminToolbar/AdminToolbarClient.tsx
```diff
@@ -17,48 +19,130 @@ import {
...
+
+    const handleMinifiedChange = React.useCallback(
+        (value: boolean) => {
+            setMinified(value);
+            onToggleMinify?.();
+        },
+        [onToggleMinify]
+    );
+
+    const toolbarControls = React.useMemo(
```
nolannbiron: I don't think you need to memoize those (useMemo/useCallback)

packages/gitbook/src/components/AdminToolbar/HideToolbarButton.tsx
```diff
@@ -0,0 +1,236 @@
...
+export function HideToolbarButton(props: HideToolbarButtonProps) {
+    const { motionValues } = props;
+    const [open, setOpen] = React.useState(false);
+    const controls = useToolbarControls();
+
+    const ref = useRef<HTMLDivElement>(null);
+    const buttonRef = useRef<HTMLDivElement>(null);
+
+    const handleClickOutsideArcMenu = (event: Event) => {
+        // Don't close the arc if we are clicking  clicking on the button itself
```
nolannbiron:
```suggestion
        // Don't close the arc if we are clicking on the button itself
```

packages/gitbook/src/components/AdminToolbar/HideToolbarButton.tsx
```diff
@@ -0,0 +1,236 @@
...
+    React.useEffect(() => {
+        if (!open) return;
+
+        const handleScroll = () => setOpen(false);
+        window.addEventListener('scroll', handleScroll, { passive: true });
+
+        return () => window.removeEventListener('scroll', handleScroll);
+    }, [open]);
+
+    const items = React.useMemo(
```
nolannbiron: same, I don't think you need a memo

packages/gitbook/src/components/AdminToolbar/AdminToolbarClient.tsx
```diff
@@ -17,48 +19,130 @@ import {
...
+    const [minified, setMinified] = React.useState(true);
     const visitorSession = useVisitorSession();
+    const [sessionClosed, setSessionClosed] = React.useState(false);
+    const [shouldHide, setShouldHide] = React.useState(false);
+
+    React.useEffect(() => {
+        const STORAGE_KEY = 'gitbook_toolbar_closed';
+
+        try {
+            const hidden = !!localStorage.getItem(STORAGE_KEY);
```
sdirosa: Shouldn't we use something like `localStorage` from `@gitbook/util-local-storage`?

nolannbiron (approved):

packages/gitbook/src/components/AdminToolbar/AdminToolbarClient.tsx
sdirosa: Or how does it work on GBO? We let it crash?

viktorrenkema: Ah thanks for pointing that out, we do actually have utils already in gbo that i didnt notice: https://github.com/GitbookIO/gitbook/blob/viktor/toolbar/hide-and-close/packages/gitbook/src/lib/browser/local-storage.ts#L24. I've pushed a commit where I am using those now, and those handle the scenario where the operation fails: https://github.com/GitbookIO/gitbook/pull/3698/commits/ef53ff949b2050f962f6f84f47e194acf6a7f774

sdirosa (approved):

packages/gitbook/src/components/AdminToolbar/utils.ts
```diff
@@ -0,0 +1,154 @@
...
+
+import {
+    getLocalStorageItem,
+    getSessionStorageItem,
+    removeSessionStorageItem,
+    setLocalStorageItem,
+    setSessionStorageItem,
+} from '@/lib/browser/local-storage';
+
+export const STORAGE_KEY = 'gitbook_toolbar_closed';
```
sdirosa: nit: no need to export this, it's used only here in the utils

packages/gitbook/src/components/AdminToolbar/utils.ts
```diff
@@ -0,0 +1,154 @@
...
+
+    const closePersistent = () => {
+        setLocalStorageItem(STORAGE_KEY, true);
+        setPersistentHidden(true);
+        setSessionHidden(true, 'persistent');
+        setSessionReason('persistent');
+        onPersistentClose?.();
+    };
+
+    const hidden = persistentHidden || (!persistentHidden && sessionReason === 'session');
```
sdirosa: I think you can simplify this to:
```persistentHidden || sessionReason === 'session'```

PR #3702, opened 2025-10-04
Fix markdownPageURL for PageActions

nolannbiron:


nolannbiron:
@gregberge yeah I know but it's either we add it for copy and view, or remove it once for LLMs

gregberge:
The good way to fix it is to pass urls property with html / markdown property and use the correct one. Or better to compute it in the component where it's actually needed (because of RSC payload).

nolannbiron:
we can't compute in the component because we can't pass context from server to client, let's merge this first and I'll update to use `urls: {html, markdown}` after

PR #3707, opened 2025-10-08
Store customization in a cookie for preview route

conico974:
Implement cookie storage for customization and theme settings specifically for the preview route, enhancing user experience by retaining preferences during previews.

### DIFF
--- a/packages/gitbook/src/middleware.ts
+++ b/packages/gitbook/src/middleware.ts
@@ -296,16 +296,41 @@ async function serveSiteRoutes(requestURL: URL, request: NextRequest) {
         requestHeaders.set(MiddlewareHeaders.SiteURLData, JSON.stringify(stableSiteURLData));
 
         // Preview of customization/theme
-        const customization = siteRequestURL.searchParams.get('customization');
+        const customizationCookie = request.cookies.get(MiddlewareHeaders.Customization);
+        const customization =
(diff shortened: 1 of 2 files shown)


packages/gitbook/src/middleware.ts
```diff
@@ -296,16 +296,41 @@ async function serveSiteRoutes(requestURL: URL, request: NextRequest) {
...
             routeType = 'dynamic';
             // We need to encode the customization headers, otherwise it will fail for some customization values containing non ASCII chars on vercel.
             requestHeaders.set(MiddlewareHeaders.Customization, encodeURIComponent(customization));
+            cookies.push({
+                name: MiddlewareHeaders.Customization,
+                value: encodeURIComponent(customization),
+                options: {
+                    httpOnly: true,
+                    sameSite: 'lax',
+                    maxAge: 10 * 60, // 10 minutes
```
viktorrenkema, line 313: Why only 10mins?

viktorrenkema (approved):
lgtm, just 1 question

packages/gitbook/src/middleware.ts
conico974: I didn't see the point in going above that really. As soon as you land on the customization tab, this cookie will be reset or if you do any navigation, and having a low value makes it less likely to encounter an edge case that I didn't anticipate here.

PR #3708, opened 2025-10-08
Display headings from reusable content in TOC

gregberge:


packages/gitbook/src/components/PageBody/PageBody.tsx
```diff
@@ -45,6 +46,7 @@ export function PageBody(props: {
     const siteWidthWide = pageWidthWide || contentFullWidth;
     const language = getSpaceLanguage(context);
     const updatedAt = page.updatedAt ?? page.createdAt;
+    const getId = createGetSectionId();
```
SamyPesse: This approach is dangerous; we can't know for sure with react, especially async server component, that the order of rendering will be predictable.

SamyPesse (approved):

PR #3709, opened 2025-10-08
Fix web frame height

clairechabas:
Changes:
- Fix web frame height calculation which was too big for frames large widths.
- Add a spinner with closest height on load to improve loading experience.

**Before**
Screenshot 2025-10-10 at 15 20 21

**After**
Screenshot 2025-10-10 at 15 29 27


packages/react-contentkit/src/ElementWebframe.tsx
```diff
@@ -161,11 +159,26 @@ export function ElementWebframe(props: ContentKitClientElementProps<ContentKitWe
         return sendMessage({ state });
     }, [element.data, renderer.state, sendMessage]);
 
-    if (!mounted) {
-        return null;
-    }
+    const [iframeWidth, setIframeWidth] = React.useState<number>(0);
```
gregberge: Why using that and not only the observedWidth?

packages/react-contentkit/src/ElementWebframe.tsx
```diff
@@ -161,11 +159,26 @@ export function ElementWebframe(props: ContentKitClientElementProps<ContentKitWe
...
+    const { width: observedWidth } = useResizeObserver({ ref: iframeRef });
+
+    React.useEffect(() => {
+        if (observedWidth && observedWidth !== 0) {
+            setIframeWidth(observedWidth);
+        }
+    }, [observedWidth]);
 
     const aspectRatio = size.aspectRatio || element.aspectRatio;
+    const width = iframeRef.current?.clientWidth;
```
gregberge: Why do you need that if you have the observedWidth?

packages/react-contentkit/src/ElementWebframe.tsx
clairechabas: I updated that, I needed was to get the correct width on load and on resize and the new version works without the additional state and effect.

packages/react-contentkit/src/ElementWebframe.tsx
clairechabas: The `observedWidth` is `undefined` initially so the width is iframe `clientWidth`, then the observer ensures that the new width impacts the height calculation to keep it correct.

PR #3711, opened 2025-10-08
Fix "Something went wrong" error after loading the initial AI answer

utkuufuk:
This PR ensures that we don't use a computed revision ID when calling the `getRevisionById` API endpoint to display the source pages for an AI answer.

Now we extract the source revision ID from the computed revision ID and use it instead.

### DIFF
--- a/.changeset/clever-ducks-double.md
+++ b/.changeset/clever-ducks-double.md
@@ -1,0 +1,6 @@
+---
+'@gitbook/react-openapi': patch
+'gitbook': patch
+---
+
+Enhance OpenAPI security scopes handling
--- a/packages/gitbook/src/components/DocumentView/OpenAPI/style.css
+++ b/packages/gitbook/src/components/DocumentView/OpenAPI/style.css
@@ -317,10 +317,11 @@
 }
 
 .openapi-securities-oauth-flows {
-    @apply flex flex-col gap-2 divide-y divide-tint-subtle;
+    @apply flex flex-col gap-3;
 }
 
-.openapi-securities-oauth-content {
+.openapi-securities-oauth-content,
+.openapi-securities-scopes {
     @apply prose *:!prose-sm *:text-tint;
 }
 
@@ -328,7 +329,7 @@
     @apply text-xs;
 }
 
-.openapi-securities-oauth-content ul {
+.openapi-securities-scopes ul {
     @apply !my-0;
 }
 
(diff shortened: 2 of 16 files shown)


packages/gitbook/src/components/Search/server-actions.tsx
```diff
@@ -157,12 +157,17 @@ export async function streamAskQuestion({
...
+                                : source.revision;
+
                             spacePromises.set(
                                 source.space,
                                 throwIfDataError(
                                     context.dataFetcher.getRevision({
                                         spaceId: source.space,
-                                        revisionId: source.revision,
+                                        revisionId,
                                     })
```
utkuufuk, line 171: Here, we call the `getRevisionById` API endpoint using a computed revision ID which looks like this:
```
computed_kYuvMBuVxyeIcLP8yB1C_306581b1fb17ec1cfde808f99ecb55e092c4e762
```

Since the source revision ID is embedded inside the computed revision ID, we can extract and use it instead:
https://github.com/GitbookIO/gitbook-x/blob/6811c63ef2ad1381c578c70a54e869ed0af5c797/packages/hive-core/src/computed/revision.ts#L30-L37

packages/gitbook/src/components/Search/server-actions.tsx
```diff
@@ -157,12 +157,17 @@ export async function streamAskQuestion({
                         }
 
                         if (!spacePromises.has(source.space)) {
+                            // Extract the source revision ID if it's a computed revision.
+                            const revisionId = source.revision.startsWith('computed_')
+                                ? (source.revision.split('_')[1] ?? '')
+                                : source.revision;
+
```
taranvohra, line 164: I don't think it's the right fix. computed IDs can change format in the future. It's risky

SamyPesse: Yeah definitelly we should NEVER NEVER do this

SamyPesse: Also it's plain wrong, the second part is not the revision ID

packages/gitbook/src/components/Search/server-actions.tsx
SamyPesse: But why ? we should work with the actual computed revision; this is what is indexed

utkuufuk: @SamyPesse We extract the pages from the revision so we can display relevant pages along with the AI answer. However, we can only look up revisions (from Hive, via the `getRevisionById` API endpoint) by the source ID, not by the computed ID.

The revision returned by the API is still the computed one:
https://github.com/GitbookIO/gitbook-x/blob/621360305b4c1563e7373e0d35dcac8ab9e495a3/services/cloudrun/api/src/operations/spaces/content/utils.ts#L98-L124

utkuufuk:
Closing in favor of https://github.com/GitbookIO/gitbook-x/pull/19841

PR #3715, opened 2025-10-09
Add alt text support to card covers

nolannbiron:


packages/gitbook/src/components/DocumentView/InlineIcon.tsx
```diff
@@ -8,10 +8,8 @@ import { textColorToStyle } from './utils/colors';
 export async function InlineIcon(props: InlineProps<DocumentInlineIcon>) {
     const { inline } = props;
     const icon = inline.data.icon as IconName;
-    // @ts-expect-error remove this comment once API is updated
     const color = inline.data.color
-        ? // @ts-expect-error remove "as DocumentMarkColor['data']['text']" once API is updated
-          (inline.data.color as DocumentMarkColor['data']['text'])
+        ? (inline.data.color as DocumentMarkColor['data']['text'])
```
conico974: The `as` is still needed ? If so it would be great to at least add a comment as a TODO to remind us to remove it later. If not you can simplify it to just `inline.data.color ?? undefined` or something like that

nolannbiron: Oh no my bad, forgot to remove the casting

PR #3717, opened 2025-10-10
Support inline icon colors

gregberge:


packages/gitbook/src/components/DocumentView/InlineIcon.tsx
```diff
@@ -1,10 +1,23 @@
...
+import { textColorToStyle } from './utils/colors';
 
 export async function InlineIcon(props: InlineProps<DocumentInlineIcon>) {
     const { inline } = props;
+    const icon = inline.data.icon as IconName;
+    // @TODO remove the useless casting once API is updated
+    const color =
+        'color' in inline.data && inline.data.color
+            ? (inline.data.color as DocumentMarkColor['data']['text'])
+            : undefined;
```
SamyPesse: Instead can you do something with `@ts-expect-error` so that next time we update the client, it'll fail and force the person to fix this line

PR #3741, opened 2025-10-18
Add end-to-end tests for table of contents navigation

conico974:
Introduce end-to-end tests to verify the visibility and functionality of expandable sections in the table of contents. These tests ensure that navigation links appear as expected when sections are expanded.

### DIFF
--- a/packages/gitbook/e2e/internal.spec.ts
+++ b/packages/gitbook/e2e/internal.spec.ts
@@ -231,7 +231,6 @@ const testCases: TestsCase[] = [
             {
                 name: 'Expandable TOC navigation',
                 url: '',
-                screenshot: false,
                 run: async (page) => {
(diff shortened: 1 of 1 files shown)


packages/gitbook/e2e/internal.spec.ts
```diff
@@ -228,6 +228,59 @@ const testCases: TestsCase[] = [
                     );
                 },
             },
+            {
+                name: 'Expandable TOC navigation',
+                url: '',
+                screenshot: false,
```
spastorelli: Should you add one that screenshots after expanding the group as in the recent regression it was a visual regression on the children items?

conico974: You're right, I'll add it for the nested one

PR #3747, opened 2025-10-22
Fix preloading images

conico974:
Fix preloading images to avoid preloading every image on every route

### DIFF
--- a/.changeset/quiet-icons-deny.md
+++ b/.changeset/quiet-icons-deny.md
@@ -1,0 +1,5 @@
+---
+'gitbook': patch
+---
+
+Fix ToggeableLinkItem display
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -54,10 +54,9 @@ After forking this repository, you'll want to [create a branch](https://docs.git
 #### 3. Install dependencies and run the project locally
 
 ##### Prerequisites:
-- Node.js (Version: >= 22.3)
-  - Use `nvm` for easy Node management
-- [Bun](https://bun.sh/) (Version: >=1.2.15)
-  - We use a text-based lockfile which isn't supported below 1.2.15
+
+- [Node.js](https://nodejs.org/en) (see "engines" in `package.json`)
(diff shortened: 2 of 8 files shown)


packages/gitbook/src/components/DocumentView/Table/RecordColumnValue.tsx
```diff
@@ -229,7 +229,7 @@ export async function RecordColumnValue<Tag extends React.ElementType = 'div'>(
                                                 },
                                             },
                                         }}
-                                        priority="lazy"
+                                        loading="lazy"
```
emmerich, line 232: `loading="lazy" preload={false}` should be the default for an image, so we shouldn't need to supply it.

packages/gitbook/src/components/utils/Image.tsx
```diff
@@ -139,7 +139,7 @@ export function Image(
                     inline={inline}
                     // We don't want to preload the dark image, because it's not visible
                     // TODO: adapt based on the default theme
-                    priority="lazy"
+                    loading="lazy"
```
emmerich, line 130: Should we also set preload={false} here?

packages/gitbook/src/components/utils/Image.tsx
emmerich: Can we remove this prop as it's no longer supported?

packages/gitbook/src/components/DocumentView/Images.tsx
```diff
@@ -135,8 +135,7 @@ async function ImageBlock(props: {
                               }
                             : null,
                     }}
-                    priority={isEstimatedOffscreen ? 'lazy' : 'high'}
-                    preload
+                    loading={isEstimatedOffscreen ? 'lazy' : 'eager'}
```
emmerich, line 138: Actually should this be:

```{ preload: !isEstimatedOffscreen }```

too

packages/gitbook/src/components/utils/Image.tsx
conico974: It's not necessary in theory

PR #3749, opened 2025-10-23
Fix: height issues with webframe

BrettJephson:
Updates algorithm for sizing webframe to fix two issues:
1. webframe adding extra space under Mermaid diagrams 
3. webframe cutting off content in integration form. 

Added another Mermaid diagram to e2e tests to test landscape. 

Tested with:
https://gitbook-v2-q3lwp8faq-gitbook.vercel.app/url/mariadb.com/docs/mariadb-cloud#mariadb-cloud-autonomous-resilient-end-to-end-secure (mermaid diagram)
https://gitbook-v2-q3lwp8faq-gitbook.vercel.app/url/gitbook.gitbook.io/test-gitbook-open/blocks/integrations (mermaid diagram)
https://gitbook-v2-q3lwp8faq-gitbook.vercel.app/url/payments.govstack.global/10-workflows (mermaid diagrams)
https://gitbook-v2-q3lwp8faq-gitbook.vercel.app/url/mariadb.com/docs/platform/post-download/mariadb-server-12.0.2 (marketo form)

### DIFF
--- a/.changeset/lazy-snails-warn.md
+++ b/.changeset/lazy-snails-warn.md
@@ -1,0 +1,5 @@
+---
+"gitbook": patch
+---
+
+Move the "Ask <query>" to the bottom of search results if query is not a question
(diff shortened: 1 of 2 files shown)


clairechabas (approved):
It fixes the height calculation when the iframe content is a form but it regresses back to the original bug on the height calculation for diagrams which have a large width and a small height.

I approve because the bug on the form iframe is more impactful than the one on the diagrams height so if we can't figure out a good fix for both we should fix the first one in priority and tackle the other separately.

PR #3750, opened 2025-10-23
Update search scope UI

zenoachtig:
CleanShot 2025-10-23 at 18 53 26@2x
CleanShot 2025-10-23 at 18 53 31@2x

### DIFF
--- a/packages/gitbook/src/components/DocumentView/Expandable/Details.tsx
+++ b/packages/gitbook/src/components/DocumentView/Expandable/Details.tsx
@@ -17,29 +17,34 @@ export function Details(props: {
 }) {
     const { children, id, className } = props;
 
-    const detailsRef = React.useRef<HTMLDetailsElement>(null);
+    const ref = React.useRef<HTMLDetailsElement>(null);
 
     const [openFromHash, setOpenFromHash] = React.useState(false);
 
     const hash = useHash();
+
     /**
      * Open the details element if the url hash refers to the id of the details element
      * or the id of some element contained within the details element.
      */
     React.useEffect(() => {
-        if (!hash || !detailsRef.current) {
+        if (!hash || !ref.current) {
             return;
         }
+
         if (hash === id) {
             setOpenFromHash(true);
+            return;
         }
+
         const activeElement = document.getElementById(hash);
-        setOpenFromHash(Boolean(activeElement && detailsRef.current?.contains(activeElement)));
+        const isOpen = Boolean(activeElement && ref.current.contains(activeElement));
+        setOpenFromHash(isOpen);
     }, [hash, id]);
 
     return (
         <details
-            ref={detailsRef}
+            ref={ref}
             id={id}
             open={props.open || openFromHash}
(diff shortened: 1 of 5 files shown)


packages/gitbook/src/components/Search/SearchScopeToggle.tsx
```diff
@@ -26,108 +25,121 @@ export function SearchScopeToggle(props: {
...
+                            </span>
+                        </div>
+                    </DropdownMenuItem>
+                    <DropdownMenuItem
+                        leadingIcon={section?.icon ?? 'crosshairs'}
+                        className="gap-3"
+                        active={!sectionScopeIsExtended}
+                        onClick={() => setSearchState({ ...state, scope: 'current' })}
+                    >
+                        <div className="flex flex-col">
```
taranvohra: While testing I noticed there weren't any gaps between these dropdowns

Image

packages/gitbook/src/components/Search/SearchScopeToggle.tsx
```diff
@@ -26,108 +25,121 @@ export function SearchScopeToggle(props: {
...
+                            <span className="text-tint-strong">
+                                {t(language, 'search_scope_variant_all')}
+                            </span>
+                            <span className="text-tint-subtle text-xs">
+                                {t(language, 'search_scope_variant_all_description')}
+                            </span>
+                        </div>
+                    </DropdownMenuItem>
+                </DropdownMenu>
```
taranvohra: Can we pull it out into components? We have so many DropdownMenu instances in this it's quite hard to read.

taranvohra (approved):

packages/gitbook/src/components/Search/SearchScopeToggle.tsx
zenoachtig: This is mostly intentional to keep the gap visually consistent between all three elements when not hovered. I'll have one more look but I don't think increasing it to have a gap works

zenoachtig: Update: nevermind I did change it a bit so there's a gap between the two. It's a bit visually inconsistent but not too noticeable

packages/gitbook/src/components/Search/SearchScopeControl.tsx
```diff
@@ -0,0 +1,175 @@
...
+            ) : null}
+
+            {withVariants && (!withSections || !sectionScopeIsExtended) ? (
+                <SearchScopeVariantControl isExtended={variantScopeIsExtended} {...props} />
+            ) : null}
+        </div>
+    );
+}
+
+function SearchScopeSectionControl(props: SearchScopeControlProps & { isExtended: boolean }) {
```
taranvohra, line 48: Another quick thing I noticed while testing. If the site has only 1 section there is no need of a filter.

taranvohra: Ex - https://library.zoom.com

PR #3752, opened 2025-10-23
Improve AI Search branding

zenoachtig:
I had a little burst of inspiration and made the AI Search UX a bit more consistent. It now has a consistent visual identity, in line with- but different from the Assistant. I'm using the same "hold messages" from the Assistant to make the loading state of AI search more deliberate. And once the answer is streaming in, we show a title and a button to get back.

CleanShot 2025-10-23 at 20 08 25
CleanShot 2025-10-23 at 19 39 13

### DIFF
--- a/packages/gitbook/src/components/AIChat/AIChatIcon.tsx
+++ b/packages/gitbook/src/components/AIChat/AIChatIcon.tsx
@@ -9,13 +9,13 @@ interface AIChatIconProps extends React.SVGProps<SVGSVGElement> {
     trademark?: boolean;
 }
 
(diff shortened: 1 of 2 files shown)


packages/gitbook/src/components/AIChat/AIChatIcon.tsx
```diff
@@ -164,3 +165,35 @@ export const AIChatIcon = ({
         </svg>
     );
 };
+
+export const AISearchIcon = ({
```
taranvohra: Use `function` to keep things consistent

PR #3757, opened 2025-10-25
Fix scrolling hash into tabs / expandable

gregberge:
This one fixes:

- Scroll to a expandable id
- Scroll to a heading inside an expandable (native with details)
- Scroll to a tab id
- Scroll to a heading inside a tab

### DIFF
--- a/packages/gitbook/src/components/DocumentView/Tabs/DynamicTabs.tsx
+++ b/packages/gitbook/src/components/DocumentView/Tabs/DynamicTabs.tsx
@@ -87,7 +87,7 @@ export function DynamicTabs(props: {
         );
     }, [id, tabs, tabsState]);
 
-    // Track is the tab has been touched by the user.
+    // Track if the tab has been touched by the user.
     const touchedRef = useRef(false);
 
     // To avoid issue with hydration, we only use the state from localStorage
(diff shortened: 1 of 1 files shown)


packages/gitbook/src/components/DocumentView/Tabs/DynamicTabs.tsx
```diff
@@ -79,28 +87,33 @@ export function DynamicTabs(props: {
         );
     }, [id, tabs, tabsState]);
 
+    // Track is the tab has been touched by the user.
```
viktorrenkema:
```suggestion
    // Track if the tab has been touched by the user.
```

packages/gitbook/src/components/DocumentView/Tabs/DynamicTabs.tsx
```diff
@@ -145,10 +159,44 @@ export function DynamicTabs(props: {
                 return;
             }
 
-            selectTab(tabPanel.id);
+            selectTab(tabPanel.id, false);
+        } else {
+            const firstTab = tabs[0];
+            if (firstTab) {
+                selectTab(firstTab.id, false);
```
viktorrenkema: What's this for, why check if firstTab is defined? Is it so that you confirm that the tabs actually exist in the DOM before you try and open the actual one that is opened?

gregberge: Yes it's useless.

PR #3758, opened 2025-10-27
Support cover image heights and positioning changes

viktorrenkema:
- Sets the new `height` property on the page cover, set in the app or through md
- Changed the way we set the `y` position of a page cover to align the app
- Added a shimmer state to better fallback UI while cover is retrieved and avoid the objectPosition shifting from default 50% to whatever value was set.

## Demo

https://github.com/user-attachments/assets/2ee456a8-1347-4a6d-9014-9dca8bb7cabe

### DIFF
--- a/packages/gitbook/e2e/internal.spec.ts
+++ b/packages/gitbook/e2e/internal.spec.ts
@@ -33,6 +33,7 @@ import {
     headerLinks,
     runTestCases,
     waitForCookiesDialog,
+    waitForCoverImages,
     waitForNotFound,
 } from './util';
 
@@ -906,7 +907,10 @@ const testCases: TestsCase[] = [
             {
                 name: 'With cover',
                 url: 'page-options/page-with-cover',
-                run: waitForCookiesDialog,
+                run: async (page) => {
+                    await waitForCookiesDialog(page);
+                    await waitForCoverImages(page);
+                },
             },
             {
                 name: 'With cover for dark mode',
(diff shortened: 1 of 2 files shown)


packages/gitbook/src/components/PageBody/PageCoverImage.tsx
```diff
@@ -18,28 +17,16 @@ interface Images {
...
-}
-
 export function PageCoverImage({ imgs, y }: { imgs: Images; y: number }) {
-    const containerRef = useRef<HTMLDivElement>(null);
+    const { containerRef, objectPositionY, isLoading } = useCoverPosition(imgs, y);
 
-    const container = useResizeObserver({
-        // @ts-expect-error wrong types
-        ref: containerRef,
-    });
```
viktorrenkema, line 42: this and more was moved to the hook useCoverPosition.ts

packages/gitbook/src/components/PageBody/PageCover.tsx
```diff
@@ -78,6 +85,7 @@ export async function PageCover(props: {
         <div
             id="page-cover"
             data-full={String(as === 'full')}
+            style={{ height: `${height}px` }}
```
gregberge:
```suggestion
            style={{ height }}
```

packages/gitbook/src/components/PageBody/useCoverPosition.ts
```diff
@@ -0,0 +1,114 @@
...
+            }
+
+            const clampedOffset = Math.max(-maxOffset, Math.min(maxOffset, offset));
+            const relative = (maxOffset - clampedOffset) / (2 * maxOffset);
+            return relative * 100;
+        },
+        [container.height, imageDimensions, maxOffset, safeRatio]
+    );
+
+    const objectPositionY = clampedObjectPositionY(y);
```
gregberge: A useMemo would be better here, it's useless to set a useCallback and use it like that.

gregberge (approved):
Can't we do it in CSS only to avoid a client component here?

viktorrenkema:
> Can't we do it in CSS only to avoid a client component here?

I don't think that's reasonably possible as ultimately we are dependent on the page cover's container dimensions which adapt to the viewport in which its rendered. Even if we could make it fully css-based, it'd require rethinking the entire setup and i dont think that's worth spending the time rn.

PR #3763, opened 2025-10-27
Add confirmation to "Copy MCP URL" page action

zenoachtig:
https://github.com/user-attachments/assets/bc7912e7-a095-4293-b57d-4dd8a0d30c99

- Introduce `key` to `useCopiedStore` so we can have multiple elements that don't all respond to the same Copy action.
- Add `useCopiedStore` functionality to "Copy MCP URL" page action (via `CopyToClipboard` action).

### DIFF
--- a/packages/gitbook/src/components/PageActions/PageActions.tsx
+++ b/packages/gitbook/src/components/PageActions/PageActions.tsx
@@ -304,7 +304,8 @@ export function CopyToClipboard(props: {
     const closeDropdown = useDropdownMenuClose();
 
     const language = useLanguage();
-    const { copied, copy } = useCopiedStore('label');
+    const labelKey = label.toLowerCase().replace(/\s+/g, '_');
+    const { copied, copy } = useCopiedStore(labelKey);
 
     return (
         <PageActionWrapper


packages/gitbook/src/components/PageActions/PageActions.tsx
```diff
@@ -293,14 +304,29 @@ export function CopyToClipboard(props: {
     icon: IconName;
 }) {
     const { type, data, label, description, icon } = props;
+
+    const closeDropdown = useDropdownMenuClose();
+
+    const language = useLanguage();
+    const { copied, copy, key } = useCopiedStore();
```
nolannbiron: can you do something to pass the key like this `const {} = useCopiedStore(key)` ?
it would separate the contexts so you know that copied is for this key

packages/gitbook/src/components/PageActions/PageActions.tsx
```diff
@@ -293,14 +300,28 @@ export function CopyToClipboard(props: {
     icon: IconName;
 }) {
     const { type, data, label, description, icon } = props;
+
+    const closeDropdown = useDropdownMenuClose();
+
+    const language = useLanguage();
+    const { copied, copy } = useCopiedStore('label');
```
nolannbiron: You need to transform the label into a key (normalize & replace spaces with _), otherwise every instance of CopyToClipboard will have the same state

nolannbiron (approved):
the state management looks good, just one thing to fix before merging

PR #3772, opened 2025-10-29
Support translations and generic variants together

zenoachtig:
Add rudimentary support for translations & generic variants to be used on the same site. The variant dropdown on the left and translation dropdown in the header can now function independently from each other.

- Refactor `withVariants` into a `variants` object that contains two arrays for the different types: `translations` and `generic`.
- Switch the logic around to check these two arrays independently of each other.
- If there are any translations (spaces with a different language than the current), the menu dropdown is shown. If there are any generic variants (spaces with the same language as the current), the variant dropdown is shown.

### DIFF
--- a/packages/gitbook/src/components/SpaceLayout/SpaceLayout.tsx
+++ b/packages/gitbook/src/components/SpaceLayout/SpaceLayout.tsx
@@ -10,7 +10,6 @@ import { Footer } from '@/components/Footer';
 import { Header, HeaderLogo } from '@/components/Header';
 import { TableOfContents } from '@/components/TableOfContents';
 import { CONTAINER_STYLE } from '@/components/layout';
-import { languages } from '@/intl/translations';
 import type { VisitorAuthClaims } from '@/lib/adaptive';
 import { GITBOOK_APP_URL } from '@/lib/env';
 import { tcls } from '@/lib/tailwind';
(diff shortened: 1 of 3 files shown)


packages/gitbook/src/components/SpaceLayout/SpaceLayout.tsx
```diff
@@ -95,6 +94,57 @@ export function SpaceLayoutServerContext(props: SpaceLayoutProps) {
     );
 }
 
+function getSpaceVariants(context: GitBookSiteContext) {
```
taranvohra:
```suggestion
function categorizeVariants(context: GitBookSiteContext) {
```
To avoid confusion because `getSpaceVariants` hints something about a specific space.

taranvohra: Also a small one liner comment for this function

taranvohra: For whenever you add the E2E test, let's also add an unit test for this function. I think it's doing a lot of filtering and checks at multiple steps so it'd be good to ensure the right behaviour

taranvohra (approved):

packages/gitbook/src/components/SpaceLayout/SpaceLayout.tsx
zenoachtig: I've added unit tests in this one already, e2e I'll do in a separate PR 😊

PR #3774, opened 2025-10-30
Initialize hash on mount and improve navigation handling

conico974:
Initialize the hash on component mount to support SSR rehydration. Ignore same-page hash navigations to prevent unnecessary state updates.

### DIFF
--- a/packages/gitbook/src/components/hooks/useHash.tsx
+++ b/packages/gitbook/src/components/hooks/useHash.tsx
@@ -51,6 +51,8 @@ export const NavigationStatusProvider: React.FC<React.PropsWithChildren> = ({ ch
 
     // Cleanup timeout on unmount
     React.useEffect(() => {
+        // Initialize hash on mount - It could be null on SSR rehydration
+        setHash(getHash());
         return () => {
             if (timeoutRef.current) {
                 clearTimeout(timeoutRef.current);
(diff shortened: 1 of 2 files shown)


packages/gitbook/src/components/hooks/useHash.tsx
```diff
@@ -73,7 +75,8 @@ export const NavigationStatusProvider: React.FC<React.PropsWithChildren> = ({ ch
         if (timeoutRef.current) {
             clearTimeout(timeoutRef.current);
         }
-        if (pathnameRef.current !== url.pathname) {
+        // We don't want to set isNavigating for same page hash navigation
+        if (pathnameRef.current !== url.pathname && !href.startsWith('#')) {
```
gregberge, line 79: I don't understand this one.

packages/gitbook/src/components/hooks/useScrollPage.ts
```diff
@@ -13,8 +12,7 @@ import { usePrevious } from './usePrevious';
 export function useScrollPage() {
     const hash = useHash();
     const previousHash = usePrevious(hash);
-    const pathname = usePathname();
```
gregberge, line 16: Why is it not needed any more?

packages/gitbook/src/components/hooks/useHash.tsx
conico974: Not related, just found another bug while i was looking into this one that made the loading bar appear when it shouldn't

packages/gitbook/src/components/hooks/useScrollPage.ts
conico974: It should never have been needed in the first place. `useHash` should have the correct value, the issue was on the `useHash` hook (because of a SSR hydration bug)

gregberge: Because when I was navigating from a page to another sometimes the scroll was not reset.

PR #3779, opened 2025-10-31
Support custom cover heights

viktorrenkema:
This PR handles the GBO-side of enabling cover height resizing. The basic idea is:
1. by default, we handle cover images the same as usual (aka there's a fixed aspect ratio as recommended in the docs that we adhere to, so this can lead to cover images [of e.g. < 100px tall](https://help.connected.illumina.com/) on smaller viewports)
2. if `height` exists on the page cover, it means the cover was resized. because we can't/shouldnt enforce an aspect ratio on a cover that has a dynamic height, different logic kicks in to determine the appropriate height depending on the dimensions so that the entire image is contained within the cover

The previous PR was [reverted](https://github.com/GitbookIO/gitbook/pull/3771), as we didn't do step 1 above: we applied the 'new' sizing rules to existing page covers, which causes a lot of existing page covers to regress.

To help explain how a page cover has a fixed aspect ratio by default but how setting a height will impact that, we have tooltip UI in the editor + I have [a CR ready](https://app.gitbook.com/o/d8f63b60-89ae-11e7-8574-5927d48c4877/s/Ua3kTfM3iWAoECzM0u90/~/changes/82/editing-content/assets-and-files/whats-the-recommended-page-cover-image-size/~/comments) to add more details in the support article.

https://github.com/user-attachments/assets/aa611951-d064-4470-8645-b9e034574613

### DIFF
--- a/packages/gitbook/src/components/PageBody/useCoverPosition.ts
+++ b/packages/gitbook/src/components/PageBody/useCoverPosition.ts
@@ -37,8 +37,8 @@ export function useCoverPosition(imgs: Images, y: number) {
 
     // Load original image dimensions if not provided in `imgs`
     useLayoutEffect(() => {
-        // Check if we have dimensions from either light or dark image
-        const hasDimensions = imgs.light.size || imgs.dark?.size;
+        // Check if we have dimensions from dark (if provided) or else the default light.
+        const hasDimensions = imgs.dark?.size || imgs.light.size;
 
         if (hasDimensions) {
             return; // Already have dimensions
@@ -65,8 +65,8 @@ export function useCoverPosition(imgs: Images, y: number) {
     }, [imgs.light, imgs.dark]);
 
     // Use provided dimensions or fall back to loaded dimensions
-    // Check light first, then dark, then loaded dimensions
-    const imageDimensions = imgs.light.size ?? imgs.dark?.size ?? loadedDimensions;
+    // Check dark first, then light, then loaded dimensions
+    const imageDimensions = imgs.dark?.size ?? imgs.light.size ?? loadedDimensions;
 
     // Calculate ratio and dimensions similar to useCoverPosition hook
     const ratio =


packages/gitbook/src/components/PageBody/useCoverPosition.ts
```diff
@@ -0,0 +1,109 @@
...
+
+        if (hasDimensions) {
+            return; // Already have dimensions
+        }
+
+        setIsLoading(true);
+
+        // Load the original image (using src, not srcSet) to get true dimensions
+        // Use dark image if available, otherwise fall back to light
+        const imageToLoad = imgs.dark || imgs.light;
```
conico974, line 51: I don't understand why we take the dark one here. Everywhere else it tries from light image first

conico974 (commented):
Isn't it an issue that both light and dark image use the same sizes/aspect-ratio/position ?

packages/gitbook/src/components/PageBody/useCoverPosition.ts
viktorrenkema: I get what you mean, but it was intentional though. Its to account for _if_ the dark img exists, it means it was intentionally uploaded as the light is the default and dark an explicit opt-in. So when we get into the situation where the img didnt ship with a size payload and we manually need to check the dimensions, we just first check if the dark one exists here to avoid always going for the light one directly and possibly not accounting for a dark mode with differign dimensions. We still fallback to light anyway if there is no dark one.

viktorrenkema:
> Isn't it an issue that both light and dark image use the same sizes/aspect-ratio/position ?

Yeah it is, I listed the same as a current limitation somewhere in another PR. I chatted about it with samy but while it'd be nice to support heights and object fit on both dark and light mode individually, its just not included yet in this feature release.

packages/gitbook/src/components/PageBody/useCoverPosition.ts
```diff
@@ -0,0 +1,109 @@
...
+        img.onerror = () => {
+            // If image fails to load, use a fallback
+            setIsLoading(false);
+        };
+        img.src = imageToLoad.src;
+    }, [imgs.light, imgs.dark]);
+
+    // Use provided dimensions or fall back to loaded dimensions
+    // Check light first, then dark, then loaded dimensions
+    const imageDimensions = imgs.light.size ?? imgs.dark?.size ?? loadedDimensions;
```
conico974: We could use dark first here as well

PR #3787, opened 2025-11-04
Update `DropdownMenu` styling

zenoachtig:
- Fix width and height calculation to be viewport-dependent
- Standardise content and subcontent classnames into constants
- Add proper enter and exit animations based on `data-state`.


viktorrenkema (approved):
not a lot of design changes? lil screenshot or video would've helped to put it in context next time :)

PR #3792, opened 2025-11-10
Filter invalid events sent to the API

conico974:
Implement filtering for events older than 5 minutes before sending them to the API, ensuring only valid events are processed. Additionally, adjust the debounce function to include a maximum wait time.


packages/gitbook/src/lib/tracking.ts
```diff
@@ -48,6 +49,22 @@ export async function serveProxyAnalyticsEvent(req: Request) {
         });
     }
 
+    const body = (await req.json()) as { events: api.SiteInsightsEvent[] };
+    // Here we should filter every event that is older than 5 minutes as the API will reject them anyway, we might as well do it here
```
SamyPesse, line 53: I don't think we should have to do this; we are doing layers of protections because we are not identifying the root bug

conico974: I get what you mean, but I think this is a good compromise for now

PR #3793, opened 2025-11-11
Resolve page meta links for metadata

taranvohra:
This PR adds the support for resolving page meta links while rendering the page. It's set at a 10% rollout for now


packages/gitbook/src/components/SitePage/fetch.ts
```diff
@@ -20,14 +23,24 @@ export type PageParams = PagePathParams | PageIdParams;
  * Optimized to fetch in parallel as much as possible.
  */
 export async function fetchPageData(context: GitBookSiteContext, params: PageParams) {
-    const pageTarget = await resolvePage(context, params);
+    const [pageTarget, metaLinks] = await Promise.all([
+        resolvePage(context, params),
+        'pageId' in params ? resolvePageMetaLinks(context, params.pageId) : null,
```
SamyPesse: `fetchPageData` is used in a bunch of places including the page body.

Fetching meta links here means we'll not benefit from the streaming of page of the page body if it has to wait for the meta links.

Can we fetch it only where it's used in the `generateMetadata`?

packages/gitbook/src/components/SitePage/fetch.ts
```diff
@@ -20,14 +23,24 @@ export type PageParams = PagePathParams | PageIdParams;
...
+    ]);
+
+    const pageMetaLinks = await (metaLinks
+        ? metaLinks
+        : pageTarget?.page
+          ? resolvePageMetaLinks(context, pageTarget.page.id)
```
SamyPesse: Related to my comment above, it's even worse if we resolve them each time

packages/gitbook/src/components/SitePage/fetch.ts
```diff
@@ -106,6 +119,50 @@ async function resolvePage(context: GitBookSiteContext, params: PagePathParams |
...
+            (acc, { space }) => {
+                acc[space.id] = space;
+                return acc;
+            },
+            {} as Record<string, Space>
+        );
+
+        const [resolvedCanonical, ...resolvedAlternates] = await Promise.all(
+            [pageMetaLinks.canonical, ...(pageMetaLinks.alternates || [])]
+                .filter(filterOutNullable)
```
SamyPesse: Your spread logic line 140 doesn't work if you do a filter, as the canonical is optional, it cam mean an alternate will be skipped and shown as canonical

packages/gitbook/src/components/SitePage/fetch.ts
```diff
@@ -106,6 +119,50 @@ async function resolvePage(context: GitBookSiteContext, params: PagePathParams |
...
+                .map(async (link) => {
+                    return {
+                        href: (await resolveContentRef(link, context))?.href,
+                        language:
+                            link.kind === 'url'
+                                ? null
+                                : spacesById[link.space ?? context.space.id]?.language || null,
```
SamyPesse: It could be better if `resolveContentRef` returns the space and you use it from there

packages/gitbook/src/components/SitePage/fetch.ts
taranvohra: good catch!

packages/gitbook/src/components/SitePage/SitePage.tsx
```diff
@@ -133,6 +136,25 @@ export async function generateSitePageMetadata(props: SitePageProps): Promise<Me
     const { site, customization, revision, linker, imageResizer } = context;
     const siteStructureTitle = getSiteStructureTitle(context);
 
+    const canonical = pageMetaLinks?.canonical
+        ? pageMetaLinks.canonical
+        : // Trim trailing slashes in canonical URL to match the redirect behavior
+          linker
+              .toAbsoluteURL(linker.toPathForPage({ pages: revision.pages, page }))
+              .replace(/\/+$/, '');
```
SamyPesse: I can't remember why we have this; maybe it should be done on the other canonical as well

packages/gitbook/src/components/SitePage/SitePage.tsx
```diff
@@ -237,12 +257,76 @@ async function getPageDataWithFallback(args: {
...
+    const pageMetaLinks = await getDataOrNull(
+        context.dataFetcher.listRevisionPageMetaLinks({
+            spaceId: context.space.id,
+            revisionId: context.revisionId,
+            pageId,
+        })
+    );
+
+    if (pageMetaLinks) {
+        const spacesById = context.siteSpaces.reduce(
```
SamyPesse: What is the point of this, since you have `resolved.space.id`?

SamyPesse (approved):

packages/gitbook/src/components/SitePage/SitePage.tsx
```diff
@@ -237,12 +257,76 @@ async function getPageDataWithFallback(args: {
 }) {
     const { context: baseContext, pagePathParams } = args;
     const { context, pageTarget } = await fetchPageData(baseContext, pagePathParams);
+    const pageMetaLinks = await (pageTarget?.page
```
SamyPesse: IMPORTANT: to avoid getting too many requests at once, do a % rolling here based on the site ID

packages/gitbook/src/components/SitePage/SitePage.tsx
taranvohra: ah yes, good point

PR #3794, opened 2025-11-11
Support rendering upcoming Update blocks

viktorrenkema:
This PR implements the rendering of the upcoming Updates and Update block (an Updates block is a wrapper that should always be present if an Update block is encountered)
- we display an Update block with the provided date and date-format (stored on the parent Updates block)
- on smaller viewports, the date is positioned above instead of on the left

More features like sticky dates and richer date formats are coming up next. Update blocks are unreleased so not used yet.

Dark mode:
CleanShot 2025-11-11 at 16 44 35@2x

https://github.com/user-attachments/assets/919cd517-ebda-4542-aae4-33adeeb7e699

### DIFF
--- a/packages/gitbook/e2e/internal.spec.ts
+++ b/packages/gitbook/e2e/internal.spec.ts
@@ -160,20 +160,20 @@ const searchTestCases: Test[] = [
             await expect(page.getByTestId('ai-chat-input')).toBeFocused();
         },
     },
-    {
-        name: 'Ask - AI Mode: Assistant - Button',
-        url: getCustomizationURL({
-            ai: {
-                mode: CustomizationAIMode.Assistant,
-            },
-        }),
-        screenshot: false,
-        run: async (page) => {
-            await page.getByTestId('ai-chat-button').click();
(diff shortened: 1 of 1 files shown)


packages/gitbook/e2e/internal.spec.ts
```diff
@@ -160,20 +160,20 @@ const searchTestCases: Test[] = [
...
-            },
-        }),
-        screenshot: false,
-        run: async (page) => {
-            await page.getByTestId('ai-chat-button').click();
-            await expect(page.getByTestId('ai-chat')).toBeVisible();
-            await expect(page.getByTestId('ai-chat-input')).toBeFocused();
-        },
-    },
+    // {
```
conico974, line 163: Not related to your PR right ?

conico974 (approved):

packages/gitbook/e2e/internal.spec.ts
viktorrenkema: Yeah sorry forgot to add this as a comment, it kept failing and seems to be caused by something else. Ill flag it to the relevant people so they know.

PR #3796, opened 2025-11-12
Use an available alternate when switching variants in a page

taranvohra:
When switching between variants we check if the current page has an alternate link pointing to the space variant that we are switching to in order more accurately navigate the user.

The default behaviour of replacing path still exists if there was no alternate configured

### DIFF
--- a/packages/gitbook/src/components/hooks/useCurrentPageMetadata.tsx
+++ b/packages/gitbook/src/components/hooks/useCurrentPageMetadata.tsx
@@ -5,6 +5,9 @@ import * as zustand from 'zustand';
 
 /**
  * A store for the current page metadata.
+ *
+ * We use a global store because the metadata is generated and set by the Page component
+ * but needs to be accessed by other components (ex - Layout) that are not its descendants.
  */
 export const currentPageMetadataStore = zustand.create<{
     metaLinks: PageMetaLinks | null;


packages/gitbook/src/components/SitePage/SitePage.tsx
```diff
@@ -325,8 +339,15 @@ async function resolvePageMetaLinks(
 /**
  * Determine whether to resolve meta links for a site based on a percentage rollout.
  */
-export function shouldResolveMetaLinks(siteId: string): boolean {
-    const META_LINKS_PERCENTAGE_ROLLOUT = 10;
+function shouldResolveMetaLinks(siteId: string): boolean {
+    const META_LINKS_PERCENTAGE_ROLLOUT = 25;
+    const ALLOWED_SITES: Record<string, boolean> = {
+        site_CZrtk: true,
+    };
```
taranvohra, line 352: This is the E2E test site

packages/gitbook/src/components/SitePage/PageClientLayout.tsx
```diff
@@ -3,16 +3,27 @@
...
+}) {
+    const { setMetaLinks } = useCurrentPageMetadata();
+
     // We use this hook in the page layout to ensure the elements for the blocks
     // are rendered before we scroll to a hash or to the top of the page
     useScrollPage();
 
+    React.useEffect(() => {
+        setMetaLinks(pageMetaLinks);
+    }, [pageMetaLinks, setMetaLinks]);
```
SamyPesse: Wtf is this..

Just use a normal context and no state

packages/gitbook/src/components/SitePage/SitePage.tsx
```diff
@@ -76,7 +95,7 @@ export async function SitePage(props: SitePageProps) {
                         insightsDisplayContext={SiteInsightsDisplayContext.Site}
                     />
                 </div>
-                <PageClientLayout />
+                <PageClientLayout pageMetaLinks={pageMetaLinks} />
```
SamyPesse, line 102: Don't pass it there, pass it on `PageContextProvider`

packages/gitbook/src/components/SitePage/SitePage.tsx
```diff
@@ -311,7 +325,7 @@ async function resolvePageMetaLinks(
         return {
             canonical: resolvedCanonical ?? null,
             alternates: resolvedAlternates.filter(
-                (alt): alt is { href: string; language: TranslationLanguage | null } => !!alt.href
+                (alt): alt is { href: string; space: Space | null } => !!alt.href
```
SamyPesse: Do we need to pass the entire `space` object?

It'd be good to pass less infos for the client side

packages/gitbook/src/components/SitePage/SitePage.tsx
SamyPesse: Mayeb you can also enable it on dev with `process.env.NODE_ENV`

packages/gitbook/src/components/hooks/useCurrentPageMetadata.tsx
SamyPesse, line 1: Instead of yet another context, can you merge this with `CurrentPageProvider` ?

packages/gitbook/src/components/hooks/useCurrentPageMetadata.tsx
```diff
@@ -0,0 +1,23 @@
+'use client';
+import type { PageMetaLinks } from '../SitePage';
+
+import * as zustand from 'zustand';
+
+/**
+ * A store for the current page metadata.
```
SamyPesse, line 7: Add a comment to explain why we use a state instead of a context: we do it this way because we want to access it from layout component, but the data is set by a page component.

PR #3798, opened 2025-11-12
Support new "link title" page option

clairechabas:
Support new page option to set a custom link title to be displayed for the table of contents and links instead of the page title.

Useful for users to set a longer SEO-friendly title for their site while keeping the page title short in the app's UI.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -1,6 +1,5 @@
 {
   "lockfileVersion": 1,
-  "configVersion": 0,
   "workspaces": {
     "": {
       "name": "gitbook",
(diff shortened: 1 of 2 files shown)


packages/gitbook/src/components/TableOfContents/PageDocumentItem.tsx
```diff
@@ -10,6 +10,7 @@ import { ToggleableLinkItem } from './ToggleableLinkItem';
 
 export function PageDocumentItem(props: { page: ClientTOCPageDocument }) {
     const { page } = props;
+    const title = page.linkTitle || page.title;
```
SamyPesse: You don't need to this here since in packages/gitbook/src/components/TableOfContents/encodeClientTableOfContents.ts you already encode the title properly

PR #3800, opened 2025-11-12
Refactor DynamicTabs to use NavigationStatusContext

conico974:
Replace the use of useHash with NavigationStatusContext for improved navigation handling in DynamicTabs. Invoke onNavigationClick when a tab is selected and refine the scroll logic for better user experience.

### DIFF
--- a/.changeset/calm-shirts-tie.md
+++ b/.changeset/calm-shirts-tie.md
@@ -1,0 +1,5 @@
+---
+'@gitbook/react-openapi': patch
+---
+
+Re-arrange OpenAPI Scopes for OAuth2
--- a/packages/react-openapi/src/OpenAPIRequiredScopes.tsx
+++ b/packages/react-openapi/src/OpenAPIRequiredScopes.tsx
@@ -25,12 +25,6 @@ export function OpenAPIRequiredScopes(props: {
     }
 
     const scopes = selectedSecurity.schemes.flatMap((scheme) => {
-        if (scheme.type === 'oauth2') {
-            return Object.entries(scheme.flows ?? {}).flatMap(([_, flow]) =>
-                Object.entries(flow.scopes ?? {})
-            );
-        }
-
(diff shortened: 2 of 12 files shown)


packages/gitbook/src/components/DocumentView/Tabs/DynamicTabs.tsx
```diff
@@ -166,8 +167,9 @@ export function DynamicTabs(props: {
 
     // Scroll to active element in the tab.
     React.useLayoutEffect(() => {
+        console.log('Scrolling to active element:', active, hash);
```
taranvohra: Do we need this console log?

packages/gitbook/src/components/DocumentView/Tabs/DynamicTabs.tsx
```diff
@@ -112,6 +112,7 @@ export function DynamicTabs(props: {
                 touchedRef.current = true;
                 const href = `#${tab.id}`;
                 if (window.location.hash !== href) {
+                    onNavigationClick(href);
                     router.replace(href, { scroll: false });
```
taranvohra, line 116: Why do we have both?

packages/gitbook/src/components/DocumentView/Tabs/DynamicTabs.tsx
conico974: Oops no it was me debugging things out. I'll remove

packages/gitbook/src/components/DocumentView/Tabs/DynamicTabs.tsx
conico974: `useHash` needs it because of some bug in Next.
https://github.com/GitbookIO/gitbook/blob/a512c90bdaefebf921fc8d69de55deefff92ae11/packages/gitbook/src/components/hooks/useHash.tsx#L100-L104

PR #3804, opened 2025-11-13
Output generic alternates for page metadata

taranvohra:


packages/gitbook/src/components/SitePage/SitePage.tsx
```diff
@@ -166,9 +181,10 @@ export async function generateSitePageMetadata(props: SitePageProps): Promise<Me
         description: page.description ?? '',
         alternates: {
             canonical,
-            languages,
+            languages: alternates?.languages,
             types: {
                 'text/markdown': `${linker.toAbsoluteURL(linker.toPathInSpace(page.path))}.md`,
+                ...(alternates?.generic ? { '': alternates?.generic } : {}),
```
taranvohra, line 213: Right now it will output like `<link rel="alternate" type href="..." />`. I talked to folks at Vercel. They said they will fix it to include `type` only if it's truthy

SamyPesse: Let's add it as a comment on the code

PR #3820, opened 2025-11-19
Add `x-gitbook-prefix` and `x-gitbook-token-placeholder` for OpenAPI security scheme

nolannbiron:


packages/openapi-parser/src/types.ts
```diff
@@ -75,6 +75,14 @@ export interface OpenAPICustomOperationProperties {
  */
 export interface OpenAPICustomPrefillProperties {
     'x-gitbook-prefill'?: string;
+    /**
+     * Placeholder to override the default one for the security scheme.
+     */
+    'x-placeholder'?: string;
```
spastorelli: should they be `x-gitbook-*` attributes as they are gitbook specific?

packages/openapi-parser/src/types.ts
```diff
@@ -75,6 +75,14 @@ export interface OpenAPICustomOperationProperties {
...
+    /**
+     * Prefix to override the default one for the security scheme (e.g., "Bearer", "Basic", "Token").
+     */
+    'x-prefix'?: string;
```
spastorelli: Why do we need this one? The HTTP auth scheme can be set via the `scheme` property:
https://swagger.io/specification/#security-scheme-object

packages/openapi-parser/src/types.ts
spastorelli: Also placeholder is not very descriptive. Can we make it more specific?

spastorelli (approved):

packages/openapi-parser/src/types.ts
nolannbiron: @spastorelli it's just pure customization for a customer who doesn't follow standards

I told them that their prefix is non-standard but it's too much work for them to switch to a standard one

spastorelli: What's their specific/custom use case? what type of security scheme they are using and in which auth context (HTTP, Auth header or OAuth)?

The issue I see is that we are introducing a very specific custom attribute for maybe only their specific use case. So wondering if there isn't a way around it.

Also if we're still want to introduce it, for HTTP auth we should not use ~`x-gitbook-prefill`~ `x-gitbook-prefix` (as it is redundant) because `scheme` prop is a `string` prop so in that case they can use it to specify their custom scheme:
https://github.com/GitbookIO/gitbook/pull/3820/files#diff-9930c129a1dbbfa665b3d67869bd920e5a9b691d9960ecd2fbe6565077bdb268R327

nolannbiron: They use `apiKey` with `Authorization: Token xxxx` but `apiKey` doesn't take any prefix in HTTP standard

nolannbiron: I know that they can use `x-gitbook-prefill` but it doesn't feel right since it's used to pre-fill, not replace a placeholder 😕

spastorelli: Ok makes sense but I still think that ~`x-gitbook-prefill`~ `x-gitbook-prefix` shouldn't be used for HTTP auth as `scheme` prop is there for that.

spastorelli: @nolannbiron sorry actually I meant `x-gitbook-prefix` in my previous comment and not `x-gitbook-prefill`

Edited them to correct my mistake. Sorry for the confusion.

nolannbiron: ah yeah I see what you mean, HTTP is standard so we should not override it, I'll do that

PR #3823, opened 2025-11-25
Update Docs Embed with new styling and tabs

zenoachtig:
- Update the embedded frame to support sites without assistant, by including two tabs: "Assistant" and "Docs". The right tabs will be loaded by default, but can be overridden with local configuration options.
- Restyle the embed frame with a sidebar for tabs + custom actions
- Add a trademark to the embed frame
- Tweak some in-site components to display properly in the embed frame
- Update the embed docs

Tests are added in separate PR: #3830

### DIFF
--- a/packages/gitbook/src/app/sites/static/[mode]/[siteURL]/[siteData]/~gitbook/embed/demo/route.ts
+++ b/packages/gitbook/src/app/sites/static/[mode]/[siteURL]/[siteData]/~gitbook/embed/demo/route.ts
@@ -114,7 +114,6 @@ export async function GET(
             'What can I ask you?',
             'Show me tips and tricks',
         ],
-        tabs: ['assistant'],
     });
         window.GitBook('open');
     </script>


packages/embed/src/client/protocol.ts
```diff
@@ -41,16 +41,20 @@ export type GitBookEmbeddableButtonDefinition = {
...
+
+    /** Additional buttons to be displayed in the header of the GitBook embed. */
+    actions: GitBookEmbeddableActionDefinition[];
 
     /** Message to be displayed in the welcome page. */
-    welcomeMessage: string;
+    greeting: {
+        title: string;
+        subtitle: string;
+    };
```
taranvohra, line 63: We're changing the API of the embed. Is it a breaking change? Do we know how many users are already using `@gitbook/embed`?

packages/gitbook/src/components/Embeddable/EmbeddableIframeAPI.tsx
```diff
@@ -4,30 +4,48 @@ import type { GitBookEmbeddableConfiguration, ParentToFrameMessage } from '@gitb
...
 }) {
-    const { baseURL } = props;
+    const { baseURL, siteTitle } = props;
 
     const router = useRouter();
     const chatController = useAIChatController();
 
+    React.useEffect(() => {
+        embeddableConfiguration.setState({ baseURL });
+        embeddableConfiguration.setState({ siteTitle });
```
taranvohra: Same `setState` being used twice? I am confused what's the goal of this

packages/gitbook/src/components/Embeddable/EmbeddableIframeAPI.tsx
```diff
@@ -4,30 +4,48 @@ import type { GitBookEmbeddableConfiguration, ParentToFrameMessage } from '@gitb
...
+    }, [baseURL, siteTitle]);
+
+    React.useEffect(() => {
+        return chatController.on('postMessage', () => {
+            router.push(`${baseURL}/assistant`);
+        });
+    }, [router, baseURL, chatController]);
```
taranvohra, line 37: On line 68 we are doing `chatController.postMessage` along with `router.push('/assistant')` then what's the purpose of this?

packages/gitbook/src/components/Embeddable/EmbeddableIframeAPI.tsx
```diff
@@ -93,23 +114,108 @@ export function useEmbeddableConfiguration<T = GitBookEmbeddableConfiguration>(
...
     );
 }
+
+export function EmbeddableIframeTabs(props: { active?: string }) {
+    const { active = 'assistant' } = props;
+    let { tabs: configuredTabs, actions } = useEmbeddableConfiguration();
+
+    if (configuredTabs.length === 0) {
+        configuredTabs = ['assistant', 'docs'];
+    }
```
taranvohra: This feels like a code smell. Why are we mutating?

packages/gitbook/src/components/Embeddable/EmbeddableIframeAPI.tsx
```diff
@@ -93,23 +114,108 @@ export function useEmbeddableConfiguration<T = GitBookEmbeddableConfiguration>(
...
+
+    const { assistants, config } = useAI();
+
+    const router = useRouter();
+    const { baseURL, siteTitle } = useEmbeddableConfiguration();
```
taranvohra: Then we invoke it again? 🤔

packages/embed/src/client/protocol.ts
zenoachtig: I've added back `buttons` with a `@deprecated` note, and merge the arrays in https://github.com/GitbookIO/gitbook/pull/3823/files#diff-acbe55c3782083fd38ead76dbd17bcff33e9ae096184108f8a39abd46fbba411R117

As for `welcomeMessage`, it was never implemented on the Assistant side so I highly doubt users configured it, otherwise we would've gotten feedback on it. I feel pretty safe renaming it.

packages/gitbook/src/components/Embeddable/EmbeddableIframeAPI.tsx
```diff
@@ -4,30 +4,47 @@ import type { GitBookEmbeddableConfiguration, ParentToFrameMessage } from '@gitb
...
+    siteTitle: string;
 }) {
-    const { baseURL } = props;
+    const { baseURL, siteTitle } = props;
 
     const router = useRouter();
     const chatController = useAIChatController();
 
+    React.useEffect(() => {
+        embeddableConfiguration.setState({ baseURL, siteTitle });
```
taranvohra: I don't see the need to store `baseURL` and `siteTitle` in a global store when we have it on the `context`. Why can't we just use it from there?

packages/gitbook/src/components/Embeddable/EmbeddableIframeAPI.tsx
```diff
@@ -4,30 +4,47 @@ import type { GitBookEmbeddableConfiguration, ParentToFrameMessage } from '@gitb
...
+    }, [baseURL, siteTitle]);
+
+    React.useEffect(() => {
+        return chatController.on('postMessage', () => {
+            router.push(`${baseURL}/assistant`);
+        });
+    }, [router, baseURL, chatController]);
```
taranvohra, line 37: Wouldn't this keep pushing the same entry over and over everytime a message is posted on the chat?

packages/gitbook/src/components/Embeddable/EmbeddableIframeAPI.tsx
zenoachtig: This is a client component, we don't want to send the whole context to the client so have to be selective. Unless I fundamentally misunderstand how `context` is supposed to work in GBO

packages/gitbook/src/components/Embeddable/EmbeddableIframeAPI.tsx
zenoachtig: You're right. Not a problem in practice, but absolutely wasteful. I've switched it around to use `on.('open')` instead, that seems like the more semantic thing to watch anyway.

packages/gitbook/src/components/Embeddable/EmbeddableIframeAPI.tsx
taranvohra: What I meant was they could be passed as props from the context object (not the whole thing) from the server component to the client. The client components are all descendants so shouldn't be an issue. Just a thought

taranvohra (approved):

PR #3828, opened 2025-11-27
Fix flaky tests in AI Chat greeting

zenoachtig:


packages/gitbook/e2e/internal.spec.ts
```diff
@@ -182,6 +189,13 @@ const searchTestCases: Test[] = [
             await page.getByTestId('ai-chat-button').click();
             await expect(page.getByTestId('ai-chat')).toBeVisible();
             await expect(page.getByTestId('ai-chat-input')).toBeFocused();
+            // Override text content for visual consistency in screenshots
+            await page.evaluate(() => {
+                const greeting = document.querySelectorAll('[data-testid="ai-chat-time-greeting"]');
```
taranvohra: Why do we have multiple greeting?

taranvohra (approved):

packages/gitbook/e2e/internal.spec.ts
zenoachtig: We don't, I guess I can make it a single selector actually. Just figured I'd do it to be safe, but I'll adapt it

PR #3831, opened 2025-11-27
Support for hidden sections

nolannbiron:


nolannbiron:
@conico974 it's enabled only for one customer and they don't really care about sitemap, it's already working for variants but I'll double check if it breaks anything for sections/groups

PR #3835, opened 2025-12-02
Add SideSheet component, refactor TOC and AIChat to use it

zenoachtig:
This pull request introduces a new `SideSheet` component and refactors the TOC and AIChat to use it.

# SideSheet component
- Adds `SideSheet` component, responsible for rendering a component in a fixed sidebar (with optional backdrop) on the left or right of the screen. It can be controlled with an `open` prop (needed for `AIChat`) or watch for changes on the body (needed for `TableOfContents`).
- Refactored `TableOfContents` and `AIChat` to use `SideSheet`

# ScrollContainer component
- Refactored `TableOfContents` to use `ScrollContainer`, removed the custom `TOCScrollContainer` in favour of the standard component we use elsewhere. This gives us goodies like faded edges and scroll buttons, in addition to standardising the component.
- Updated `active` prop on `ScrollContainer` to take any query selector instead of an ID.

# Other layout updates
- Moved layout of TOC's `innerHeader` to live in one place, namely `SpaceLayout`, instead of splitting across two files
- Simplified `HeaderMobileMenu` to no longer have 
- Updated layout components (`CustomizationRootLayout`, `SiteDynamicLayout`, `SiteStaticLayout`) to accept both a `htmlClassName` and `bodyClassName`, needed to set some sheet-related classes that only work on `html` or `body`.
- Improved support for devices with notches by adding safe-area-insets and `viewportFit: cover` to the viewport meta tag.
- Simplified `HeaderMobileMenu` button by removing scroll tracking and state management now it's always occluded by a sidesheet.
- Fixed minor UI issues in `AnnouncementBanner` when using safe-area-insets

### DIFF
--- a/packages/gitbook/src/components/primitives/SideSheet.tsx
+++ b/packages/gitbook/src/components/primitives/SideSheet.tsx
@@ -73,19 +73,20 @@ export function SideSheet(
 
     // Track if component has been opened to prevent initial animation
     React.useEffect(() => {
+        let timer: ReturnType<typeof setTimeout> | undefined;
         if (isOpen) {
             wasOpenRef.current = true;
             setShouldHide(false);
         } else if (wasOpenRef.current) {
-            // Delay hiding until after exit animation completes
-            const timer = setTimeout(() => {
+            timer = setTimeout(() => {
                 setShouldHide(true);
             }, ANIMATION_DURATION);
-            return () => clearTimeout(timer);
         } else {
-            // Never been opened, hide immediately
             setShouldHide(true);
         }
+        return () => {
+            if (timer) clearTimeout(timer);
+        };
     }, [isOpen]);
 
     const handleClose = React.useCallback(() => {


packages/gitbook/src/components/primitives/SideSheet.tsx
```diff
@@ -0,0 +1,275 @@
...
+        modal?: true | false | 'mobile';
+        /**
+         * Controls visibility. If provided, component is controlled (parent manages state).
+         * If undefined, component is uncontrolled (manages its own state).
+         */
+        open?: boolean;
+        /** Called when the open state changes. Receives the new state (true/false). Only used in controlled mode. */
+        onOpenChange?: (open: boolean) => void;
+        /** Show a backdrop overlay when modal */
+        withScrim?: boolean;
```
nolannbiron: I think `withOverlay` is clearer

packages/gitbook/src/components/primitives/SideSheet.tsx
```diff
@@ -0,0 +1,275 @@
...
+            // Delay hiding until after exit animation completes
+            const timer = setTimeout(() => {
+                setShouldHide(true);
+            }, ANIMATION_DURATION);
+            return () => clearTimeout(timer);
+        } else {
+            // Never been opened, hide immediately
+            setShouldHide(true);
+        }
+    }, [isOpen]);
```
nolannbiron, line 90:
```suggestion
React.useEffect(() => {
    let timer: ReturnType<typeof setTimeout> | undefined;

    if (isOpen) {
        wasOpenRef.current = true;
        setShouldHide(false);
    } else if (wasOpenRef.current) {
        timer = setTimeout(() => {
            setShouldHide(true);
        }, ANIMATION_DURATION);
    } else {
        setShouldHide(true);
    }

    return () => {
        if (timer) clearTimeout(timer);
    };
}, [isOpen]);
```

PR #3839, opened 2025-12-03
Remove corner radius of hint block with heading

zenoachtig:
# Before
CleanShot 2025-12-03 at 10 36 57@2x

# After
CleanShot 2025-12-03 at 10 36 45@2x

### DIFF
--- a/packages/gitbook/src/components/DocumentView/Hint.tsx
+++ b/packages/gitbook/src/components/DocumentView/Hint.tsx
@@ -32,10 +32,9 @@ export function Hint({
             className={tcls(
                 'hint',
                 'transition-colors',
-                'rounded-md',
-                hasHeading ? 'rounded-l-none!' : null,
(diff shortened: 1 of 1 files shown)


packages/gitbook/src/components/DocumentView/Hint.tsx
```diff
@@ -33,7 +33,7 @@ export function Hint({
                 'hint',
                 'transition-colors',
                 'rounded-md',
-                hasHeading ? 'rounded-l-sm' : null,
+                hasHeading ? 'rounded-l-none!' : null,
```
spastorelli: nit: Since we apply `'rounded-md'` just above for all case, can't we just apply what's needed in case of `hasHeading` and fallback to `'rounded-md'` when no heading? That would avoid having to make `rounded-l-none` `important` in case of heading?

PR #3845, opened 2025-12-04
Fix link title fallback in ToC item

clairechabas:
We need to account for the `linkTitle` being an empty string.


packages/gitbook/src/components/TableOfContents/encodeClientTableOfContents.ts
```diff
@@ -72,7 +72,7 @@ export async function encodeClientTableOfContents(
                 result.push(
                     removeUndefined({
                         id: page.id,
-                        title: page.linkTitle ?? page.title,
+                        title: page.linkTitle || page.title,
```
conico974, line 75: Could you add a small comment about that

conico974 (commented):
You missed the one in `packages/gitbook/src/lib/references.tsx`

PR #3847, opened 2025-12-05
Deference reusable content at the API level

SamyPesse:
## Demo

You can see the issue RND-8768 on https://gitbook-open-e2e-sites.gitbook.io/reusable-contents/
and fixed on https://gitbook-v2-eep7gqtc5-gitbook.vercel.app/url/gitbook-open-e2e-sites.gitbook.io/reusable-contents

## Tasks

1. [x] Implement new data API to fetch a document page with reusable content dereferenced
2. [x] For a % of traffic, use it instead of the old method
3. [x] If `JSONDocument.meta.token` is defined, used this to resolve content-refs within the document

### DIFF
--- a/packages/gitbook/e2e/internal.spec.ts
+++ b/packages/gitbook/e2e/internal.spec.ts
@@ -1129,7 +1129,7 @@ const testCases: TestsCase[] = [
         tests: [
             {
                 name: 'All cases',
-                url: '/',
+                url: '',
             },
         ],
     },


packages/gitbook/src/lib/references.tsx
```diff
@@ -77,6 +78,42 @@ export interface ResolveContentRefOptions {
...
+    options: ResolveContentRefOptions = {}
+): Promise<ResolvedContentRef | null> {
+    if (isContentRefInDifferentSpace(contentRef, context)) {
+        const withinSite = !!getBestTargetSpaceFromSite(context, contentRef.space);
+        if (!withinSite) {
+            // When the content ref points to some content outside the current site,
+            // we use the potentially provided token to resolve the content ref in the target space.
+            // This is the case when the document is composed of reusable blocks from spaces outside the current site.
+            const token = document.meta?.token;
+            if (token) {
```
conico974, line 100: Do we need to check if it's a reusable block or is the fact that we have a token enough ?

packages/gitbook/src/lib/references.tsx
```diff
@@ -77,6 +78,42 @@ export interface ResolveContentRefOptions {
...
+ */
+export async function resolveContentRefInDocument(
+    document: JSONDocument,
+    contentRef: ContentRef,
+    context: GitBookAnyContext,
+    options: ResolveContentRefOptions = {}
+): Promise<ResolvedContentRef | null> {
+    if (isContentRefInDifferentSpace(contentRef, context)) {
+        const withinSite = !!getBestTargetSpaceFromSite(context, contentRef.space);
+        if (!withinSite) {
```
conico974, line 95: I'm wondering if this one could backfire with people having 100 different space not in the site for all their reusable block ?

packages/gitbook/src/lib/references.tsx
SamyPesse: it's enough, the idea of the `meta.token` is that it's an added token to fetch any resources from the document. GBO doesn't need know anything about reusable content anymore.

packages/gitbook/src/lib/references.tsx
SamyPesse: Already the case right now; we are fetching all of these spaces from GBO.

Basically this PR goes from:

>  Each reusable block has its own meta.token and GBO fetches the space, its revision, the document

to

> GBO fetches a page document, the API fetches the different reusable content and inlines them, and only images/files require fetching the targeted space, and for this we provided an extended token in `meta.token`

PR #3852, opened 2025-12-06
chore: update dependencies and remove Suspense on ISR requests

conico974:
- Bump React and React-DOM versions to 19.0.1 in package.json.
- Update @opennextjs/cloudflare to version 1.14.4 in packages/gitbook/package.json.
- Update Next.js version to 15.4.8 in packages/gitbook/package.json.
- Modify SitePage and EmbeddableDocsPage components to accept an `isSSR` prop for better server-side rendering handling.
- Introduce OptionalSuspense component to conditionally wrap children in a Suspense boundary based on SSR status, improving UX for ISR pages.
- Adjust page components to utilize the new `isSSR` prop for consistent rendering behavior.

### DIFF
--- a/packages/gitbook/src/app/sites/static/[mode]/[siteURL]/[siteData]/(content)/[pagePath]/page.tsx
+++ b/packages/gitbook/src/app/sites/static/[mode]/[siteURL]/[siteData]/(content)/[pagePath]/page.tsx
@@ -18,7 +18,7 @@ export default async function Page(props: PageProps) {
     const { context } = await getStaticSiteContext(params);
     const pathname = getPagePathFromParams(params);
 
-    return <SitePage context={context} pageParams={{ pathname }} isSSR={false} />;
+    return <SitePage context={context} pageParams={{ pathname }} staticRoute />;
 }
 
 export async function generateViewport(props: PageProps): Promise<Viewport> {
--- a/packages/gitbook/src/components/PageBody/OptionalSuspense.tsx
+++ b/packages/gitbook/src/components/PageBody/OptionalSuspense.tsx
@@ -6,13 +6,13 @@ import React from 'react';
  * causing flickering and bad UX.
  */
 export default function OptionalSuspense(props: {
-    isSSR: boolean;
+    staticRoute: boolean;
     fallback: React.ReactNode;
     children: React.ReactNode;
 }) {
-    const { isSSR, fallback, children } = props;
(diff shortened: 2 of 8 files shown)


packages/gitbook/src/app/sites/static/[mode]/[siteURL]/[siteData]/(content)/[pagePath]/page.tsx
```diff
@@ -18,7 +18,7 @@ export default async function Page(props: PageProps) {
     const { context } = await getStaticSiteContext(params);
     const pathname = getPagePathFromParams(params);
 
-    return <SitePage context={context} pageParams={{ pathname }} />;
+    return <SitePage context={context} pageParams={{ pathname }} isSSR={false} />;
```
Comment (taranvohra) [inline packages/gitbook/src/app/sites/static/[mode]/[siteURL]/[siteData]/(content)/[pagePath]/page.tsx]: Why `isSSR` is `false`? I'm not sure I understand the solution

Comment (conico974) [reply packages/gitbook/src/app/sites/static/[mode]/[siteURL]/[siteData]/(content)/[pagePath]/page.tsx]: Because it's a static ISR route. The problem is that the Suspense boundary triggers every single time here (even when cached) causing flickering and bad UX

packages/gitbook/src/components/PageBody/OptionalSuspense.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+
+/**
+ * Conditionally wrap children in a Suspense component based on SSR status.
+ * We need to do that because otherwise ISR pages will trigger the suspense boundary on every request,
+ * causing flickering and bad UX.
+ */
+export default function OptionalSuspense(props: {
+    isSSR: boolean;
```
SamyPesse: The prop name is a bit misleading, both are SSR.

Should it be:


```suggestion
    staticRoute: boolean;
```

conico974: I guess it make sense, I'm probably a bit too biased because of all the other Next terminology :smile:

packages/gitbook/src/app/sites/static/[mode]/[siteURL]/[siteData]/(content)/[pagePath]/page.tsx
Comment (taranvohra) [reply packages/gitbook/src/app/sites/static/[mode]/[siteURL]/[siteData]/(content)/[pagePath]/page.tsx]: Okay makes sense. The name was confusing

PR #3857, opened 2025-12-10
Create `<Input>` component

zenoachtig:
Standardise on a new `<Input>` component that takes care of submitting, clearing, and validating. Can be controlled or uncontrolled, multiline (renders as `textarea`) or single-line (renders as `input`).

https://github.com/user-attachments/assets/3795f589-c7d7-49b8-b1a3-35438c8cf85f

All existing inputs on the site have been reworked with all functionality intact and minimal style changes.

PageFeedbackForm has been updated more substantially, using the new styling.
CleanShot 2025-12-12 at 15 16 55@2x

### DIFF
--- a/packages/gitbook/src/components/primitives/Input.tsx
+++ b/packages/gitbook/src/components/primitives/Input.tsx
@@ -250,6 +250,8 @@ export const Input = React.forwardRef<InputElement, InputProps>((props, passedRe
                     <input
                         {...inputProps}
                         ref={ref as React.RefObject<HTMLInputElement>}
+                        type="text"
+                        size={1} // Size controls the intrinsic width of the input, but we want to control the width with CSS
                         {...(htmlProps as React.InputHTMLAttributes<HTMLInputElement>)}
                     />
                 )}


packages/gitbook/src/components/primitives/Input.tsx
```diff
@@ -0,0 +1,313 @@
...
+                setValue(newValue);
+            }
+            onChange?.(event);
+
+            // Reset submitted state when user edits the value to allow re-submission
+            if (submitted) {
+                setSubmitted(false);
+            }
+
+            if (multiline && resize && 'current' in ref && ref.current) {
```
gregberge:
```suggestion
            if (multiline && resize && ref.current) {
```

packages/gitbook/src/components/primitives/Input.tsx
```diff
@@ -0,0 +1,313 @@
...
+                ref.current.style.height = 'auto';
+                ref.current.style.height = `${ref.current.scrollHeight}px`;
```
gregberge: Better to use a state for that, React is not imperative but reactive.

Set a state with the height, based on that state to render the style.

packages/gitbook/src/components/primitives/Input.tsx
```diff
@@ -0,0 +1,313 @@
...
+            }
+        };
+
+        const handleClear = () => {
+            if (!('current' in ref) || !ref.current) return;
```
gregberge:
```suggestion
            if (!ref.current) return;
```

packages/gitbook/src/components/primitives/Input.tsx
```diff
@@ -0,0 +1,313 @@
...
+
+            const syntheticEvent = {
+                target: { value: '' },
+                currentTarget: ref.current,
+            } as React.ChangeEvent<HybridInputElement>;
+
+            handleChange(syntheticEvent);
```
gregberge: Bad idea. My suggestion for that is to create a `onInputChange` callback that provides only the value. And then you call safely call it here. But faking the change event is not good.

packages/gitbook/src/components/primitives/Input.tsx
```diff
@@ -0,0 +1,313 @@
...
+        };
+
+        const handleClick = () => {
+            if (!('current' in ref)) return;
```
gregberge:
```suggestion
```

packages/gitbook/src/components/primitives/Input.tsx
```diff
@@ -0,0 +1,313 @@
...
+            ref.current?.focus();
+        };
+
+        const handleSubmit = () => {
+            if (hasValue && onSubmit) {
+                onSubmit(value);
+                setSubmitted(true);
+                if (!isControlled && 'current' in ref && ref.current) {
```
gregberge:
```suggestion
                if (!isControlled && ref.current) {
```

packages/gitbook/src/components/primitives/Input.tsx
```diff
@@ -0,0 +1,313 @@
...
+                    ref.current.value = '';
```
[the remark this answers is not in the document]
gregberge: Yeah as GPT say, you can't do imperative things like that.

If you want to have an internal state that can be controlled, you can use something like [useControlledState](https://github.com/adobe/react-spectrum/blob/main/packages/%40react-stately/utils/src/useControlledState.ts)

packages/gitbook/src/components/primitives/Input.tsx
```diff
@@ -0,0 +1,313 @@
...
+
+        const handleKeyDown = (event: React.KeyboardEvent<HybridInputElement>) => {
+            if (event.key === 'Enter' && !event.shiftKey && hasValue) {
+                event.preventDefault();
+                handleSubmit();
+            } else if (event.key === 'Escape') {
+                event.preventDefault();
+                event.currentTarget.blur();
+            }
+            onKeyDown?.(event);
```
gregberge:
```suggestion
            onKeyDown?.(event);
            // If the user wants to handle the keydown by itself, we let him do it.
            if (event.defaultPrevented()) {
              return
            }
            if (event.key === 'Enter' && !event.shiftKey && hasValue) {
                event.preventDefault();
                handleSubmit();
            } else if (event.key === 'Escape') {
                event.preventDefault();
                event.currentTarget.blur();
            }
            
```

packages/gitbook/src/components/primitives/Input.tsx
```diff
@@ -0,0 +1,313 @@
...
+import { tString, useLanguage } from '@/intl/client';
+import { tcls } from '@/lib/tailwind';
+import { Icon, type IconName } from '@gitbook/icons';
+import React, { type ReactNode } from 'react';
+import { Button, type ButtonProps } from './Button';
+import { KeyboardShortcut, type KeyboardShortcutProps } from './KeyboardShortcut';
+
+export type InputProps = {
+    label: string;
+    tag?: 'div' | 'span';
```
gregberge: Why do you need that?

packages/gitbook/src/components/primitives/Input.tsx
```diff
@@ -0,0 +1,313 @@
...
+                                '-mx-1.5 hidden shrink-0 animate-fade-in p-1.5 text-tint',
+                                multiline ? '-my-0.25' : '-my-1.5',
+                                hasValue ? 'group-focus-within/input:flex' : '',
+                                typeof clearButton === 'object' ? clearButton.className : ''
+                            )}
+                        />
+                    ) : null}
+                    {multiline ? (
+                        <textarea
+                            {...(inputProps as React.TextareaHTMLAttributes<HTMLTextAreaElement>)}
```
gregberge: You should not have to do `as`, if you have to it means something is wrong in terms of typing.

packages/gitbook/src/components/primitives/Input.tsx
zenoachtig: The input can appear as both a block element (in site layout) as well as an inline element (which will be used by the new inline search input rendered by button actions). I considered an `inline: boolean` instead but felt this was more semantically correct.

gregberge: Hmm yeah I think inline is more explicit to be honest. And inside you should not use it it's only for the outer element anyway.

PR #3897, opened 2026-01-11
Clean up button and input across the app

zenoachtig:
This PR...
- Properly standardises the dimensions of `Button` and `Input` to be in line with each other
  - `large` and `medium` for prominent buttons that scale with text
  - `small` and `xsmall` for more subtle UI buttons, always a consistent size 

CleanShot 2026-01-10 at 22 34 23@2x
CleanShot 2026-01-10 at 22 34 38@2x

- Introduces a `small` size for `Input` that matches the dimensions of `small` Buttons
- Cleans up buttons and inputs across the app to be the right sizes.
- Introduces a set of text-size-dependent `spacing` utilities in Tailwind
  - Fixes a longstanding issue where `iconOnly` buttons would render with different dimensions than ones with labels. Each button size now has a consistent height, since icons have the necessary text-size-dependent margins applied to them.
- Introduces a `container` property to the `useIsMobile` utility check. When provided, it switches the check to use a container query instead of a viewport query.
  - Applies the `useIsMobile` check to the `AIChatButton` in the header

### DIFF
--- a/packages/gitbook/src/components/AIChat/AIChatButton.tsx
+++ b/packages/gitbook/src/components/AIChat/AIChatButton.tsx
@@ -6,6 +6,8 @@ import { useIsMobile } from '../hooks/useIsMobile';
 import { Button } from '../primitives';
 import { KeyboardShortcut } from '../primitives/KeyboardShortcut';
 
+const MOBILE_BREAKPOINT = 672; // 42rem, equal to Tailwind's @max-2xl container breakpoint
+
 /**
  * Button to open/close the AI chat.
  */
@@ -16,7 +18,7 @@ export function AIChatButton(props: {
 }) {
     const { assistant, showLabel = true, withShortcut = true } = props;
     const language = useLanguage();
-    const isMobile = useIsMobile(672, '#header-content');
+    const isMobile = useIsMobile(MOBILE_BREAKPOINT, '#header-content');
 
     return (
         <Button
--- a/packages/gitbook/src/components/hooks/useIsMobile.tsx
+++ b/packages/gitbook/src/components/hooks/useIsMobile.tsx
@@ -9,6 +9,7 @@ export function useIsMobile(breakpoint = 1024, container?: string): boolean {
         if (container) {
             const containerElement = document.querySelector(container);
             if (containerElement) {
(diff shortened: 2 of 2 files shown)


packages/gitbook/src/components/AIChat/AIChatButton.tsx
```diff
@@ -15,15 +16,15 @@ export function AIChatButton(props: {
 }) {
     const { assistant, showLabel = true, withShortcut = true } = props;
     const language = useLanguage();
+    const isMobile = useIsMobile(672, '#header-content');
```
spastorelli: I would use a constant to describe what `672` means.

packages/gitbook/src/components/hooks/useIsMobile.tsx
```diff
@@ -2,10 +2,29 @@
 
 import * as React from 'react';
 
-export function useIsMobile(breakpoint = 1024): boolean {
+export function useIsMobile(breakpoint = 1024, container?: string): boolean {
     const [isMobile, setIsMobile] = React.useState(false);
 
     React.useEffect(() => {
+        if (container) {
```
spastorelli, line 9: Why not rely the `change` event listener logic that's. already below instead of adding a resize observer on top of the event listener?

packages/gitbook/src/components/primitives/Button.tsx
spastorelli, line 1: nit: It would be nice to add (not in this PR of course) some visual tests/storybook dedicated to all the `components/primitives`.

spastorelli (approved):

packages/gitbook/src/components/hooks/useIsMobile.tsx
zenoachtig: In the case of a container we need a ResizeObserver, since the `matchMedia.change` listener might fire at different times than the container size changes. A doc might be unchanged in terms of matchMedia (`900px` → `950px`) but the container (that might be sized to only part of the screen, such as when the AI Chat sidebar is visible) might have passed the breakpoint (`650px` → `700px`) in that time.

spastorelli: Makes sense 👍 , thanks for the details.

PR #3904, opened 2026-01-13
Standardise toggle chevrons across the app

zenoachtig:
We have a bunch of places where we need an animated chevron that changes based on open/closed state.

This PR introduces a new `ToggleChevron` component to standardise the use across the app.

### DIFF
--- a/packages/gitbook/e2e/internal.spec.ts
+++ b/packages/gitbook/e2e/internal.spec.ts
@@ -295,10 +295,10 @@ const testCases: TestsCase[] = [
                     await expect(navigationLink).not.toBeVisible();
 
                     // Find and click the chevron element that is next to "Editor" in the TOC
-                    // It is a span inside the link
+                    // It is a button inside the link
                     const editorChevron = page
                         .getByRole('link', { name: 'Editor' })
-                        .locator('span');
+                        .locator('button');
                     await editorChevron.click();
 
                     // At this stage the link should still not be visible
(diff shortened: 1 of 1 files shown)


packages/gitbook/src/components/primitives/ToggleChevron.tsx
```diff
@@ -0,0 +1,44 @@
+import { type ClassValue, tcls } from '@/lib/tailwind';
+import { Icon, type IconName } from '@gitbook/icons';
+
+/**
+ * Animated chevron to display in dropdowns and other toggleable elements.
+ * Uses the `data-[state=open]` attribute to animate the chevron, or you can pass it an `open` prop to control the state.
+ */
+export function ToggleChevron(props: {
+    open?: boolean;
+    type?: 'down-to-up' | 'right-to-down'; // The direction of the chevron when open.
```
clairechabas: I'd suggest more precise names for clarity sake as "open" alone can be mistaken for a callback prop and "type" is too generic and easily mistaken with typing:

```suggestion
    isOpen?: boolean;
    direction?: 'down-to-up' | 'right-to-down';
```

gregberge (approved):

brinkcoin486 (commented):
Commère,

brinkcoin486:
Voleur

packages/gitbook/src/components/primitives/ToggleChevron.tsx
zenoachtig: I renamed `type` to `direction`, makes a lot more sense 👌 

As for `open`, I'm following [Radix's pattern](https://www.radix-ui.com/primitives/docs/components/tooltip#root) of using `open` to specify the controlled state of a component. In my opinion a callback prop would be `onOpen`, and `isOpen` would be more of an internal convenience variable.

PR #3914, opened 2026-01-19
Serve markdown version of the page depending on the accept type

SamyPesse:


packages/gitbook/src/middleware.ts
```diff
@@ -588,21 +593,21 @@ function encodePathInSiteContent(rawPathname: string): {
...
 
+    // We skip encoding for paginated llms-full.txt pages (i.e. llms-full.txt/100)
+    if (pathname.match(LLMS_FULL_PATH_REGEX)) {
+        return { pathname, routeType: 'static' };
+    }
+
     // If the pathname is a markdown file, we rewrite it to ~gitbook/markdown/:pathname
-    if (pathname.match(MARKDOWN_PATH_REGEX)) {
-        const pagePathWithoutMD = pathname.slice(0, -3);
+    if (pathname.match(MARKDOWN_PATH_REGEX) || isMarkdownPreferred(request)) {
```
conico974: Shouldn't `isMarkdownPreferred` be inside default (or at least as the last checked things)
All other route should take precedence over this one (i.e. llms.txt, mcp etc...)

packages/gitbook/e2e/internal.spec.ts
```diff
@@ -677,70 +677,6 @@ const testCases: TestsCase[] = [
             },
         ],
     },
-    {
-        name: 'Markdown page',
```
conico974, line 681: What was the problem with these tests ?

packages/gitbook/src/middleware.ts
```diff
@@ -666,3 +671,16 @@ async function writeResponseCookies<R extends NextResponse>(
 
     return response;
 }
+
+const MARKDOWN_MEDIA_TYPES = ['text/plain', 'text/markdown', 'text/x-markdown'];
```
conico974: Maybe drop text/plain for now and just log it in case it is text/plain (along with the user agent)

packages/gitbook/e2e/internal.spec.ts
SamyPesse: I moved them to the vitest who are more direct/easier to run

packages/gitbook/tests/llms.test.ts
```diff
@@ -0,0 +1,56 @@
...
+        const response = await fetch(
+            getContentTestURL('https://gitbook.gitbook.io/test-gitbook-open/llms-full.txt'),
+            {
+                headers: {
+                    Accept: 'text/markdown',
+                },
+            }
+        );
+
+        expect(response.status).toBe(200);
```
conico974, line 52: https://github.com/GitbookIO/gitbook/blob/0b931241867f3939858105bbdbe60800d06550f5/packages/gitbook/src/middleware.ts#L567
It doesn't include the basic `llms-full.txt`, only the paginated ones

PR #3915, opened 2026-01-19
Add decodeURLPath function for URL decoding and validation

conico974:
Implement a new function to decode URL path components, ensuring validation against invalid characters and limiting decoding iterations.

This will now error on path with too deeply nested percent encoding, which should reduces 404 for voluntarily bad URL.
It can also handle percent encoded character properly in the URL.

### DIFF
--- a/packages/gitbook/e2e/internal.spec.ts
+++ b/packages/gitbook/e2e/internal.spec.ts
@@ -991,6 +991,11 @@ const testCases: TestsCase[] = [
                 url: 'blocks/columns',
                 run: waitForCookiesDialog,
             },
+            {
+                name: 'Mermaid',
+                url: 'blocks/mermaid',
+                run: waitForCookiesDialog,
+            },
         ],
     },
     {
--- a/packages/gitbook/src/components/DocumentView/StepperStep.tsx
+++ b/packages/gitbook/src/components/DocumentView/StepperStep.tsx
@@ -48,7 +48,7 @@ export function StepperStep(props: BlockProps<DocumentBlockStepperStep>) {
                 {...contextProps}
                 nodes={block.nodes}
                 ancestorBlocks={[...ancestorBlocks, block]}
-                style={['flex-1 pb-6 [&>*+*]:mt-5', marginAdjustClassName]}
+                style={['min-w-0 flex-1 pb-6 [&>*+*]:mt-5', marginAdjustClassName]}
             />
         </div>
     );


packages/gitbook/src/lib/data/urls.ts
```diff
@@ -127,6 +128,93 @@ export function normalizeURL(url: URL) {
...
 }
 
+/**
+ * This function checks if a decoded URL path segment contains characters that are not allowed
+ * in GitBook content paths. These characters are valid in generic RFC 3986 URL paths, but are
+ * rejected here as an application-level constraint for GitBook routing and security.
+ * https://developer.mozilla.org/en-US/docs/Glossary/Percent-encoding
+ * "%" itself is excluded because it could be part of percent-encoding and may require decoding.
+ */
+function containsInvalidURLCharacters(segment: string): boolean {
```
spastorelli, line 138: Have you tested this with a Rison-encoded payload passed as a query parameter (like how customization would use it)? Could you add a test case for that scenario?

Also, as a quick due diligence check, could you add a test using a JWT token as the value as well? I don’t expect any conflicting characters since JWTs only use character like `.` `_` (on top of alpha characters) and you don't have them as invalid character but just to cover us against regression in case someone decided to add these as invalid characters.

packages/gitbook/src/lib/data/urls.ts
```diff
@@ -127,6 +128,93 @@ export function normalizeURL(url: URL) {
...
+ * Decode the url path component, we redirect URLs with encoded path components
+ * so that we don't end up with multiple URLs for the same content (especially important because of caching).
+ * If after decoding the path contains invalid characters, we throw an error.
+ * We limit the number of decoding passes to 2 to prevent DoS attacks via deeply-nested
+ * percent-encoding. Legitimate URLs are at most singly encoded; double-encoding covers
+ * any reasonable proxy behaviour.
+ */
+export function decodeURLPath(url: URL): URL {
+    // Reject excessively long paths up-front to bound per-request work.
+    if (url.pathname.length > 2048) {
```
spastorelli, line 172: why that restriction? with JWTs (with some payloads) combined with long paths can't we run into that restricting quickly?

packages/gitbook/src/lib/data/urls.ts
conico974: I could add some test for that yeah, but this should only applies to the pathname, not on the query.
And it only applies pre-middleware which should not contain any Rison-encoded payload in the path itself

packages/gitbook/src/lib/data/urls.ts
conico974: To avoid any useless computation, we already discard pathname that are bigger than 512 char in length during resolution https://github.com/GitbookIO/gitbook/blob/9be66e3e8a67b14c44c1c3438eb9b986fcc4a1b0/packages/gitbook/src/components/SitePage/fetch.ts#L56
Here it has to be bigger because of the potential encoding, the number is a bit arbitrary though

packages/gitbook/src/lib/data/urls.ts
spastorelli: Ha sorry missed that. Makes sense.

PR #3916, opened 2026-01-20
Support social accounts customization option

zenoachtig:
CleanShot 2026-01-20 at 14 45 48@2x


packages/gitbook/src/components/Footer/Footer.tsx
```diff
@@ -137,6 +138,25 @@ export function Footer(props: { context: GitBookSiteContext }) {
                             ) : null
                         }
 
+                        {
+                            // Social Links
+                            customization.socialAccounts.filter(
+                                (account) => account.display?.footer
```
SamyPesse: Extract this list at the top, then check its length here and map on it below

PR #3924, opened 2026-01-21
Prevent duplicate Content-Type headers in OpenAPI code samples

nolannbiron:


packages/react-openapi/src/OpenAPICodeSample.tsx
```diff
@@ -125,8 +125,9 @@ function generateCodeSamples(props: {
     const mediaTypeRendererFactories = Object.entries(requestBody?.content ?? {}).map(
         ([mediaType, mediaTypeObject]) => {
             return (generator: CodeSampleGenerator) => {
+                // Remove any existing content-type header (case-insensitive) before adding the proper one
                 const mediaTypeHeaders = {
-                    ...genericHeaders,
+                    ...normalizeHeaders(genericHeaders, ['content-type']),
                     'Content-Type': mediaType,
                 };
```
gregberge: Can you instead write a function `mergeHeaders` that has the signature `mergeHeaders(genericHeaders, { 'Content-Type': mediaType })`?

You can put this in util and write a separated test for it. Like this it works but it's messy.

PR #3925, opened 2026-01-21
Add support for custom cookie banner registration

nolannbiron:
This PR introduces support for custom cookie banner handling. This is a v1 and is subject to change once we start using it.

### DIFF
--- a/packages/browser-types/src/index.ts
+++ b/packages/browser-types/src/index.ts
@@ -88,7 +88,7 @@ export type GitBookGlobal = {
      * Whether the user has granted cookie consent.
      * Returns true if cookies are granted, false if rejected or unknown.
      */
-    hasGrantedCookies: () => boolean;
+    hasApprovedCookies: () => boolean;
 };
 
 declare global {
--- a/packages/gitbook/src/components/Integrations/LoadIntegrations.tsx
+++ b/packages/gitbook/src/components/Integrations/LoadIntegrations.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import { hasGrantedCookies, setCookiesTracking } from '@/components/Insights';
+import { hasApprovedCookies, setCookiesTracking } from '@/components/Insights';
 import type {
     GitBookGlobal,
     GitBookIntegrationEvent,
@@ -102,7 +102,7 @@ if (typeof window !== 'undefined') {
                 },
             });
         },
-        hasGrantedCookies: hasGrantedCookies,
+        hasApprovedCookies: hasApprovedCookies,
     };
     window.GitBook = gitbookGlobal;
 }
--- a/packages/gitbook/src/components/Insights/cookies.ts
+++ b/packages/gitbook/src/components/Insights/cookies.ts
@@ -36,7 +36,7 @@ export function isCookiesTrackingDisabled() {
  * Return true if cookies are accepted.
  * Return false if cookies are rejected or unknown.
  */
-export function hasGrantedCookies() {
+export function hasApprovedCookies() {
     const state = getBrowserCookie(GRANTED_COOKIE);
     return state === 'yes';
 }


packages/gitbook/src/components/Integrations/LoadIntegrations.tsx
```diff
@@ -26,6 +26,29 @@ export const integrationsAssistantTools = zustand.createStore<{
...
+
+// Store for custom cookie banner registration
+export const customCookieBannerStore = zustand.createStore<CustomCookieBannerStore>(() => {
+    return {
+        hasCustomCookieBanner: false,
+    };
+});
+
+// Track if we've already invoked a cookie banner handler on this page load
+let hasInvokedCookieBanner = false;
```
SamyPesse: Why do we need this global variable?

packages/gitbook/src/components/Integrations/LoadIntegrations.tsx
```diff
@@ -65,6 +88,29 @@ if (typeof window !== 'undefined') {
...
             };
         },
+        registerCookieBanner: (handler) => {
+            customCookieBannerStore.setState((state) => ({
+                ...state,
+                hasCustomCookieBanner: true,
+            }));
+
+            // Only invoke the handler once per page load and if the cookie preference hasn't been set yet
+            if (hasInvokedCookieBanner || isCookiesTrackingDisabled() !== undefined) {
```
SamyPesse: I don't think we can do this. The cookie banner has to be loaded anyway, as for example with Onetrust the user can change their mind and edit their settings.

packages/gitbook/src/components/Integrations/LoadIntegrations.tsx
```diff
@@ -76,19 +122,40 @@ export function useIntegrationAssistants(): Array<Assistant> {
...
  */
-export function LoadIntegrations() {
+export function LoadIntegrations(props: { scripts?: Array<{ script: string }> }) {
+    const { scripts = [] } = props;
+    const hasScripts = scripts.length > 0;
     React.useEffect(() => {
-        dispatchGitBookIntegrationEvent('load');
-    }, []);
+        // Only dispatch 'load' event when there are scripts to load
+        if (hasScripts) {
```
SamyPesse: I think we can probably do it in all cases anyway no?

the check on scripts length has not much impact

nolannbiron:
@SamyPesse for the footer button, I guess we could do something like this : 

```javascript
window.GitBook.registerFooterItem({
    label: 'Cookie Preferences',
    onClick: () => {
        OneTrust.ToggleInfoDisplay();
    }
});
```

the only issue I see is what happens if the footer is disabled, and where should we put the button? Under the copyright section ? Inside a dedicated “Policies” group ? I’m not sure what the best approach is

packages/gitbook/src/components/Integrations/LoadIntegrations.tsx
nolannbiron: It was a check to make sure we’re not invoking the banner multiple times if registerCookieBanner is called more than once, but I’ll remove it for now and test to see whether it actually causes any issues

packages/browser-types/src/index.ts
```diff
@@ -46,6 +46,11 @@ export type GitBookAssistant = {
     ui?: boolean;
 };
 
+export type GitBookCookieBannerHandler = (options: {
+    onApprove: () => void;
+    onReject: () => void;
+}) => void;
```
SamyPesse, line 52: This type is inconsistent with what you use in the e2e, where it looks like it you don't pass these callbacks.

Should it be always required or optional?

packages/browser-types/src/index.ts
```diff
@@ -72,6 +77,18 @@ export type GitBookGlobal = {
...
+     * Register a custom cookie banner handler.
+     * When registered, the built-in cookie banner will not be displayed.
+     */
+    registerCookieBanner: (handler: GitBookCookieBannerHandler) => void;
+
+    /**
+     * Whether the user has granted cookie consent.
+     * Returns true if cookies are granted, false if rejected or unknown.
+     */
+    hasGrantedCookies: () => boolean;
```
SamyPesse: `hasApprovedCookies` maybe? to match the `onApprove` above

SamyPesse (approved):

packages/browser-types/src/index.ts
nolannbiron: the callbacks are required in the handler’s type definition, so the handler must accept them when you define it. Integrations can still decide whether to use them or not in their implementation (like in e2e, we call the handler but don't need to use the callbacks)

PR #3927, opened 2026-01-21
Bust data cache

conico974:
There was a change in the API that add a required property, that is indeed provided to us by the API, but we do not bust the data cache automatically.

It can cause stale data to be served while our code expect this property to be there.

This is a workaround until we figure out a better solution for this

### DIFF
--- a/.changeset/four-clowns-begin.md
+++ b/.changeset/four-clowns-begin.md
@@ -1,0 +1,5 @@
+---
+"@gitbook/react-openapi": patch
+---
+
+Fix OpenAPI example not showing
--- a/packages/react-openapi/src/OpenAPICodeSample.tsx
+++ b/packages/react-openapi/src/OpenAPICodeSample.tsx
@@ -10,6 +10,7 @@ import { type OpenAPIContext, getOpenAPIClientContext } from './context';
 import { generateMediaTypeExamples, generateSchemaExample } from './generateSchemaExample';
 import { stringifyOpenAPI } from './stringifyOpenAPI';
 import type { OpenAPIOperationData } from './types';
+import { mergeHeaders } from './util/headers';
 import { getDefaultServerURL } from './util/server';
 import {
     resolvePrefillCodePlaceholderFromSecurityScheme,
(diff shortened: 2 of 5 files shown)


packages/gitbook/src/lib/data/api.ts
```diff
@@ -13,6 +13,14 @@ import { cache } from '../cache';
 import { DataFetcherError, wrapCacheDataFetcherError } from './errors';
 import type { GitBookDataFetcher } from './types';
 
+// We should bump this version when we do breaking changes to the API client.
+// What is considered a breaking change is subjective, but generally:
+// Adding a required property to an existing object, changing the type of an existing property, removing an existing property.
+// We have to do it because otherwise Next.js cache might return cached data that is incompatible with the new code expectations.
+// Especially because typescript types may lead us to think that we are getting the right data when in fact the cached data is old.
+// Right now it's only passed to getPublishedContentSite, but we should use expand its usage to wherever we cache API data and breaking changes happen.
+const API_VERSION = '0.158';
```
gregberge: We should use semver for our API and make it automatic?

conico974: I don't think that's the solution either, we need to find a better way to do that. A semver doesn't tell us what has changed really, and we don't want to bust the entire cache if only 1 endpoint has changed.
I need to think of a right solution, potentially something exported from the api client that gives us a "version" for each endpoint.
This would mean that we'd need to be a lot more careful in our updates to it

packages/gitbook/src/lib/data/api.ts
```diff
@@ -13,6 +13,14 @@ import { cache } from '../cache';
 import { DataFetcherError, wrapCacheDataFetcherError } from './errors';
 import type { GitBookDataFetcher } from './types';
 
+// We should bump this version when we do breaking changes to the API client.
```
SamyPesse: Why should it be global and not per data cache function?

conico974: Yeah it could be, right now it's only used in one function really. 
It's still a bit unclear for me how we could do that in a good way (ideally automatically)

gregberge: We can do a static analysis from type on each return of the cache. Then diff it. If something's break (a new required prop is used), then it breaks in the CI.

conico974: Optional property would break it. I don't think there is a fully automated solution, we just need to figure a good enough one, where we can flag it only once in GBX for example.

PR #3929, opened 2026-01-21
Limit MAX_DPR to 3 in Image component

nolannbiron:
Reduces the maximum device pixel ratio (DPR) from 4 to 3 in the Image component to align with the configuration in routes/image.ts.


gregberge:
Close in favor of #3931

PR #3937, opened 2026-01-25
Improve OpenAPI server URL validation

nolannbiron:


packages/react-openapi/src/util/server.ts
```diff
@@ -45,3 +45,30 @@ function parseServerURL(url: string) {
...
+    return servers.some((server) => {
+        const url = interpolateServerURL(server);
+        return isValidServerURL(url);
+    });
+}
+
+/**
+ * Check if the server URL is valid for making direct HTTP requests.
+ */
+export function isValidServerURL(url: string): boolean {
```
gregberge: Why not using "new URL"?

nolannbiron: Because it can be something like `api.gitbook.com` without a scheme, `new URL` would throw, but in our case it’s still valid, so we can directly check for a domain or subdomain

gregberge: Then the func has a bad naming IMO. It's not an URL.

nolannbiron: Renamed it to `isValidServerHost` 👍🏼

PR #3938, opened 2026-01-25
Add OpenAPI 3.1 nullable array support

nolannbiron:
Fix for this discussion: https://github.com/orgs/GitbookIO/discussions/1138

### DIFF
--- a/packages/react-openapi/src/OpenAPISchema.tsx
+++ b/packages/react-openapi/src/OpenAPISchema.tsx
@@ -664,8 +664,7 @@ function getSchemaProperties(
                 items.type === 'number' ||
                 items.type === 'boolean' ||
                 items.type === 'integer') &&
-            !items.enum &&
-            !Array.isArray(items.type)
+            !items.enum
         ) {
             return null;
         }


packages/react-openapi/src/OpenAPISchema.tsx
```diff
@@ -657,7 +664,8 @@ function getSchemaProperties(
                 items.type === 'number' ||
                 items.type === 'boolean' ||
                 items.type === 'integer') &&
-            !items.enum
+            !items.enum &&
+            !Array.isArray(items.type)
```
gregberge: How can it be array given the check above?

packages/react-openapi/src/OpenAPISchemaName.tsx
```diff
@@ -82,8 +82,11 @@ function getAdditionalItems(schema: OpenAPIV3.SchemaObject, context: OpenAPIClie
         additionalItems += ` · ${tString(context.translation, 'max').toLowerCase()}: ${schema.maximum || schema.maxLength || schema.maxItems}`;
     }
 
-    if (schema.nullable) {
-        additionalItems = ` | ${tString(context.translation, 'nullable').toLowerCase()}`;
+    // Check for nullable in both OpenAPI 3.0 (nullable: true) and OpenAPI 3.1 (type: ['null', ...])
+    const isNullable =
```
gregberge: Shouldn't we use the function `getEffectiveArrayType` here?

packages/react-openapi/src/OpenAPISchema.tsx
nolannbiron: We are checking if schema.items.type is an array, and the check above is checking if schema.type is an array

gregberge: Not it checks `items.type`

nolannbiron: yeah because `const items = schema.items`, and `arrayInfo.isArray` is `Array.isArray(schema.type)`

gregberge: If this is true:

```
(items.type === 'string' ||
                items.type === 'number' ||
                items.type === 'boolean' ||
                items.type === 'integer')
```

Then `items.type` can't be an array.

nolannbiron: ah my bad I thought you were talking about the if above, removed it

PR #3944, opened 2026-01-27
Avoid keyboard-shortcut flicker by adding OS class and using CSS

gregberge:
### Motivation
- Prevent the keyboard shortcut (mod key) from flashing when navigating between sections by avoiding runtime React state for OS detection and using the same static-class approach as theme (dark/light).

### Description
- Add `OperatingSystemClassScript` to inject an inline script that sets an `os-mac` class on the `<html>` element as early as possible (`packages/gitbook/src/components/RootLayout/OperatingSystemClassScript.tsx`).
- Wire the script into the root layout head by rendering `<OperatingSystemClassScript />` from `CustomizationRootLayout` so the class is applied before hydration (`packages/gitbook/src/components/RootLayout/CustomizationRootLayout.tsx`).
- Replace client-side OS state in `KeyboardShortcut` with CSS-driven rendering by always outputting both mac and windows labels and toggling visibility/size via `[html.os-mac_&]` utility classes, removing the `getOperatingSystem`/`useEffect` logic (`packages/gitbook/src/components/primitives/KeyboardShortcut.tsx`).

### Testing
- No automated tests were run for this change.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_b_6978ce6fbb408324a757d2575052f5f2)

### DIFF
--- a/packages/gitbook/src/components/RootLayout/OperatingSystemClassScript.tsx
+++ b/packages/gitbook/src/components/RootLayout/OperatingSystemClassScript.tsx
@@ -2,6 +2,7 @@
 
 import * as React from 'react';
 
+// This function has to be pure because we do a `.toString()` on it.
 function setOperatingSystemClass() {
     if (typeof navigator === 'undefined') return;
 


packages/gitbook/src/components/RootLayout/OperatingSystemClassScript.tsx
```diff
@@ -0,0 +1,38 @@
...
+export function OperatingSystemClassScript() {
+    React.useEffect(() => {
+        setOperatingSystemClass();
+    }, []);
+
+    return (
+        <script
+            suppressHydrationWarning
+            dangerouslySetInnerHTML={{
+                __html: `(${setOperatingSystemClass.toString()})()`,
```
SamyPesse, line 34: I think we should inline the code here instead of using `toString()`

packages/gitbook/src/components/RootLayout/OperatingSystemClassScript.tsx
```diff
@@ -0,0 +1,38 @@
...
+}
+
+/**
+ * Inject a script to detect the OS and apply a class to the <html> element as early as possible.
+ * Bypasses react state to prevent flickering.
+ */
+export function OperatingSystemClassScript() {
+    React.useEffect(() => {
+        setOperatingSystemClass();
+    }, []);
```
SamyPesse, line 28: We should remove this

packages/gitbook/src/components/RootLayout/OperatingSystemClassScript.tsx
gregberge: Why?

packages/gitbook/src/components/RootLayout/OperatingSystemClassScript.tsx
gregberge: No we can't, React re-render the HTML tag and removes the class.

packages/gitbook/src/components/RootLayout/OperatingSystemClassScript.tsx
gregberge: I added a comment about it.

PR #3946, opened 2026-01-27
Fix GIF rendering

gregberge:


packages/gitbook/src/routes/image.ts
```diff
@@ -86,6 +86,11 @@ export async function serveResizedImage(
 
     const defaultFormat = getOriginalFormatFromURL(url);
 
+    // If the format is GIF, always return the original image.
+    if (defaultFormat === 'gif') {
+        return NextResponse.redirect(url, 302);
```
conico974: I do think we want to resize it if possible with avif/webp, it should just not fallback to jpeg but rather redirect directly

gregberge: OK done it.

packages/gitbook/src/routes/image.ts
```diff
@@ -89,7 +89,8 @@ export async function serveResizedImage(
     // Cloudflare-specific options are in the cf object.
     const options: CloudflareImageOptions = {
         fit: 'scale-down',
-        format: defaultFormat,
+        // For GIF, we will use webp as default format for resizing.
+        format: defaultFormat === 'gif' ? 'webp' : defaultFormat,
```
SamyPesse, line 93: Can we actually resize gif? even if animated?

gregberge: It seems so, AVIF and WEBP supports animations.

PR #3948, opened 2026-01-27
Fix for the assistant

conico974:


packages/gitbook/src/components/Search/SearchContainer.tsx
```diff
@@ -80,11 +81,13 @@ export function SearchContainer({
         if (assistants.length === 0) return;
         if (state?.ask === undefined || state?.ask === null) return;
 
-        initialRef.current = true;
-
         // For simplicity we're only triggering the first assistant
-        assistants[0]?.open(state.ask ?? undefined);
-    }, [state?.ask, assistants.length, assistants[0]?.open]);
+        // Because this is in the layout, we need to await for the body to be loaded.
+        if (isLoaded) {
```
gregberge, line 86: Why not following the same early return pattern?

gregberge (approved):

packages/gitbook/src/components/Search/SearchContainer.tsx
conico974: I find it clearer this way, and functionally it's the same so...

PR #3985, opened 2026-02-10
Fix unexpected error thrown when using the embed in standalone mode

gregberge:
The `GitBook(configure, {})` was potentially called before initialization


packages/embed/src/standalone/index.ts
```diff
@@ -193,5 +193,3 @@ const precalls = (window.GitBook as GitBookStandalone | undefined)?.q ?? [];
 // @ts-expect-error - GitBook is not defined in the global scope
 window.GitBook = GitBook;
 precalls.forEach((call) => GitBook(...call));
-
-GitBook('configure', {});
```
viktorrenkema, line 197: We don't need this anymore? Got a bit more context why we had this and why it's safe to remove?

gregberge: I am not sure why we had it because I checked in the doc and we say to the user to add it. Since I didn't write the code I don't have the context of why it is present.

PR #3987, opened 2026-02-10
Fix links handling in embedded Assistant

gregberge:
Several things here:

- If the docs tab is omitted, all links are considered as external
- Fix external links pointing to the agent by switching the href
- Prepare the ground to use the proxy config to detect external links (still using the origin)

### DIFF
--- a/packages/gitbook/src/components/primitives/Link.tsx
+++ b/packages/gitbook/src/components/primitives/Link.tsx
@@ -4,7 +4,6 @@ import NextLink, { type LinkProps as NextLinkProps } from 'next/link';
 import React from 'react';
 
 import { tcls } from '@/lib/tailwind';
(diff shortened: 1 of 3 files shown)


packages/gitbook/src/components/primitives/Link.tsx
```diff
@@ -34,13 +33,18 @@ export type LinkProps = Omit<BaseLinkProps, 'href'> &
         classNames?: DesignTokenName[];
     };
 
+type LinkTarget = '_blank' | '_self';
+
+export type LinkContextType = {
+    externalTarget: LinkTarget;
```
viktorrenkema, line 39: nitpicky, but `_self` is not really external right, so the name of this prop isn't super clear.

gregberge: It's the target for an external link, so I think it is.

PR #3999, opened 2026-02-12
Hide cookie banner for AI user agents

nolannbiron:


packages/gitbook/src/lib/browser/user-agent.ts
```diff
@@ -0,0 +1,69 @@
...
+ * Returns true if the current User-Agent appears to be an AI crawler or AI-assisted browser.
+ * Used to hide cookie banners and avoid showing consent UI to non-human viewers.
+ */
+export function isAIUserAgent(): boolean {
+    if (typeof navigator === 'undefined' || !navigator.userAgent) {
+        return false;
+    }
+    const ua = navigator.userAgent.toLowerCase();
+    return AI_USER_AGENT_PATTERNS.some((pattern) => ua.includes(pattern.toLowerCase()));
+}
```
SamyPesse, line 69: This should be a hook with useMemo

You are currently calling it from a lot of components that will re-render and will need to do a find on a list of 50 items

SamyPesse: or just cache it globally as the user-agent will not change in the browser

PR #4006, opened 2026-02-13
Add support for rendering Tags

viktorrenkema:
# Summary
Adds tag rendering for the upcoming tags release, this PR supports:
- Changelogs: render tags below the date in the sticky sidebar of update blocks
- Page tags: display all page tags inline next to the title, with individual tag wrapping for responsive layouts
- Table of contents tag: show the primary tag next to page titles in the sidebar

Todo next:
- colors: it uses a neutral styles for now, this will change soon tho as we add support for colored tags.
- better truncating in TOC
- i already changed `tags` on `revision` being required so there's a couple inline TODO's noted which ill improve later

CleanShot 2026-02-19 at 10 50 22@2x

CleanShot 2026-02-19 at 10 50 41@2x


packages/gitbook/src/components/TableOfContents/encodeClientTableOfContents.ts
```diff
@@ -122,3 +127,24 @@ export async function encodeClientTableOfContents(
 
     return result;
 }
+
+/**
+ * Resolve the primary tag for a page if it exists.
+ */
+function resolvePrimaryTag(
```
zenoachtig, line 135: Seems like this should be in /lib/tags if it's not completely specific to the TOC, which I think it isn't

packages/gitbook/src/components/PageBody/PageHeader.tsx
```diff
@@ -128,16 +129,19 @@ export async function PageHeader(props: {
...
+                        'flex-wrap',
                         'items-center',
-                        'gap-[.5em]',
+                        'gap-y-2',
                         'grow',
                         'text-pretty',
                         'clear-right',
-                        'xs:clear-none'
+                        'xs:clear-none',
+                        '[&>[data-tag]:first-of-type]:ml-4' // we replaced the gap with margins, as we dont want tags rendered in 1 container since the UX of having multiple tags wrap one by one is much better than having to wrap them all at once at a much larger viewport.
```
zenoachtig: I don't think tags should be shown after the page title (and have these margin overrides as a result). In GBX we show them above the page title, I think that makes a lot more sense, wdyt?

packages/gitbook/src/components/TableOfContents/encodeClientTableOfContents.ts
```diff
@@ -122,3 +127,24 @@ export async function encodeClientTableOfContents(
...
+        return undefined;
+    }
+
+    const primary = pageTags.find((t) => t.primary);
+    if (!primary) {
+        return undefined;
+    }
+
+    const revisionTags = getRevisionTags(context.revision);
+    return resolveTag(primary.tag.tag, revisionTags);
```
zenoachtig, line 150: Nit: `primary.tag.tag` is nested kinda weird, but feel free to disregard if that's simply the data structure

packages/gitbook/src/components/TableOfContents/encodeClientTableOfContents.ts
```diff
@@ -122,3 +127,24 @@ export async function encodeClientTableOfContents(
 
     return result;
 }
+
+/**
+ * Resolve the primary tag for a page if it exists.
```
zenoachtig: I don't understand the concept of primary tags exactly, it's returned by the API right? Perhaps add its intention in a comment here, or how we display

packages/gitbook/src/components/DocumentView/Update.tsx
```diff
@@ -35,13 +37,12 @@ export function Update(props: BlockProps<DocumentBlockUpdate>) {
         short: formatDateShort(parsedDate),
     }[dateFormat];
 
+    // Resolve tags from the block data using revision-level tag definitions
+    const revisionTags = getRevisionTags(contextProps.context.contentContext?.revision);
+    const resolvedTags = resolveBlockTags(block.data.tags, revisionTags);
```
zenoachtig, line 42: What's the difference between revision tags and block tags? Are block tags not saved to the page-level tags?

packages/gitbook/src/components/TableOfContents/encodeClientTableOfContents.ts
viktorrenkema: It is specific to there. A primary tag is the sole tag that we display in the TOC. It has no other use in gbo except for the TOC, nowhere else do we care about the primary tag.

packages/gitbook/src/components/TableOfContents/encodeClientTableOfContents.ts
viktorrenkema: I get it, it can use a bit more context. Basically we store tags as reusable entities on the space. Then in the content we can specify which of the available tags on the space we want to attach to this content entity. 

Right now we have 2 types of content that can be tagged:
- tags on pages
- tags on update blocks

For tags on pages, we _also_ want to allow displaying one of these tags in the TOC. Not all, as there is no space, so for now we limit to 1, and this tag needs to be identified which we do through a `primary` prop.

Because `primary` tags have no utility for update block tags, they're really only something for page tags, the concept of primary should be living on the page tag definition, not on the global tag definition.

And to answer your q, yeah they are returned by the API. 'Resolve' in this context means 'check and retrieve any primary tag if it exists'. I updated the comment to better summarize what the primary tag is.

For extra clarity, see PR description [here](https://github.com/GitbookIO/gitbook-x/pull/21293).

packages/gitbook/src/components/DocumentView/Update.tsx
viktorrenkema: No, the explanation i gave here: https://github.com/GitbookIO/gitbook/pull/4006#discussion_r2826316475 also had this question in mind. Tags are a type of content ref that can be pulled into various places, right now 1. pages or 2. update blocks. 

So here we are basically getting the available list of all tags available on the space, and then checking which of those we need on the update block. This is because the list of tags on the update block isnt complete: it's just a list of the id's (we use the slug as id), but we dont yet have the additional properties like label (and soon icon, color etc).

packages/gitbook/src/components/TableOfContents/encodeClientTableOfContents.ts
viktorrenkema: Yeah when I was building the API in gbx i agreed but I checked against other content refs, and we always have it like this and i didnt want to break a pattern.

packages/gitbook/src/components/PageBody/PageHeader.tsx
viktorrenkema: [Discussion in Slack](https://gitbook.slack.com/archives/C0ADBL2S7K7/p1771489945984359), fixed now.

zenoachtig (approved):

PR #4012, opened 2026-02-16
Fix external links logic for proxy

gregberge:


packages/gitbook/src/components/utils/link.ts
```diff
@@ -22,7 +22,23 @@ export function isExternalLink(href: string, origin: string | null = null) {
...
+    }
+
+    // Compare pathname exactly
+    if (parsed.pathname === originUrl.pathname) {
+        return false;
+    }
+
+    // Then compare the pathname by adding "/" to ensure we don't match "gitbook.com/docs-x"
+    return !parsed.pathname.startsWith(
+        originUrl.pathname.endsWith('/') ? originUrl.pathname : `${originUrl.pathname}/`
```
taranvohra: I don't understand this and it looks like a hack.

gregberge: No it's not check the tests, it's to handle gitbook.com/docs-thing

taranvohra: Ah okay I read it incorrectly. But you should use `withTrailingSlash(originUrl.pathname)`, it's cleaner and less mental overhead

PR #4013, opened 2026-02-16
Rework page layout

zenoachtig:
- **Standardized layout state via Tailwind variants**: the branch introduces **two** layout mode variants (`layout-default`, `layout-wide`) and uses `body:has(.layout-*)` selectors to drive responsive structure consistently.
- **Added `CONTENT_STYLE` to centralize content width rules**: separates “site chrome” sizing (`CONTAINER_STYLE`) from the actual page content max-width, so multiple components share the same width logic.
- **Replaced per-component width overrides with layout-mode-driven constraints**: `PageBody` now determines `wideLayout` once (page config + full-width blocks) and applies `CONTENT_STYLE` + `layout-*` classes, reducing ad-hoc `max-w-*` / special-case overrides.
- **Reworked both sidebars around the new layout modes**: left TOC/sidebar gets `no-sidebar` / `has-sidebar` behavior with **minimal/fixed variants in wide layout**, and the right outline is refactored into a proper `SideSheet` with overlay/close + explicit “outline open” state.

Deliberate effects of this rework:
- **When one block is wide, the page is wide.** The page layout treats a single wide block the same as the `wide` page-layout option. No more individual blocks to adjust. While this is technically a reduction of block options, in practice this results in more uniform layouts and the system functioning as people expect.
- **When the page is wide, all blocks that _can be wide_ are wide as well.**
- **Moving "On this page" into a side sheet** will result in the page outline being available on mobile screens as well.


zenoachtig:
@codex review

packages/gitbook/e2e/internal.spec.ts
```diff
@@ -781,7 +781,9 @@ const testCases: TestsCase[] = [
                 run: async (page) => {
                     await expect(page.locator('[data-testid="table-of-contents"]')).toBeVisible();
                     // Trademark exists by default
-                    await expect(page.getByTestId('gb-trademark')).toHaveCount(1);
+                    expect(await page.getByTestId('gb-trademark').count()).toBeGreaterThanOrEqual(
+                        1
+                    );
```
viktorrenkema: was this needed/better for some reason?

packages/gitbook/src/components/PageAside/PageAside.tsx
```diff
@@ -30,46 +31,20 @@ export function PageAside(props: {
...
     const { customization, site } = context;
+    const language = getSpaceLanguage(context);
 
     return (
-        <aside
+        <SideSheet
+            side="right"
+            toggleClass="outline-open"
+            withOverlay={true}
+            withCloseButton={true}
```
viktorrenkema:
```suggestion
            withOverlay
            withCloseButton
```

packages/gitbook/src/components/PageAside/PageAside.tsx
```diff
@@ -87,80 +62,68 @@ export function PageAside(props: {
...
-                        {document ? (
-                            <React.Suspense fallback={null}>
-                                <PageAsideSections document={document} context={context} />
-                            </React.Suspense>
-                        ) : null}
-                        <PageAsideActions page={page} withPageFeedback={withPageFeedback} />
-                    </div>
+                    <>
+                        <div className="mb-3 ml-3 flex layout-wide:3xl:hidden page-no-outline:hidden items-center justify-between max-lg:hidden layout-default:xl:hidden page-api-block:2xl:hidden">
+                            <h6 className="flex items-center gap-1 font-semibold text-tint text-xs uppercase leading-wider">
```
viktorrenkema, line 106: Dont we have Heading components in gbo?

packages/gitbook/src/components/PageAside/PageAsideButton.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+export function PageAsideToggleButton() {
+    const language = useLanguage();
+
+    const pathname = usePathname();
+
+    // Close the navigation when navigating to a page
+    useEffect(() => {
+        document.body.classList.remove(globalClassName);
+    }, [pathname]);
```
viktorrenkema, line 18: Is this how we've already been doing this? Feels hacky somehow?

packages/gitbook/src/components/SpaceLayout/SpaceLayout.tsx
```diff
@@ -128,23 +128,28 @@ export function SpaceLayout(props: SpaceLayoutProps) {
...
                 <div
                     className={tcls(
                         'flex',
                         'flex-col',
                         'lg:flex-row',
                         'lg:justify-center',
                         CONTAINER_STYLE,
-                        'site-width-wide:max-w-screen-4xl',
-                        'transition-[max-width] duration-300 motion-reduce:transition-none',
+                        'transition-[max-width] duration-300',
```
viktorrenkema, line 140: i know this isnt new code, but `duration-300` seems long for a chat panel?

packages/gitbook/tests/rss.test.ts
```diff
@@ -11,7 +11,7 @@ it('should expose a RSS feed for a page with updates', async () => {
 
     expect(feed.title).toBe('Updates | E2E Tests GitBook Open');
     expect(feed.items.length).toBe(4);
-});
+}, 10000);
```
viktorrenkema: Added on purpose?

viktorrenkema (approved):
lgtm!

packages/gitbook/e2e/internal.spec.ts
zenoachtig: Needed — we output 2 trademark buttons that are quite different (one is iconOnly, the other isn't), and can only determine which one is appropriate based on screen size / CSS. So we will _always_ have both and render either one conditionally. I'll add a note

packages/gitbook/src/components/PageAside/PageAside.tsx
zenoachtig: Nope, and it's a bit of a one-off for now, but I'll keep it in mind

packages/gitbook/src/components/SpaceLayout/SpaceLayout.tsx
zenoachtig: You'd be surprised haha, it's pretty snappy because it's a big element

packages/gitbook/tests/rss.test.ts
zenoachtig: It wasn't resolving before, I can double-check it works now

packages/gitbook/src/components/PageAside/PageAsideButton.tsx
zenoachtig: Yep, that's how the mobile side nav has always been doing it 😅

PR #4015, opened 2026-02-16
Integrate sites OAuth server flow in GBO for protected resources like MCP

spastorelli:


packages/gitbook/src/middleware.ts
```diff
@@ -160,11 +168,25 @@ async function serveSiteRoutes(requestURL: URL, request: NextRequest) {
...
     }
 
+    // Handler that forwards redirections from upstream auth provider during a site's OAuth /authorize session
+    // back to the site's OAuth server.
+    const siteOAuthAuthorizeMatch = new URLPattern({
+        pathname: '*/~gitbook/oauth2/v1/:siteId/authorize',
+    }).exec(siteRequestURL.toString());
+    if (siteOAuthAuthorizeMatch) {
+        const siteOAuthAuthorizeURL = new URL(
+            `/oauth2/v1/${siteOAuthAuthorizeMatch.pathname.groups.siteId}/authorize?${siteOAuthAuthorizeMatch.search.input}`,
```
taranvohra: We seem to be hardcoding `/oauth2/v1` in a bunch of places. Can't we use the env value here?

spastorelli: Yes good idea, I'll make the change. Thanks.

.github/composite/deploy-cloudflare/action.yaml
```diff
@@ -52,6 +52,7 @@ runs:
             GITBOOK_API_URL: ${{ inputs.opItem }}/GITBOOK_API_URL
             GITBOOK_API_PUBLIC_URL: ${{ inputs.opItem }}/GITBOOK_API_PUBLIC_URL
             GITBOOK_API_TOKEN: ${{ inputs.opItem }}/GITBOOK_API_TOKEN
+            GITBOOK_SITES_OAUTH_SERVER_URL: ${{ inputs.opItem }}/GITBOOK_SITES_OAUTH_SERVER_URL
```
emmerich: Do we expect to run multiple oauth servers? Otherwise a clearer name might be:

```suggestion
            GITBOOK_OAUTH_SERVER_URL: ${{ inputs.opItem }}/GITBOOK_OAUTH_SERVER_URL
```

packages/gitbook/src/lib/visitors.ts
```diff
@@ -397,6 +411,31 @@ function getVisitorAuthTokenFromCookies(
...
+export function getVisitorTokenForOAuthProtectedResource(args: {
+    url: URL | NextRequest['nextUrl'];
+    headers: Headers;
+}) {
+    const { url, headers } = args;
+
+    // Check first if it is included in the headers otherwise fallback to query param.
+    const fromAuthHeader = headers.get('Authorization');
+    if (fromAuthHeader) {
+        const match = fromAuthHeader.match(/^Bearer\s+(.+)$/i);
```
emmerich: I'm not sure how often this is going to be called, but something like:


```
const parts = fromAuthHeader.split(' ')
if (parts[0] === 'Bearer' && parts[1]?.trim()) {
```

will be slightly faster than a regex

spastorelli: Yes, good idea. It could be called often, I'll make the change.

PR #4025, opened 2026-02-18
Fix issue where customizations stored in cookie is shared between different site previews

spastorelli:
This PR aims to fix an issue where customizations stored in cookie is shared between different site previews. This caused an issue in the app where a customization change in a site settings was applied to all sites preview in the All site screen.

### DIFF
--- a/packages/gitbook/src/middleware.ts
+++ b/packages/gitbook/src/middleware.ts
@@ -28,6 +28,7 @@ import {
 import { serveResizedImage } from '@/routes/image';
 import { cookies } from 'next/headers';
 import type { SiteURLData } from './lib/context';
+import { getPreviewRequestIdentifier } from './lib/preview';
 import { serveProxyAnalyticsEvent } from './lib/tracking';
 export const config = {
     matcher: [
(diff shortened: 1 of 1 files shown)


packages/gitbook/src/middleware.ts
```diff
@@ -327,7 +328,9 @@ async function serveSiteRoutes(requestURL: URL, request: NextRequest) {
                     httpOnly: true,
                     sameSite: 'lax',
                     maxAge: 10 * 60, // 10 minutes
-                    path: '/url/preview', // Only send the cookie to preview routes
```
emmerich, line 330: I think there's another use of this? Line 346 - should we apply to both?

spastorelli: I thought of doing it for theme initially too but thought that we may want to have the selected theme applied across all previews so left it as it is.

spastorelli:
> Approved. I don't know the `/url/preview/...` URL by heart but I'm guessing you already tested that the cookie still matches, even with the new (more specific) path.

Yes preview URLs are of the shape  `https://open-2v.gitbook.com/url/preview/site_bvY46/` so path should be `url/preview/site_bvY46/`

PR #4026, opened 2026-02-19
Allow to interpolate `{location.href}` and `{location.pathname}` in external links

gregberge:
If a link contains `{location.href}` or `{location.pathname}`, we automatically interpolate these values. This allows users to create redirect flows from an external app. These placeholders work in any link.


gregberge:
Discussed in Triage, we don't want to go that way because the feature is too large for the use case.

PR #4027, opened 2026-02-19
Integrate Ask AI with upcoming insights by passing visitor session IDs

SamyPesse:


packages/gitbook/src/components/Insights/visitorId.tsx
```diff
@@ -43,15 +53,15 @@ export function VisitorSessionProvider(
...
 
-        const pendingSession = fetchSession({ appURL, visitorCookieTrackingEnabled });
-        visitorSessionStore.setState({ pendingSession, session: null });
-        pendingSession.then((session) => {
-            visitorSessionStore.setState({ pendingSession: null, session });
+        const pendingVisitor = fetchGlobalVisitor({ appURL, visitorCookieTrackingEnabled });
+        visitorStore.setState({ pendingVisitor, visitor: null });
+        pendingVisitor.then((visitor) => {
+            visitorStore.setState({ pendingVisitor: null, visitor });
         });
```
gregberge, line 65:
```suggestion
        let canceled = false;
        const pendingVisitor = fetchGlobalVisitor({ appURL, visitorCookieTrackingEnabled });
        visitorStore.setState({ pendingVisitor, visitor: null });
        pendingVisitor.then((visitor) => {
            if (!canceled) {
              visitorStore.setState({ pendingVisitor: null, visitor });
            }
});

return () => {
  canceled = true;
}
```

You can also check if `pendingVisitor` is the same but I don't know if you have a `getState` method to access the current state, or a `previous => ...` signature.

SamyPesse: It cannot happen because the promise will be directly set as `pendingVisitor` so the next call will return  before

PR #4030, opened 2026-02-20
Handle non-array required field in OpenAPI schema

nolannbiron:
## Summary
- Fix `TypeError: ((intermediate value) || []) is not iterable` when rendering OpenAPI specs that use `"required": true` (boolean) on individual properties instead of the standard `string[]` format
- Guard `mergeRequiredFields` with `Array.isArray()` checks so non-array values are safely ignored

### DIFF
--- a/packages/react-openapi/src/OpenAPISchema.tsx
+++ b/packages/react-openapi/src/OpenAPISchema.tsx
@@ -1108,17 +1108,15 @@ function mergeRequiredFields(
     const ancestorRequired = Array.isArray(latestAncestor?.required)
         ? latestAncestor.required
         : undefined;
-    const schemaRequired =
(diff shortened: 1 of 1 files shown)


packages/react-openapi/src/OpenAPISchema.tsx
```diff
@@ -1105,15 +1105,21 @@ function mergeRequiredFields(
...
+            : undefined;
+
+    if (!ancestorRequired && !schemaRequired) {
         return undefined;
     }
 
     if (checkIsReference(schemaOrRef)) {
-        return latestAncestor?.required;
+        return ancestorRequired;
     }
```
gregberge, line 1114: Should be moved before even trying to check `schemaRequired`.

PR #2626, opened 2024-12-12
Send redirectOnError param to getPublishedContent when token is pulled from cookie

spastorelli:
This PR aims to update GBO after changes made in https://github.com/GitbookIO/gitbook-x/pull/16138

PR #4031, opened 2026-02-20
Upgrade Argos & Playwright

gregberge:


packages/gitbook/e2e/internal.spec.ts
```diff
@@ -1640,33 +1636,36 @@ const testCases: TestsCase[] = [
...
+            //                 name: getVisitorAuthCookieName(basePath),
+            //                 value: getVisitorAuthCookieValue(basePath, invalidToken),
+            //                 httpOnly: true,
+            //             },
+            //         ];
+            //     })(),
+            //     run: async (page) => {
+            //         await expect(page).toHaveURL(/https:\/\/www.google.com/);
+            //     },
+            // },
```
viktorrenkema, line 1668: Doesn't have to block the upgrading in this PR, but we should make note to either try to make this test work or double check that it can be removed and covered in a different way. Maybe good to check with @spastorelli as you added it originally here https://github.com/GitbookIO/gitbook/pull/2626

gregberge: Unfortunately it does not work with Playwright, it seems to be a very nasty bug here. I tried different approach and it just doesn't work. Yeah for sure we will not block the upgrade for that.

@spastorelli if the redirect is done server-side, it can be done in Node.js, if done client-side, I am afraid we have to wait for Playwright to fix the bug at some point to re-enable it.

gregberge: I will merge it and @spastorelli can figure out when he will come back.

PR #4045, opened 2026-02-24
Include headings inside columns and tabs in published page outline

john-gitbook:
### Motivation
- The published site extractor missed headings placed inside `columns` and `tabs` blocks, causing a mismatch with the editor outline and broken hash links for column/tab sections.

### Description
- Update `getSectionsFromNodes` in `packages/gitbook/src/lib/document-sections.ts` to handle `columns` and `tabs` blocks by filtering valid child blocks (`column` for `columns`, `tabs-item` for `tabs`) and recursing into each child to extract nested heading sections.
- The new logic preserves the current `depth` handling so headings inside these containers keep the same depth semantics as other headings.
- Add unit tests in `packages/gitbook/src/lib/document-sections.test.ts` that assert headings inside `columns` and `tabs` are returned with expected `id`, `title`, and `depth`.

### Testing
- Ran `bun format` after changes and the formatter completed (repository has pre-existing lint warnings unrelated to this change). 
- Added and ran `bun test packages/gitbook/src/lib/document-sections.test.ts`; the test file includes `mock.module` for OpenAPI helpers, but the test run in this environment failed due to missing monorepo runtime modules (e.g. `@gitbook/cache-tags`) required by other imports, so the unit test could not be executed end-to-end here.
- The test file itself asserts that headings inside `columns` and `tabs` resolve to sections and is included in the PR for CI to run in the full monorepo environment where dependencies are available.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_b_699de00adb4c832aa412a9570e97eb49)

### DIFF
--- a/.changeset/bright-streets-hammer.md
+++ b/.changeset/bright-streets-hammer.md
@@ -1,0 +1,7 @@
+---
+"@gitbook/openapi-parser": patch
+"@gitbook/react-openapi": patch
+"gitbook": patch
+---
+
+Add server-side proxy for Scalar API client to bypass CORS
--- a/.changeset/chatty-webs-reply.md
+++ b/.changeset/chatty-webs-reply.md
@@ -1,0 +1,5 @@
+---
+"gitbook": patch
+---
+
+Show "Ask Assistant" search result immediately
--- a/packages/gitbook/src/app/~scalar/proxy/route.ts
+++ b/packages/gitbook/src/app/~scalar/proxy/route.ts
@@ -1,0 +1,32 @@
+import { handleOpenAPIProxyOptions, handleOpenAPIProxyRequest } from '@/routes/openapi-proxy';
+import type { NextRequest } from 'next/server';
+
+export const dynamic = 'force-dynamic';
+
+export async function GET(request: NextRequest) {
+    return handleOpenAPIProxyRequest(request);
+}
+
+export async function POST(request: NextRequest) {
+    return handleOpenAPIProxyRequest(request);
+}
+
+export async function PUT(request: NextRequest) {
+    return handleOpenAPIProxyRequest(request);
+}
+
+export async function DELETE(request: NextRequest) {
+    return handleOpenAPIProxyRequest(request);
+}
+
+export async function PATCH(request: NextRequest) {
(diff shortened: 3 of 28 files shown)


john-gitbook:
https://6093357a-gitbook-open-v2-preview.gitbook.workers.dev/url/example-21.gitbook.io/check

packages/gitbook/src/lib/document-sections.ts
```diff
@@ -77,6 +77,19 @@ async function getSectionsFromNodes(
                 }
                 continue;
             }
+            case 'columns':
+            case 'tabs': {
```
clairechabas: Headings in tabs would currently render only their text content. It could be confusing for 2 reasons: 
- headings in non-active tabs would not be visible on the page without clicking on the outline link to make them active.
- headings in a tabs item may require the context of the tab items title to be fully clear.

Should we add the tabs item's title as prefix? We'd need to have a way to do that while maintaining the outline item's readability high since it could become very long.

packages/gitbook/src/lib/document-sections.ts
```diff
@@ -77,6 +77,17 @@ async function getSectionsFromNodes(
...
+                const columns = (block.nodes ?? []).filter((child) => child.type === 'column');
+                const nestedSections = await Promise.all(
+                    columns.map((child) => getSectionsFromNodes(child.nodes ?? [], context, depth))
+                );
+
+                for (const childSections of nestedSections) {
+                    sections.push(...childSections);
+                }
+                continue;
+            }
```
clairechabas: Seems like it's the same need as the stepper isn't it?

We could use it to avoid duplicated logic:

```suggestion
case 'stepper': 
case 'columns': {
const childrenNodes = await Promise.all(
block.nodes.map(async (child) =>
getSectionsFromNodes(child.nodes, context, depth)
)
);
for (const sections of childrenNodes) {
sections.push(...sections);
}
continue;
}
```

Or is there a specificity to columns that I'm missing? Wdyt?

gregberge: Yeah AI..

PR #4050, opened 2026-02-25
Add server-side proxy for Scalar API client to bypass CORS

nolannbiron:
Route API requests from Scalar's "Try It" button through a server-side proxy at `/~scalar/proxy`, avoiding browser CORS restrictions.

- Add `x-enable-proxy` spec-level extension to opt-in per OpenAPI spec
- Extract and pass `x-enable-proxy` through the OpenAPI data pipeline
- Configure Scalar's proxyUrl when `x-enable-proxy` is true
- Implement proxy route handler with SSRF protection (private IP blocking, DNS resolution checks, redirect validation)
- Strip upstream CORS headers and apply permissive ones
- Handle `X-Scalar-Cookie` and `X-Scalar-User-Agent` forwarding

Mainly inspired by: https://github.com/scalar/scalar/tree/main/projects/proxy-scalar-com

### DIFF
--- a/packages/gitbook/src/routes/openapi-proxy.test.ts
+++ b/packages/gitbook/src/routes/openapi-proxy.test.ts
@@ -364,7 +364,7 @@ describe('handleOpenAPIProxyRequest', () => {
 
         expect(res.status).toBe(502);
         const body = (await res.json()) as { error: string };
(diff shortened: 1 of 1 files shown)


gregberge (approved):
It looks good to me.

PR #4056, opened 2026-02-26
Add real IP extraction to analytics event proxy

conico974:
Implement real IP extraction from request headers to enhance analytics event tracking. This change improves the accuracy of IP data sent to the analytics service.

### DIFF
--- a/packages/gitbook/src/lib/tracking.ts
+++ b/packages/gitbook/src/lib/tracking.ts
@@ -91,13 +91,19 @@ export async function trackServerInsightsEvents(args: {
         timestamp: event.timestamp ?? new Date().toISOString(),
     })) as api.SiteInsightsEvent[];
 
-    const realIp = getRealIp(request.headers);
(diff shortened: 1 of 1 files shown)


packages/gitbook/src/lib/tracking.ts
```diff
@@ -75,9 +77,36 @@ export async function serveProxyAnalyticsEvent(req: Request) {
...
+        return headers.get('x-forwarded-for');
+    }
+
+    if (headers.has('x-vercel-forwarded-for')) {
+        return headers.get('x-vercel-forwarded-for');
+    }
+
+    if (headers.has('x-vercel-proxied-for')) {
+        return headers.get('x-vercel-proxied-for');
+    }
```
SamyPesse, line 174: This ones are not IPs, but array of IPs; no?

conico974: No this specific one is single IP, x-forwarded-for and the vercel one are just classic one

PR #4058, opened 2026-02-26
Fix cards styles

clairechabas:
Fixes:
- Use primary color for card's ring color on hover (as we do this for links)
- Use `text-base` for cards description text color
- Remove the text color change on link cards hover
- Remove the ring color change on non-link cards on hover (non-interactive elements shouldn't have a hover state)

Closes RND-9805

### DIFF
--- a/packages/gitbook/src/components/hooks/useScrollPage.ts
+++ b/packages/gitbook/src/components/hooks/useScrollPage.ts
@@ -6,16 +6,33 @@ import { useHash } from './useHash';
 import { usePrevious } from './usePrevious';
 
 /**
- * Scroll the page to the hash or reset scroll to the top.
- * Only triggered while navigating in the app, not for initial load.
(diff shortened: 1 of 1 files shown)


packages/gitbook/src/components/DocumentView/Table/styles.ts
```diff
@@ -21,6 +21,6 @@ export const RecordCardStyles = [
     'before:ring-tint-12/2',
     'before:z-10',
     'before:relative',
-
-    'hover:before:ring-tint-12/5',
 ] as ClassValue;
+
+export const RecordCardLinkStyles = ['hover:before:ring-primary'] as ClassValue;
```
zenoachtig:
```suggestion
export const RecordCardLinkStyles = ['hover:before:ring-primary-hover'] as ClassValue;
```
We have a `-hover` border colour that is slightly more pronounced for this use case

PR #4061, opened 2026-02-26
Harden visitor session detection to improve GBO toolbar visibility

gregberge:
### Motivation
- The authenticated GBO toolbar is shown only when `useVisitor()` resolves a visitor with an `organizationId`, and transient failures resolving the visitor can cause the toolbar to not open.
- The previous implementation trusted any `__session` cookie value and used a single short 500ms fetch to revalidate, which can cause false anonymous states on slow or cold networks.

### Description
- Modified `packages/gitbook/src/components/Insights/visitorId.tsx` to add a `isSignedInVisitor` type guard so cookie-only state is trusted only when it contains both `userId` and `organizationId`.
- Stop short-circuiting on anonymous cookie state and revalidate server-side by calling `GET /__session/2/` when needed.
- Replaced the single 500ms fetch with `fetchGlobalVisitorWithRetry` that retries with longer timeouts (`1000ms` then `2500ms`) and validates `resp.ok` to treat non-2xx responses as failures.
- Preserve prior cookie visitor state on fetch failure as a fallback to reduce transient drops in detected signed-in status.

### Testing
- Ran the repository formatter with `bun format`, which completed but reported project-wide diagnostics and fixed two files. (Formatting step succeeded.)
- Ran a targeted static lint/check on the changed file with `bunx biome check packages/gitbook/src/components/Insights/visitorId.tsx`, which completed with no fixes required.
- Attempted package typecheck with `bun run typecheck` in `packages/gitbook`, which failed due to unrelated workspace/module resolution errors in the environment and not due to the change in this file.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_b_69a05e67eb30832492e859ba86e766d6)

### DIFF
--- a/packages/gitbook/src/components/Insights/visitorId.tsx
+++ b/packages/gitbook/src/components/Insights/visitorId.tsx
@@ -168,7 +168,23 @@ function getGlobalVisitor({
         url.searchParams.set('proposed', proposedId);
 
         try {
-            const visitor = await fetchGlobalVisitorWithRetry(url);
+            const resp = await fetch(url, {
+                method: 'GET', // Use GET to play nicely with SameSite cookies.
+                credentials: 'include', // Make sure to send/receive cookies.
+                cache: 'no-cache',
+                mode: 'cors', // Need to use cors as we are on a different domain.
+                signal: AbortSignal.timeout(500),
+            });
+
+            if (!resp.ok) {
+                throw new Error(`Unexpected __session response: ${resp.status}`);
+            }
+
+            const visitor = await resp.json();
+
+            if (!isVisitor(visitor)) {
+                throw new Error(`Unexpected __session format: ${JSON.stringify(visitor)}`);
+            }
 
             // When cookie tracking is disabled we still allow a signed-in session to be detected,
             // but otherwise we preserve the no-cookie behavior by returning an anonymous visitor.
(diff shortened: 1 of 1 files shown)


packages/gitbook/src/components/Insights/visitorId.tsx
```diff
@@ -112,8 +118,9 @@ async function fetchGlobalVisitor({
 
     const { existing, proposedId } = getVisitorFromCookies();
 
-    if (existing) {
-        // If the cookie already exists, we'll just use that. Avoids a server request.
+    if (isSignedInVisitor(existing)) {
+        // If we already have a signed-in visitor in cookies, we can skip the server request.
+        // This is the common case when users navigate between pages.
```
SamyPesse: This is a big change that will result in a LOT of requests at scale, as it means every new page view will do a tracking request now.

packages/gitbook/src/components/Insights/visitorId.tsx
```diff
@@ -93,51 +144,93 @@ export function getVisitor(): MaybePromise<VisitorResponse> {
...
-        const visitor = (await resp.json()) as VisitorResponse;
-        return visitor;
-    } catch (error) {
-        console.error('Failed to fetch visitor session ID', error);
-        return {
-            deviceId: proposedId,
-        };
+async function fetchGlobalVisitorWithRetry(url: URL): Promise<VisitorResponse> {
+    let lastError: unknown;
+    for (const timeoutMs of [1000, 2500]) {
```
SamyPesse: I don't think we should retry, otherwise this will block the tracking of events.

If it fails, we should just use the proposed one.

This is why we need a short timeout (500ms)

packages/gitbook/src/components/Insights/visitorId.tsx
```diff
@@ -93,51 +144,93 @@ export function getVisitor(): MaybePromise<VisitorResponse> {
...
+        try {
+            const resp = await fetch(url, {
+                method: 'GET', // Use GET to play nicely with SameSite cookies.
+                credentials: 'include', // Make sure to send/receive cookies.
+                cache: 'no-cache',
+                mode: 'cors', // Need to use cors as we are on a different domain.
+                signal: AbortSignal.timeout(timeoutMs),
```
SamyPesse: Change it back to 500ms, 1s is a long time to block some features in the docs

PR #4063, opened 2026-02-26
Add server-side insight event tracking for MCP route

nolannbiron:
Extract postInsightsEvents utility from serveProxyAnalyticsEvent to allow sending insight events directly from server-side code. Use it in the MCP route to track search_type_query events.

### DIFF
--- a/packages/gitbook/src/lib/document-sections.test.ts
+++ b/packages/gitbook/src/lib/document-sections.test.ts
@@ -49,14 +49,12 @@ describe('getDocumentSections', () => {
             nodes: [
                 {
                     object: 'block',
-                    // @ts-expect-error columns is missing from top-level blocks, fixed in the next API update
                     type: 'columns',
                     data: {},
                     isVoid: false,
                     nodes: [
                         {
                             object: 'block',
-                            // @ts-expect-error columns is missing from top-level blocks, fixed in the next API update
                             type: 'column',
                             data: {},
                             nodes: [


packages/gitbook/src/lib/tracking.ts
```diff
@@ -23,23 +23,85 @@ export function shouldTrackEvents(headers?: Awaited<ReturnType<typeof nextHeader
...
+/**
+ * Post insight events to the GitBook API.
+ * Can be used from server-side code (e.g. MCP route handlers) to track events directly.
+ */
+export async function postInsightsEvents(args: {
+    organizationId: string;
+    siteId: string;
+    events: ServerInsightsEventInput[];
+    /** Optional request to extract geolocation headers from. */
+    request?: Request;
```
SamyPesse: I think the request should not be optional and we should use it here to fill the event as much as possible

packages/gitbook/src/lib/tracking.ts
```diff
@@ -23,23 +23,85 @@ export function shouldTrackEvents(headers?: Awaited<ReturnType<typeof nextHeader
...
+    space: null,
+    revision: null,
+    page: null,
+};
+
+/**
+ * Post insight events to the GitBook API.
+ * Can be used from server-side code (e.g. MCP route handlers) to track events directly.
+ */
+export async function postInsightsEvents(args: {
```
SamyPesse:
```suggestion
export async function trackServerInsightsEvents(args: {
```

packages/gitbook/src/lib/tracking.ts
```diff
@@ -23,23 +23,85 @@ export function shouldTrackEvents(headers?: Awaited<ReturnType<typeof nextHeader
...
+              session?: Partial<api.SiteInsightsEventSession>;
+              location?: Partial<api.SiteInsightsEventLocation>;
+              timestamp?: string;
+          }
+        : never
+    : never;
+
+const defaultSession: api.SiteInsightsEventSession = {
+    sessionId: '',
+    visitorId: '',
```
SamyPesse: We should really avoid putting this as empty string; instead we should have a function to extract both from the `Request`, which right now will be only equivalent to generating new ones using `crypto.randomUUID()`

packages/gitbook/src/lib/tracking.ts
SamyPesse: Maybe for MCP we could persist them between tool calls in the same MCP session?

packages/gitbook/src/app/sites/static/[mode]/[siteURL]/[siteData]/~gitbook/mcp/route.ts
```diff
@@ -31,6 +36,27 @@ async function handler(
...
+                        siteId: site.id,
+                        events: [
+                            {
+                                type: 'search_type_query',
+                                query,
+                                session: {
+                                    userAgent: nextRequest.headers.get('user-agent') ?? '',
+                                },
+                                location: {
+                                    url,
```
Comment (SamyPesse) [inline packages/gitbook/src/app/sites/static/[mode]/[siteURL]/[siteData]/~gitbook/mcp/route.ts]: Can't it be `request.url` ?

packages/gitbook/src/app/sites/static/[mode]/[siteURL]/[siteData]/~gitbook/mcp/route.ts
```diff
@@ -31,6 +36,27 @@ async function handler(
...
+                    postInsightsEvents({
+                        organizationId: context.organizationId,
+                        siteId: site.id,
+                        events: [
+                            {
+                                type: 'search_type_query',
+                                query,
+                                session: {
+                                    userAgent: nextRequest.headers.get('user-agent') ?? '',
+                                },
```
Comment (SamyPesse) [inline packages/gitbook/src/app/sites/static/[mode]/[siteURL]/[siteData]/~gitbook/mcp/route.ts]: `session` should be automatically filled by `postInsightsEvents`

packages/gitbook/src/app/sites/static/[mode]/[siteURL]/[siteData]/~gitbook/mcp/route.ts
```diff
@@ -31,6 +34,23 @@ async function handler(
...
+                    trackServerInsightsEvents({
+                        organizationId: context.organizationId,
+                        siteId: site.id,
+                        events: [
+                            {
+                                type: 'search_type_query',
+                                query,
+                                location: {
+                                    //!!TODO: Update this when we API is bumped
+                                    displayContext: 'mcp' as SiteInsightsDisplayContext,
```
Comment (SamyPesse) [inline packages/gitbook/src/app/sites/static/[mode]/[siteURL]/[siteData]/~gitbook/mcp/route.ts]: You can update this now as we published the API client

packages/gitbook/src/lib/tracking.ts
```diff
@@ -23,23 +23,92 @@ export function shouldTrackEvents(headers?: Awaited<ReturnType<typeof nextHeader
...
+        session: { ...requestSession, ...event.session },
+        location: { ...defaultLocation, url: request.url, ...event.location },
+        timestamp: event.timestamp ?? new Date().toISOString(),
+    })) as api.SiteInsightsEvent[];
+
+    const url = new URL(
+        `${GITBOOK_API_PUBLIC_URL}/v1/orgs/${organizationId}/sites/${siteId}/insights/events`
+    );
+
+    return await fetch(url.toString(), {
```
SamyPesse: You can use the `GitBookAPI` client instead, also we should not use `GITBOOK_API_PUBLIC_URL`

PR #4067, opened 2026-02-27
Don't use server actions for search

conico974:
This PR replace server actions for search by a route.
Server actions are not the best fit for this use case as we can only have one single server action running at the same time (others get queued) and we cannot cancel them.

### DIFF
--- a/.changeset/shaky-books-hear.md
+++ b/.changeset/shaky-books-hear.md
@@ -1,0 +1,5 @@
+---
+"gitbook": patch
+---
+
+Show social accounts in header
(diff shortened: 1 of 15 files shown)


packages/gitbook/src/components/Search/server-actions.tsx
```diff
@@ -206,88 +162,6 @@ export async function streamRecommendedQuestions(args: { siteSpaceId?: string })
     });
 }
 
-/**
- * Search for content in a site by scoping the search to all content, a specific spaces or current space.
- */
-export async function searchSiteContent({
```
SamyPesse, line 212: Will it crash it for existing site (cached) once we deploy?

SamyPesse (approved):

packages/gitbook/src/components/Search/server-actions.tsx
conico974: No it is handled by skew protection already in 2v, and in 2c it will have the same effect as any new deployment

PR #4070, opened 2026-02-27
Handle circular oneOf with discriminator and allOf in OpenAPI schemas

nolannbiron:
Detect when allOf entries reference an ancestor schema (e.g. Dog allOf: [Pet, ...]) and break the cycle by stripping polymorphic fields before merging. Also default schema titles to component names before dereference so they propagate through $ref resolution.


nolannbiron:
@gregberge can you check this one please

gregberge (approved):

PR #4073, opened 2026-03-02
Add interactive controls to docs assistant

gregberge:


packages/gitbook/src/components/AI/controls/ConfirmControl.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { Button } from '@/components/primitives';
+import { z } from 'zod';
+import { AIToolContainer } from './common';
+import { type GetAIControlProps, createAIControl } from './helpers';
+
+export const ConfirmControlOutputSchema = z.object({
+    result: z.enum(['confirmed', 'cancelled']),
+});
+
+export const ConfirmControlDef = createAIControl({
```
SamyPesse, line 12: Do we actually need this one compared to `SingleChoiceControl`?

Can't the AI outputs a confirmation as a choice between "Yes" / "No".
This is what Codex seems to be doing when asking to implement a plan.

packages/gitbook/src/components/AI/controls/SingleChoiceControl.tsx
```diff
@@ -0,0 +1,192 @@
...
+import { Button, Input } from '@/components/primitives';
+import { tcls } from '@/lib/tailwind';
+import * as React from 'react';
+import { z } from 'zod';
+import { AIToolContainer } from './common';
+import { type GetAIControlProps, createAIControl } from './helpers';
+
+const OTHER_OPTION_ID = '$other';
+
+export const SingleChoiceControlDef = createAIControl({
```
SamyPesse, line 20: I think this is a powerful tool for the AI. I'm wondering if we should not integrate it more in GBX (in the system prompt) and in the API protocol.

The benefits could be to have the assistant use it at better moment, and also to have it work in the agent UI as well.

SamyPesse: Discussed in a call, we'll do this later

packages/gitbook/src/components/AI/controls/SingleChoiceControl.tsx
```diff
@@ -0,0 +1,192 @@
...
+    name: 'single-choice',
+    description: 'Render a question where the user must select exactly one option from a list.',
```
SamyPesse: I think the name and description should written more from the perspective of how the agent should use it.

PR #4162, opened 2026-04-03
Migrate to Next 16

conico974:
Migrate to Next 16.2.3

I had to patch `react-dom-server` in development in order to avoid some OOM issues.
I also had to modify some components because of an hydration error, probably something similar to this https://github.com/radix-ui/primitives/issues/3780
It bump OpenNext to the latest versions


packages/gitbook/next.config.mjs
```diff
@@ -1,10 +1,16 @@
 // @ts-check
 
+// We don't use the deployment ID yet on 2c, we need to remove it because of https://github.com/opennextjs/opennextjs-aws/issues/1136
+const deploymentId =
```
SamyPesse, line 4: Why do we use it on `2v` ? isn't it automatic there?

patches/next@16.2.3.patch
SamyPesse, line 1: Why do we need this patch? do we have a PR on next for it?

SamyPesse (approved):

packages/gitbook/next.config.mjs
conico974: No, because we use the vercel cli https://vercel.com/docs/skew-protection#skew-protection-with-next.js

patches/next@16.2.3.patch
conico974: Without it, `next dev` server will grow OOM, it is just attaching debug info when in dev.
I'll talk about this with the Next team, the fix needs to be done in react anyway

PR #4169, opened 2026-04-08
Calculate a safe viewport to prevent icon clipping

BrettJephson:
Use a safe viewport to prevent clipping on svg icons.


BrettJephson:
Trying a different approach with using svg symbols

PR #4173, opened 2026-04-10
Extend gitbook/expr std lib with some additional methods

spastorelli:
Extends the `@gitbook/expr` std lib symbol table with some additional methods to improve the expression editor autocomplete ([context](https://gitbook.slack.com/archives/C05DA0MQ7J9/p1775676244029489)).

### DIFF
--- a/packages/expr/src/__tests__/autocomplete.test.ts
+++ b/packages/expr/src/__tests__/autocomplete.test.ts
@@ -25,6 +25,7 @@ describe('autocomplete', () => {
         description: 'An array of string',
         items: SymbolString(),
     });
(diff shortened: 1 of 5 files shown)


packages/expr/src/__tests__/autocomplete.test.ts
```diff
@@ -25,6 +25,7 @@ describe('autocomplete', () => {
         description: 'An array of string',
         items: SymbolString(),
     });
+    const visitorClaimsKeyPropStrinSymbol = SymbolString({ name: 'key' });
```
SamyPesse:
```suggestion
    const visitorClaimsKeyPropStringSymbol = SymbolString({ name: 'key' });
```

SamyPesse (approved):

PR #4177, opened 2026-04-13
Make ToC groups collapsible and other styles fixes/improvements

clairechabas:
## Proposed changes
- Make groups in table of contents collapsible.
- Improve focus styles on table of contents items.
- Fix fade displayed on scroll which was displayed on top of groups first items.

## Demo
https://www.loom.com/share/186ecd4e197347798a72595a5f6ee99e

### DIFF
--- a/packages/gitbook/src/components/TableOfContents/PageGroupItem.tsx
+++ b/packages/gitbook/src/components/TableOfContents/PageGroupItem.tsx
@@ -1,24 +1,36 @@
 'use client';
 
+import { AnimatePresence, motion } from 'motion/react';
+import React from 'react';
 import type { ClientTOCPageGroup } from './encodeClientTableOfContents';
 
 import { tcls } from '@/lib/tailwind';
+import { ToggleChevron } from '../primitives';
 
 import { PagesList } from './PagesList';
 import { TOCPageIcon } from './TOCPageIcon';
+import { ToggleableLinkItemStyles } from './styles';
 
 export function PageGroupItem(props: { page: ClientTOCPageGroup; isFirst?: boolean }) {
     const { page, isFirst } = props;
+    const descendants = page.descendants ?? [];
+    const hasDescendants = descendants.length > 0;
+    const [isOpen, setIsOpen] = React.useState(true);
+
+    const handleToggle = () => {
+        if (!hasDescendants) {
+            return;
+        }
+
+        setIsOpen((prev) => !prev);
+    };
 
     return (
         <li className="page-group-item flex flex-col">
             <div
                 className={tcls(
-                    '-top-4 sticky z-1 flex items-center gap-3 px-3',
-                    'font-semibold text-xs uppercase tracking-wide',
-                    'mt-2 pt-4 pb-3', // Add extra padding to make the header fade a bit nicer
-                    '-mb-1.5', // Then pull the page items a bit closer, effective bottom padding is 1.5 units / 6px.
-                    'mask-[linear-gradient(rgba(0,0,0,1)_70%,rgba(0,0,0,0))]', // Fade out effect of fixed page items. We want the fade to start past the header, this is a good approximation.
+                    '-top-4 sticky z-1',
+                    'mt-1 pt-2.5 pb-0',
                     'bg-tint-base',
                     'sidebar-filled:bg-tint-subtle',
                     'theme-muted:bg-tint-subtle',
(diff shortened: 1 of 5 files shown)


packages/gitbook/src/components/TableOfContents/PageGroupItem.tsx
```diff
@@ -27,14 +39,58 @@ export function PageGroupItem(props: { page: ClientTOCPageGroup; isFirst?: boole
...
+                                'toc-group-chevron ml-auto flex shrink-0 transition-opacity duration-150',
+                                isOpen
+                                    ? 'pointer-events-none opacity-0 delay-75'
+                                    : 'opacity-6 delay-0'
+                            )}
+                        >
+                            <ToggleChevron
+                                open={isOpen}
+                                orientation="right-to-down"
+                                className="m-0! size-3!"
```
zenoachtig, line 84:
```suggestion
                                className="m-1 size-3!"
```
The chevrons rely on a little bit of horizontal space to align. Normally it's part of the button they are inside, but since this is just a chevron we need to add some margin ourselves.

packages/gitbook/src/components/TableOfContents/PageGroupItem.tsx
```diff
@@ -1,24 +1,36 @@
...
 
     return (
         <li className="page-group-item flex flex-col">
             <div
                 className={tcls(
-                    '-top-4 sticky z-1 flex items-center gap-3 px-3',
-                    'font-semibold text-xs uppercase tracking-wide',
-                    'mt-2 pt-4 pb-3', // Add extra padding to make the header fade a bit nicer
-                    '-mb-1.5', // Then pull the page items a bit closer, effective bottom padding is 1.5 units / 6px.
-                    'mask-[linear-gradient(rgba(0,0,0,1)_70%,rgba(0,0,0,0))]', // Fade out effect of fixed page items. We want the fade to start past the header, this is a good approximation.
```
zenoachtig, line 21: This removes the faded background of the sticky page group heading, which i think we should keep

packages/gitbook/src/components/TableOfContents/PageGroupItem.tsx
```diff
@@ -27,14 +39,58 @@ export function PageGroupItem(props: { page: ClientTOCPageGroup; isFirst?: boole
...
-                <PagesList pages={page.descendants} />
+            {hasDescendants ? (
+                <AnimatePresence initial={false}>
+                    {isOpen ? (
+                        <motion.div
+                            initial={{ opacity: 0, height: 0 }}
+                            animate={{ opacity: 1, height: 'auto' }}
+                            exit={{ opacity: 0, height: 0 }}
+                            className="mt-px flex flex-col overflow-hidden"
+                        >
```
zenoachtig: Why introduce Motion here? We're kind-of trying to get rid of it in favour of native CSS animation because it wasn't very performant I believe

zenoachtig (changes requested):

packages/gitbook/src/components/TableOfContents/PageGroupItem.tsx
clairechabas: I see. The problem with this fading, as it was, is that it was displayed even when the ToC is not scrolled which make it hide part of the top of the first page item in the group:

Screenshot 2026-04-14 at 12 55 28
Screenshot 2026-04-14 at 12 55 17

I found an efficient and simple way to show the fade on scroll while not displaying it on top of the first item when not scrolling: I'm using a 0-height sentinel and watch with an intersection observer to know when the ToC is being scrolled, in which case the fade is shown.

https://www.loom.com/share/186ecd4e197347798a72595a5f6ee99e

packages/gitbook/src/components/TableOfContents/PageGroupItem.tsx
clairechabas: Just habit, I replaced it with native CSS animations. I agree it's better to default to using platform when we can.

packages/gitbook/src/components/TableOfContents/PageGroupItem.tsx
zenoachtig: Hmm, that's a fine solution I suppose. We do also have a `ScrollContainer` component wrapping the TOC which also has an intersectionobserver for scrolled-to-top or scrolled-to-bottom. Perhaps we could extend/hook into that instead of setting up something custom. But honestly I'm fine with this too if it's performant

zenoachtig (approved):

packages/gitbook/src/components/TableOfContents/PageGroupItem.tsx
clairechabas: Plugging to `ScrollContainer` for this would actually be more expensive for our use case. 

`ScrollContainer` tracks the overall scroll position. Here we need to know for each group item whether it's stuck at the top of the container. Using `ScrollContainer`'s existing logic would mean each `PageGroupItem` would read its offset against `scrollPosition` on every scroll frame, which is expensive. The [sentinel approach](https://developer.chrome.com/docs/css-ui/sticky-headers#sentinels) is cheaper: each group just places an invisible marker above itself and the browser tells us when that marker scrolls out of the ToC container (no continuous measuring needed).

Btw I think using the sentinel approach instead of the current continuous measuring in `ScrollContainer` would be a net win. I can't say for sure as I haven't looked in more detail at what it's used for, but worth testing.

PR #4185, opened 2026-04-14
Add Search tab to Docs Embed, refactor search into an embeddable frame

zenoachtig:
This PR refactors search so it can be rendered as a reusable embeddable frame, while preserving the existing docs-site search experience.

CleanShot 2026-04-14 at 11 12 01@2x

On the docs site, search still uses the current model:
- the input lives outside the results surface
- focusing the input opens the existing popover-based results UI

Inside the docs embed, search can now be mounted as a first-class `search` tab:
- the frame owns its own input
- the same search core is reused instead of maintaining a separate embed-specific UI path

The main goal here was structural. We wanted search to stop being tightly coupled to one popover presentation, so we can support other layouts later without reworking the search logic again.

## What changed

Search is now organized around a shared frame/controller split:
- `SearchFrame` owns the frame-level UI
- `useSearchController` owns search behavior and derived search state
- `SearchContainer` keeps the existing docs-site input + popover composition
- `EmbeddableSearch` renders the same frame directly as an embed tab

This also adds `search` as an embed tab and includes it in the default embed tab set.

## Navigation and embed behavior

A large part of this work was simplifying how embedded navigation works.

The initial version introduced client-side search href rewriting to reinterpret site links inside the embed. That worked, but it pushed search-specific navigation logic into the render tree and expanded `LinkContext` in a way that felt too foundational for a narrow use case.

This PR aligns Search with AIChat instead:
- search result fetches now accept `asEmbeddable`
- the shared search route uses an embeddable linker when requested
- the Ask server action also accepts `asEmbeddable`
- Ask answer body links, Ask source links, and search result links are now generated correctly at the source

This keeps the client simpler and makes Search and AIChat consistent: both now treat embeddable mode as a server-side render/output option.


packages/gitbook/src/app/sites/dynamic/[mode]/[siteURL]/[siteData]/~gitbook/search/route.ts
```diff
@@ -107,55 +92,68 @@ export async function POST(request: NextRequest) {
...
+        }))
+    );
 
     const page: ComputedPageResult = {
         type: 'page',
         id: `${spaceItem.id}/${pageItem.id}`,
         title: pageItem.title,
         href: spaceURL
-            ? linker.toLinkForContent(joinPathWithBaseURL(spaceURL, pageItem.path))
+            ? asEmbeddable
```
Comment (conico974) [inline packages/gitbook/src/app/sites/dynamic/[mode]/[siteURL]/[siteData]/~gitbook/search/route.ts]: Double ternary like that are a bit hard to read, I prefer this one to be outside

packages/gitbook/src/app/sites/dynamic/[mode]/[siteURL]/[siteData]/~gitbook/search/route.ts
```diff
@@ -166,7 +164,9 @@ function transformSitePageResult(
                 id: `${page.id}/${section.id}`,
                 title: section.title,
                 href: spaceURL
-                    ? linker.toLinkForContent(joinPathWithBaseURL(spaceURL, section.path))
+                    ? asEmbeddable
```
Comment (conico974) [inline packages/gitbook/src/app/sites/dynamic/[mode]/[siteURL]/[siteData]/~gitbook/search/route.ts]: Same comment about the ternary

packages/gitbook/src/app/sites/dynamic/[mode]/[siteURL]/[siteData]/~gitbook/embed/search/page.tsx
```diff
@@ -0,0 +1,16 @@
+import type { RouteLayoutParams } from '@/app/utils';
```
Comment (conico974) [inline packages/gitbook/src/app/sites/dynamic/[mode]/[siteURL]/[siteData]/~gitbook/embed/search/page.tsx:1]: Why do we need a dynamic version of this page ?

packages/gitbook/src/components/Search/SearchInput.tsx
```diff
@@ -50,22 +54,36 @@ export const SearchInput = React.forwardRef<HTMLDivElement, SearchInputProps>(
         }, [isOpen, value]);
 
         return (
-            <div className="relative flex @max-2xl:size-9.5 grow">
+            <div
+                className={
+                    mode === 'frame'
```
conico974: You could extract the mode === 'frame' into an isFrame variable

packages/gitbook/src/components/Search/server-actions.tsx
```diff
@@ -198,7 +210,15 @@ async function transformAnswer(
                 const spaceURL = found?.siteSpace.urls.published;
 
                 const href = spaceURL
-                    ? joinPathWithBaseURL(spaceURL, page.page.path)
+                    ? asEmbeddable
```
conico974: Same comment about the ternary

packages/gitbook/src/components/Search/useSearchController.tsx
```diff
@@ -0,0 +1,187 @@
...
+import { useAI } from '../AI';
+import { useTrackEvent } from '../Insights';
+import { useBodyLoaded } from '../primitives';
+import type { SearchResultsRef } from './SearchResults';
+import type { SearchBaseProps } from './search-props';
+import { useSearchState, useSetSearchState } from './useSearch';
+import { useSearchResults } from './useSearchResults';
+import { useSearchResultsCursor } from './useSearchResultsCursor';
+
+export function useSearchController(props: SearchBaseProps) {
```
conico974, line 94: It's probably better to split this one into multiple smaller hooks

packages/gitbook/src/lib/embeddable-linker.ts
```diff
@@ -0,0 +1,63 @@
...
+ * Get a linker to generate links in the embeddable context.
+ */
+export function getEmbeddableLinker(linker: GitBookLinker): GitBookLinker {
+    return {
+        ...linker,
+        toPathForPage({ pages, page, anchor }) {
+            const pagePath = getPagePath(pages, page);
+            const embedPagePath = joinPath('~gitbook/embed/page', pagePath);
+
+            return linker.toPathInSpace(embedPagePath) + (anchor ? `#${anchor}` : '');
```
conico974: We should not use `+` to concatenate string

bun.lock
```diff
@@ -7,7 +7,7 @@
       "devDependencies": {
         "@biomejs/biome": "^1.9.4",
         "@changesets/cli": "^2.30.0",
-        "turbo": "^2.9.2",
+        "turbo": "^2.9.6",
```
conico974, line 10: Why the turbo bump ?

packages/gitbook/src/app/sites/dynamic/[mode]/[siteURL]/[siteData]/~gitbook/search/route.ts
```diff
@@ -16,54 +16,45 @@ import type {
...
 
-type SearchResultGroup = {
-    score: number;
-    items: OrderedComputedResult[];
-};
-
 export async function POST(request: NextRequest) {
-    const [context, { organization, site, shareKey }] = await Promise.all([
+    let [context, siteURLData] = await Promise.all([
         getServerActionBaseContext(),
```
Comment (conico974) [inline packages/gitbook/src/app/sites/dynamic/[mode]/[siteURL]/[siteData]/~gitbook/search/route.ts]: We should make getServerActionBaseContext accept an `isEmbeddable` so that it creates the correct linker there instead of mutating the context

zenoachtig:
@gitbook review

zenoachtig:
@GitBook review and update the docs

zenoachtig:
@gitbook-bot review and update docs

packages/gitbook/src/app/sites/dynamic/[mode]/[siteURL]/[siteData]/~gitbook/embed/search/page.tsx
Comment (zenoachtig) [reply packages/gitbook/src/app/sites/dynamic/[mode]/[siteURL]/[siteData]/~gitbook/embed/search/page.tsx:1]: From Codex, please call its bluff because I don't know static vs dynamic well enough: 

We need the dynamic page for route parity with the rest of the embed surface. Search is now a first-class embed tab, so it needs to exist in both the static and dynamic site trees, just like embed docs and embed assistant. It’s not additional behavior, just the dynamic route entry point using getEmbeddableDynamicContext().

packages/gitbook/src/components/Search/server-actions.tsx
zenoachtig: Seems we bumped to `2.9.6` yesterday but didn't update the `bun.lock`

packages/gitbook/src/app/sites/dynamic/[mode]/[siteURL]/[siteData]/~gitbook/embed/search/page.tsx
Comment (conico974) [reply packages/gitbook/src/app/sites/dynamic/[mode]/[siteURL]/[siteData]/~gitbook/embed/search/page.tsx:1]: I don't think we need it for any of the embed actually, except if we have customization preview on it.
That's the only reason I can think of to have dynamic route for that.
We can do it in another PR though

conico974 (approved):

addisonschultz:
@gitbook can you please make a change request to add this to our docs?

PR #4187, opened 2026-04-14
Refine Assistant phase UI

zenoachtig:
https://github.com/user-attachments/assets/a8263c9d-16e1-4686-a98b-ad4dabed123d


packages/gitbook/src/components/AI/server-actions/api.tsx
```diff
@@ -65,7 +63,14 @@ export async function streamRenderAIMessage(
         return parseResponse<{
             content: React.ReactNode;
             event: AIStreamResponse;
+            steps: AIMessageStep[];
         }>(rawStream, async (event) => {
+            if ('stepIndex' in event && 'phase' in event) {
```
SamyPesse: This should move below as a `case 'response_step_start'`

packages/gitbook/src/components/AI/server-actions/types.ts
```diff
@@ -1,11 +1,12 @@
-import type { AIStreamResponse } from '@gitbook/api';
+import type { AIMessageStep, AIStreamResponse } from '@gitbook/api';
 
 /**
  * Stream when rendering an AI message.
  */
 export type AIMessageRenderStream = AsyncIterable<{
     content: React.ReactNode;
     event: AIStreamResponse;
+    steps: AIMessageStep[];
```
SamyPesse: This is a lot to expose to the frontend in every stream event. We should look at how you use it and reduce it.

SamyPesse: If you only use the last step, I think it should be `currentPhase: AIMessageStepPhase`

SamyPesse: Also if above we forward the whole `event`, then we don't need to forward the accumulated steps everytime; the frontend can build its own message

packages/gitbook/src/components/AI/server-actions/api.tsx
```diff
@@ -67,6 +65,18 @@ export async function streamRenderAIMessage(
             event: AIStreamResponse;
         }>(rawStream, async (event) => {
             switch (event.type) {
+                /**
+                 * A new step started (phase update).
+                 */
+                case 'response_step_start': {
+                    if ('stepIndex' in event && 'phase' in event) {
```
SamyPesse: This is not needed, it'll always be defined on this type of event

SamyPesse (approved):

PR #4194, opened 2026-04-16
Allow to scroll to top of the page

gregberge:
Make "On this page" always visible and a button to scroll to the top of the page

https://github.com/user-attachments/assets/fc30fe6b-0f74-49be-8b89-b06f10fb3fd9

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -23,6 +23,7 @@
             "packages/emoji-codepoints/index.ts",
             "packages/icons/src/data/*.json",
             "packages/gitbook/worker-configuration.d.ts",
+            "gitbook/tsconfig.json",
             "**/*.css"
         ]
     },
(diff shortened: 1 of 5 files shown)


biome.json
```diff
@@ -23,6 +23,7 @@
             "packages/emoji-codepoints/index.ts",
             "packages/icons/src/data/*.json",
             "packages/gitbook/worker-configuration.d.ts",
+            "gitbook/tsconfig.json",
```
zenoachtig, line 26: Necesary?

packages/gitbook/tsconfig.json
zenoachtig, line 1: Necessary?

zenoachtig (approved):

biome.json
gregberge: Yes, it's handled by Next.js, we will keep updating it and we don't care about the format of this file.

PR #4204, opened 2026-04-21
Handle revision fetch failures in streamAskQuestion

conico974:
Ensure that failures in fetching revisions do not disrupt the entire answer process.
If an error is emitted, we display an error message at the end of the message instead of replacing the entire already displayed response with an error message.


utkuufuk (approved):

utkuufuk:
Nitpick: how about a warning log for easier debugging?

SamyPesse (changes requested):
It should not be handled; it should be fixed in gbx

PR #4205, opened 2026-04-21
Allow AI agents to ask questions using ?ask=

SamyPesse:


packages/gitbook/src/lib/markdownPage.ts
```diff
@@ -230,3 +234,55 @@ async function renderGroupPageMarkdown(args: {
...
+ * Re-writes URLs in a markdown content:
+ * -
+ * - the URL of every relative <a> link so it is expressed from the site-root.
+ */
+async function rewriteMarkdownLinks(
+    context: GitBookAnyContext,
+    tree: Root,
+    currentPagePath: string
+): Promise<Root> {
+    const currentDir = path.posix.dirname(currentPagePath);
```
conico974, line 248: I think it is supported, but we need to verify that it works on 2c

packages/gitbook/src/routes/markdownAsk.ts
```diff
@@ -0,0 +1,142 @@
...
+                        title: source.title,
+                        url: source.url,
+                    };
+                }
+
+                const revision =
+                    source.space === context.space.id && source.revision === context.revisionId
+                        ? context.revision
+                        : await throwIfDataError(
+                              context.dataFetcher.getRevision({
```
conico974, line 100: Just had the issue here https://github.com/GitbookIO/gitbook/pull/4204, this one can sometimes return sources that cannot be accessed with the current API token

conico974 (approved):

packages/gitbook/src/lib/markdownPage.ts
SamyPesse: I just copied the code, it was already there

packages/gitbook/src/routes/markdownAsk.ts
SamyPesse: Mmmm this is an upstream problem, not a GBO problem

PR #4216, opened 2026-04-28
Fix hidden pages in sitemap exports

rodrcastro:
## Summary
- Include hidden-but-indexable pages in `llms.txt`, `llms-full.txt`, and sitemap page exports
- Preserve ancestor-aware `noIndex` / `noRobotsIndex` filtering
- Add regression coverage for hidden pages and nested group traversal

## Testing
- `bun test packages/gitbook/src/lib/sitemap.test.ts`
- `bun run format`
- `cd packages/gitbook && bun run typecheck`

Fixes [RND-10663](https://linear.app/gitbook-x/issue/RND-10663/hidden-page-missing-from-llms-fulltxt)


packages/gitbook/src/lib/sitemap.ts
```diff
@@ -9,33 +9,42 @@ export type FlatPageEntry = { page: RevisionPageDocument; depth: number };
...
         if (!allowed) {
             return [];
         }
 
         return [
             ...(page.type === 'document' ? [{ page, depth }] : []),
             ...page.pages.flatMap((child) =>
-                child.type === 'document' ? flattenPage(child, depth + 1) : []
+                child.type !== 'link' && child.type !== 'computed'
+                    ? flattenPage(child as RevisionPageDocument | RevisionPageGroup, depth + 1, [
```
gregberge: Why this change? the `as` is a bad smell here, also why not testing `type === 'group' || type === 'document'` instead of testing the inverse?

rodrcastro: Removed the `as` and rewrote that part as an explicit loop. Does it look better now?

gregberge (approved):

PR #4223, opened 2026-04-29
Move search into a web worker

conico974:
Move search into a web worker to avoid blocking main thread on initial indexing


gregberge (approved):
Seems it is just a port so I trust you.

PR #4224, opened 2026-04-29
Improve markdown serving logic

conico974:
Avoid serving markdown for regular crawling bot

### DIFF
--- a/packages/gitbook/src/middleware.ts
+++ b/packages/gitbook/src/middleware.ts
@@ -3,7 +3,7 @@ import {
     SiteInsightsDisplayContext,
     SiteInsightsLLMSVariant,
 } from '@gitbook/api';
-import { shouldServeMarkdown } from '@vercel/agent-readability';
+import { acceptsMarkdown, isAIAgent, shouldServeMarkdown } from '@vercel/agent-readability';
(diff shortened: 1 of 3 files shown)


packages/gitbook/src/middleware.ts
```diff
@@ -775,7 +775,13 @@ function encodePathInSiteContent(
         default: {
             // If the pathname is a markdown file or the request is ing markdown,
             // we rewrite it to ~gitbook/markdown/:pathname
-            if (pathname.match(MARKDOWN_PATH_REGEX) || shouldServeMarkdown(request).serve) {
+            const aiAgentDetection = isAIAgent(request);
+            // Using heuristic detection incorrectly detects some legitimate bot requests as AI agents (e.g. Slackbot)
+            // We don't want to serve markdown for these requests as it can cause issues like breaking slack unfurling.
+            const shouldServeMarkdown =
```
conico974, line 781: Another option could be to patch `@vercel/agent-readability` so that it properly detects slack, discord or other bots (but we'll need to hardcode a tons of bot, and it might break later or for ones that we don't know)

gregberge (approved):

PR #4227, opened 2026-04-30
Add tests for markdown serving based on user agents

conico974:
Implement tests to verify that the application serves markdown content correctly based on different user agents, ensuring proper handling for AI agents and traditional bots.

### DIFF
--- a/packages/gitbook/tests/robots.test.ts
+++ b/packages/gitbook/tests/robots.test.ts
@@ -1,6 +1,56 @@
 import { describe, expect, it } from 'bun:test';
 import { getContentTestURL } from './utils';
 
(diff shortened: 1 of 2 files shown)


packages/gitbook/tests/robots.test.ts
SamyPesse, line 1: why put these tests in the tests for `robots.txt` 🤔 ?

`markdown.test.ts` would be better

SamyPesse (approved):
Approving but move the tests

packages/gitbook/tests/robots.test.ts
conico974: Not sure I hesitated between the 2, and ended up here :wink: 
I'll move them

PR #4229, opened 2026-04-30
Icons render as svg symbols (fix icon clipping)

BrettJephson:
This change fixes the rendering of Font Awesome icons, which can become clipped.  

Fixes RND-10291

Because we have been loading in the svg as an image file and applying a mask for colour we could **not** simply use `overflow: visible` to prevent clipping of images that overflow the svg’s viewbox. It is already a rasterized image so we only show the bit in the image canvas. 

The solution here is to render svg as an actual svg symbol in the document creating a spritesheet so we are using symbol definitions and `use` and some id. This way we can reuse the svg symbol if it appears in multiple places in a page.

This better aligns with Font Awesome’s recommended usage on the web and fixes the clipping and slightly improves rendering while keeping the size of icons the same.

## Demo
Before:
Screenshot 2026-04-30 at 15 37 17

After:
Screenshot 2026-04-30 at 15 36 54


conico974 (changes requested):
Maybe I misunderstood something, but are we loading the icons with fetch on the client side ?
It means that we have to wait for hydration to be done and for JS to be fully loaded before even attempting to load the icons.
I can also see the behavior on initial load here https://gitbook-v2-fh22isgp4-gitbook.vercel.app/url/gitbook.com/docs, it takes a noticeable amount of times before icon start showing

PR #4235, opened 2026-05-04
bust revision page markdown

conico974:
To be merged and deployed after https://github.com/GitbookIO/gitbook/pull/4234

### DIFF
--- a/packages/gitbook/src/lib/data/api.ts
+++ b/packages/gitbook/src/lib/data/api.ts
@@ -94,9 +94,10 @@ export function createDataFetcher(
                     revisionId: params.revisionId,
                     pageId: params.pageId,
                 },
-                // This API version is used only to invalidate the cache on breaking changes.
-                // It doesn't actually have to be linked to the API version
-                '0.180'
+                // We pass the name of the function to distinguish between cache entries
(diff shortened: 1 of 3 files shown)


packages/gitbook/src/lib/data/api.ts
```diff
@@ -87,11 +87,18 @@ export function createDataFetcher(
...
+            return getRevisionPageMarkdown(
+                input,
+                {
+                    spaceId: params.spaceId,
+                    revisionId: params.revisionId,
+                    pageId: params.pageId,
+                },
+                // We pass the name of the function to distinguish between cache entries
+                // By default Next only uses the arguments, the filename and the position of the function inside the file
+                // By adding a dummy argument with the function name, we can change the order without breaking anything
```
SamyPesse, line 99: But isn't it an issue only with openenxt? should it be fixed there?

SamyPesse (approved):

packages/gitbook/src/lib/data/api.ts
conico974: Nope this is us who are removing the build id to keep the cache persistent between deployment

PR #4238, opened 2026-05-06
Add negotiator for markdown acceptance in middleware

conico974:
Implement middleware to handle markdown acceptance using the negotiator dependency. 
Add Vary headers to responses.

### DIFF
--- a/packages/gitbook/src/middleware.ts
+++ b/packages/gitbook/src/middleware.ts
@@ -62,8 +62,6 @@ export async function middleware(request: NextRequest) {
     try {
         const requestURL = new URL(request.url);
 
-        console.log('accepts markdown', acceptsMarkdown(request));
-
         // Reject malicious requests
         const rejectResponse = await validateServerActionRequest(request);
         if (rejectResponse) {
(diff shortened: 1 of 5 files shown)


conico974:
It gets us to 88/100. I don't think it's a good idea to do the 406, by default we should serve html, not return an error.
image

packages/gitbook/src/middleware.ts
```diff
@@ -61,6 +62,8 @@ export async function middleware(request: NextRequest) {
     try {
         const requestURL = new URL(request.url);
 
+        console.log('accepts markdown', acceptsMarkdown(request));
```
gregberge: log

packages/gitbook/src/middleware.ts
```diff
@@ -498,6 +501,14 @@ async function serveSiteRoutes(requestURL: URL, request: NextRequest) {
         response.headers.set('x-gitbook-route-type', routeType);
         response.headers.set('x-gitbook-route-site', siteURLWithoutProtocol);
 
+        // AI related headers
+        // This one is technically useless, but is used by a bunch of scoring systems
+        // TODO: test if vercel overrides it or does the same as in Next standalone and keep it as is.
+        response.headers.set(
+            'vary',
+            'rsc, next-router-state-tree, next-router-prefetch, next-router-segment-prefetch, accept-encoding, accept'
```
gregberge, line 506: Weird to have to set next internals

gregberge (approved):

packages/gitbook/src/middleware.ts
conico974: Yeah I'm not sure how vercel would handles it, easier to do it this way

PR #4248, opened 2026-05-11
Support code block // [!code -- or ++] notation

viktorrenkema:
## Proposed changes
Following [the app's PR here](https://github.com/GitbookIO/gitbook-x/pull/22959), code blocks now recognize `// [!code ++]` and `// [!code --]` notation — lines tint green (added) or red (deleted), matching how gitbook-x marks diff lines in the editor
- The marker comment is stripped from the rendered output so writers never see `// [!code ++]` leak onto their published page
- Explicitly decided that the red/green don't follow any site customization but instead adopt the same colors we have in the app.
- Supports all comment styles: `//`, `#`, `--`, `;`, `<!-- -->`, `/* */`, plus the no-syntax fallback path
- A line that's both diff-marked AND highlighted gets a deeper shade — same `--green-3/5` / `--red-3/5` tokens as the editor. 
- Tests cover each comment style, the "marker not at end of line" case, the plain highlighter fallback, and a marker co-existing with an inline annotation

## Demo
Editor input:
CleanShot 2026-05-11 at 11 30 18@2x

GBO output:

CleanShot 2026-05-11 at 11 48 13@2x

## Changelog
- [Feature] Code blocks now support diff notation (`// [!code ++]` / `// [!code --]`) — lines tint green or red, the marker comment is hidden from the rendered output, and lines that are also highlighted use a deeper shade


packages/gitbook/src/components/DocumentView/CodeBlock/CodeBlockRenderer.tsx
```diff
@@ -114,27 +114,23 @@ function CodeHighlightLine(props: {
...
-                        color: fg?.color,
-                        ...fg?.vars,
-                        backgroundColor: bg?.color,
-                        ...bg?.vars,
-                    }}
-                />
+            className={tcls(
+                'highlight-line',
+                line.diff === 'added' && 'diff-added',
+                line.diff === 'deleted' && 'diff-deleted',
```
BrettJephson, line 128: Should we add some `+`/`-` prefix to these lines? 

I seem to bring this up a lot but red-green colour blindness is pretty common and we are just using those colours to differentiate.

BrettJephson (approved):
lgtm

packages/gitbook/src/components/DocumentView/CodeBlock/CodeBlockRenderer.tsx
viktorrenkema: Good idea. Added:

CleanShot 2026-05-11 at 16 54 59@2x
CleanShot 2026-05-11 at 16 57 41@2x

PR #4249, opened 2026-05-11
Fix anchor URL resolution

clairechabas:
## Proposed changes
Clicking anchor links was appending the hash to the URL instead of replacing the existing hash.

## Demo
**Before**
https://www.loom.com/share/bdbd4b34eeb54227996667e66c485086

**After**
https://www.loom.com/share/0db353e0128e42bb90f7f56ee9e1412d

### DIFF
--- a/packages/gitbook/src/components/primitives/Link.tsx
+++ b/packages/gitbook/src/components/primitives/Link.tsx
@@ -4,6 +4,7 @@ import NextLink, { type LinkProps as NextLinkProps } from 'next/link';
 import React from 'react';
 
 import { tcls } from '@/lib/tailwind';
+import { checkIsAnchor, resolveAnchorURL } from '@/lib/urls';
 import { type TrackEventInput, useTrackEvent } from '../Insights';
 import { NavigationStatusContext } from '../hooks';
 import { isExternalLink, toNonEmbedLink } from '../utils/link';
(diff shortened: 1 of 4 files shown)


packages/gitbook/src/components/primitives/Link.tsx
```diff
@@ -106,7 +108,10 @@ export function Link(props: LinkProps) {
 
         // When the page is embedded in an iframe
         // for security reasons other urls cannot be opened.
-        if (isInIframe) {
+        if (isAnchor && !isExternal && target !== '_blank' && !event.ctrlKey && !event.metaKey) {
+            event.preventDefault();
+            window.history.pushState(null, '', resolveAnchorURL(href, window.location));
+        } else if (isInIframe) {
```
BrettJephson: The conditional logic of this `if/else if/else if` block isn’t that readable.

The comment on l.109-110 now makes less sense because it is referring to the `isInIframe` condition.

BrettJephson: Also, should we be attempting to `pushState` if this is embedded in an iframe?

BrettJephson (approved):

packages/gitbook/src/components/primitives/Link.tsx
clairechabas: You're right, it was not the right location to fix this and it made me miss another change that was needed in useHash to correclty handle hash-only links. Updated.

PR #4254, opened 2026-05-13
Add missing translations

gregberge:
Add 27 languages to GBO and make the load of language async to avoid loading all languages in memory.


clairechabas (approved):
Lgtm but I don't speak Cantonese so

PR #4257, opened 2026-05-17
Fix: clipping of FA icons

BrettJephson:
Render SVG icons as svg markup to fix clipping issues with FontAwesome


packages/gitbook/src/lib/icons/inline.ts
```diff
@@ -0,0 +1,384 @@
...
+    const record = node as Record<string, unknown>;
+
+    if (record.type === 'icon') {
+        const data = record.data;
+        if (data && typeof data === 'object') {
+            addIconSourceRequest(requests, (data as Record<string, unknown>).icon, iconStyle);
+        }
+    }
+
+    if (record.type === 'button') {
```
conico974: Exact same as icon just above, could be in a single if

packages/gitbook/src/lib/icons/inline.ts
```diff
@@ -0,0 +1,384 @@
...
+
+            return parseRawSVG(await response.text());
+        })
+        .catch(() => null);
+
+    rawSvgPromises.set(cacheKey, request);
+    return request;
+}
+
+function parseRawSVG(document: string): InlineIconSource | null {
```
conico974: We definitely need some test for this

packages/gitbook/src/lib/icons/inline.ts
```diff
@@ -0,0 +1,384 @@
...
+    style: string,
+    icon: IconName
+): Promise<InlineIconSource | null> {
+    const cacheKey = getInlineIconSourceKey(style, icon);
+    const existing = rawSvgPromises.get(cacheKey);
+    if (existing) {
+        return existing;
+    }
+
+    const request = fetch(getIconAssetURL(style, icon), {
```
conico974: We should have a retry on that (with p-retry)

packages/gitbook/src/lib/icons/inline.ts
```diff
@@ -0,0 +1,384 @@
...
+}
+
+function getIconAssetURL(style: string, icon: string): string {
+    const url = new URL(
+        joinPathWithBaseURL(
+            getAbsoluteIconAssetBaseURL(style),
+            joinPath('svgs', style, `${icon}.svg`)
+        )
+    );
+    url.searchParams.set('v', ICON_ASSET_VERSION);
```
conico974: We need to enforce that on every change to any icons, otherwise we will end up in a bad state.
Could you add a comment here and at the top of the docs ?

conico974 (approved):
LGTM, but I'd like to have a bit more tests on this so that we don't break anything in the future

packages/gitbook/src/lib/icons/inline.ts
BrettJephson: Made the source of truth for this the icons package.

PR #4260, opened 2026-05-18
Fix llms.txt not including all content

gregberge:


packages/gitbook/src/routes/llms-full.test.ts
SamyPesse, line 1: Remove these tests please, they are useless

Instead add a real test in https://github.com/GitbookIO/gitbook/blob/main/packages/gitbook/tests/llms.test.ts

SamyPesse (approved):

PR #4264, opened 2026-05-19
Always allow LLMs to read the content

gregberge:


packages/gitbook/src/routes/robots.ts
```diff
@@ -27,7 +35,20 @@ export async function serveRobotsTxt(context: GitBookSiteContext) {
               'Allow: /',
               `Sitemap: ${sitemapUrl}`,
           ]
-        : ['User-agent: *', 'Content-Signal: ai-train=no, search=no, ai-input=no', 'Disallow: /'];
+        : [
+              // Allow user-triggered AI assistants to read pages even when the
+              // site is not indexable, so end-users can pull content into an LLM.
+              // Training crawlers and search engines remain blocked.
+              ...AI_USER_AGENTS.flatMap((userAgent) => [
```
conico974: Not sure if it make sense for VA as it will likely not work as expected though

conico974 (approved):

packages/gitbook/src/routes/robots.ts
gregberge: Good catch, I will do this only for public sites.

PR #4272, opened 2026-05-26
Don't display agents instructions for page action

conico974:
Page markdown actions now defaults to not displaying agents instructions.

Agent accessing the site will still have access to it

### DIFF
--- a/packages/gitbook/src/components/PageActions/PageActions.tsx
+++ b/packages/gitbook/src/components/PageActions/PageActions.tsx
@@ -86,7 +86,7 @@ function useCopiedStore(stateKey: string) {
 }
 
 /**
- * Cache for the markdown versbion of the page.
+ * Cache for the markdown version of the page.
  */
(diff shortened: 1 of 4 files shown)


packages/gitbook/src/components/PageActions/PageActions.tsx
```diff
@@ -160,7 +161,7 @@ export function ActionViewAsMarkdown(props: { markdownPageURL: string; type: Pag
             icon="markdown"
             label={tString(language, 'view_page_markdown')}
             description={tString(language, 'view_page_plaintext')}
-            href={markdownPageURL}
+            href={`${markdownPageURL}?displayAgentInstructions=false`}
```
conico974, line 164: I added that both for `View as Markdown` and for `Copy Page`. Not sure if we should keep it that way or not ?

gregberge (approved):
Approving but I am not convinced of adding instructions in this, it pollutes the content, it's not the place to instruct an agent.

PR #4273, opened 2026-05-27
Make shortcuts work on non-QWERTY keyboards

### LINKED ISSUE #4008: Cmd+c opens assistant instead of copying text with Dvorak
## Bug description

Pressing cmd+c on macOS on https://gitbook.com/docs/?ask= opens the assistant instead of copying text

## How to reproduce

1. Open macOS system settings, set keyboard layout to dvorak
2. Open https://gitbook.com/docs/
3. Select some text
4. Press cmd+c to copy the text

Expected: text is copied
Actual: clipboard content stays the same as what it was (empty or other text I copied previously) and the assistant is opened

## Additional context

macOS 26.2
Google Chrome 144, also an issue on Safari

verhovsky:
I did not test this. You also need to update the library to 5.2.0, the version that introduced this flag.

Fixes https://github.com/GitbookIO/gitbook/issues/4008

https://github.com/JohannesKlauss/react-hotkeys-hook/issues/1323

https://github.com/JohannesKlauss/react-hotkeys-hook#options

### DIFF
--- a/packages/gitbook/src/components/AIChat/AIChat.tsx
+++ b/packages/gitbook/src/components/AIChat/AIChat.tsx
@@ -58,7 +58,9 @@ export function AIChat() {
         () => {
             chatController.close();
         },
-        []
+        {
+            useKey: true,
+        }
     );
 
     // Track the view of the AI chat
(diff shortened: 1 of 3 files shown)


peterwhite:
Thanks for the fix @verhovsky! I've applied it in [#4516](https://github.com/GitbookIO/gitbook/issues/4516), so I'll close this one as superseded. Thanks for tracking down the root cause!

PR #4283, opened 2026-06-01
Adjust caching strategy for icon fetching based on runtime environment

conico974:
Modify the caching strategy for fetching icons to optimize costs in different runtime environments. Use 'force-cache' for Cloudflare and 'no-cache' for other environments.

### DIFF
--- a/packages/gitbook/src/lib/icons/inline.ts
+++ b/packages/gitbook/src/lib/icons/inline.ts
@@ -233,7 +233,8 @@ async function getInlineIconSource(
             () =>
                 fetch(getIconAssetURL(style, icon), {
                     // There is no benefit in caching this in vercel. It just cost tons of money
-                    cache: process.env.GITBOOK_RUNTIME === "cloudflare" ? 'force-cache' : 'no-cache',
+                    cache:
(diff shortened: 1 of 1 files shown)


packages/gitbook/src/lib/icons/inline.ts
```diff
@@ -232,7 +232,9 @@ async function getInlineIconSource(
         const request = pRetry(
             () =>
                 fetch(getIconAssetURL(style, icon), {
-                    cache: 'force-cache',
+                    // There is no benefit in caching this in vercel. It just cost tons of money
```
emmerich:
```suggestion
                    // There is no benefit in caching this in Vercel, as we already cache in the Runtime Cache.
```

Is it worth doing on other providers then?

emmerich (approved):

packages/gitbook/src/lib/icons/inline.ts
conico974: I think it's better to keep it on cloudflare. the fetch cache will be faster on 2c than having to do a dns + hitting the cache

PR #4286, opened 2026-06-03
Update footer for llms

addisonschultz:
Exploring an opportunity to serve information about GitBook across markdown files served by GitBook.

### DIFF
--- a/packages/gitbook/src/routes/llms.ts
+++ b/packages/gitbook/src/routes/llms.ts
@@ -203,12 +203,12 @@ export async function getMarkdownForPagesTree(
 }
 
 function renderAskFooter(context: GitBookSiteContext) {
-    return `\n\n---\n\n# Agent Instructions: Querying This Documentation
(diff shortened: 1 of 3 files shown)


packages/gitbook/src/routes/llms.ts
```diff
@@ -203,23 +203,21 @@ export async function getMarkdownForPagesTree(
 }
 
 function renderAskFooter(context: GitBookSiteContext) {
-    return `\n\n---\n\n# Agent Instructions: Querying This Documentation
+    return `\n\n---\n\n# Agent Instructions
+This documentation is published with GitBook. GitBook is the documentation platform built for the AI era — designed so that both humans and AI agents can read, navigate, and reason over technical content effectively. Learn more at gitbook.com.
```
talagluck: The "built for the AI era" doesn't sound objective enough to me. I think removing just that piece makes it sound more factual, and doesn't change the message too much.

zenoachtig (approved):

PR #4289, opened 2026-06-08
Fix caching strategy for icon fetching

conico974:
Update the caching strategy to prevent storing failed fetch attempts for icons, which could lead to issues in subsequent attempts.

Very visible in dev, where the map is kept around a lot.

### DIFF
--- a/packages/gitbook/src/lib/icons/inline.ts
+++ b/packages/gitbook/src/lib/icons/inline.ts
@@ -234,7 +234,7 @@ async function getInlineIconSource(
                 fetch(getIconAssetURL(style, icon), {
                     // There is no benefit in caching this in Vercel, as we already cache in the Runtime Cache.
                     cache:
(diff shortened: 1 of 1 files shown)


packages/gitbook/src/lib/icons/inline.ts
```diff
@@ -234,7 +234,7 @@ async function getInlineIconSource(
                 fetch(getIconAssetURL(style, icon), {
                     // There is no benefit in caching this in Vercel, as we already cache in the Runtime Cache.
                     cache:
-                        process.env.GITBOOK_RUNTIME === 'cloudflare' ? 'force-cache' : 'no-cache',
+                        process.env.GITBOOK_RUNTIME === 'cloudflare' ? 'force-cache' : 'no-store',
```
conico974, line 237: `no-cache` is not documented by vercel, let's use `no-store` instead https://nextjs.org/docs/app/api-reference/functions/fetch#optionscache

clairechabas (approved):

PR #4290, opened 2026-06-08
Fix context handling in resolveContentRef for page resolution issues

conico974:
Improve context management in `resolveContentRef` to prevent resolution problems when referencing pages in different spaces. This change addresses potential conflicts arising from duplicate page IDs across spaces.

### DIFF
--- a/packages/gitbook/src/lib/references.tsx
+++ b/packages/gitbook/src/lib/references.tsx
@@ -164,7 +164,15 @@ export async function resolveContentRef(
         case 'anchor':
         case 'page': {
             if (isContentRefInDifferentSpace(contentRef, context)) {
-                return resolveContentRefInSpace(contentRef.space, context, contentRef, options);
+                let contextWithoutPage: GitBookAnyContext = context;
+                if ('page' in contextWithoutPage) {
+                    // We need to remove the page from the context to avoid issues when resolving the content ref in the target space.
+                    // The problem happens when the same page exists in the target space (i.e. a duplicate space), it then contaminates the resolution because it looks like the page is already resolved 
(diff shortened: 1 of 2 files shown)


packages/gitbook/src/lib/references.tsx
```diff
@@ -164,7 +164,20 @@ export async function resolveContentRef(
         case 'anchor':
         case 'page': {
             if (isContentRefInDifferentSpace(contentRef, context)) {
-                return resolveContentRefInSpace(contentRef.space, context, contentRef, options);
+                let contextWithoutPage: GitBookAnyContext = context;
+                if ('page' in contextWithoutPage) {
+                    // We need to remove the page from the context to avoid issues when resolving the content ref in the target space.
+                    // The problem happens when the same page exists in the target space (i.e. a duplicate space), it then contaminates the resolution because it looks like the page is already resolved
+                    // while it is not the case as it is a different page with the same id in another space.
+                    const { page, ...rest } = contextWithoutPage;
```
spastorelli, line 172: Aren't we already excluding the page from the context in `resolveContentRefInSpace` when doing the `resolveContentRef` resolution since we pass `ctx.spaceContext`?
https://github.com/GitbookIO/gitbook/blob/main/packages/gitbook/src/lib/references.tsx#L484

Or are we excluding it here also for an issue outside the `resolveContentRef`?

spastorelli (approved):

packages/gitbook/src/lib/references.tsx
conico974: ctx.spaceContext is built using spread syntax, which means that it includes page by default

PR #4291, opened 2026-06-09
Add support for passing visitor claims to integration webframes

spastorelli:
**Demo**

https://github.com/user-attachments/assets/015a5060-4ae8-4ee0-beee-d0fca2954b87

using demo integration here: https://github.com/GitbookIO/integrations/pull/1195


conico974 (approved):
LGTM, but we'll need to update the docs for that

PR #4293, opened 2026-06-10
Fix inflated visitor counts from unstable anonymous visitorId

nolannbiron:
## Summary

Anonymous visitors were counted as new on every page load, inflating the **Visitors** metric (it tracked roughly *total visits*). This restores a stable `visitorId` per visitor while keeping the signed-in detection introduced in #4061.

## Background

#4061 (RND-8584) hardened visitor session detection so the GBO toolbar reliably opens for signed-in users. In doing so (commit `c5bdde44`), the client started discarding the stable `deviceId` returned by the cross-origin `GET /__session/2/` worker for anonymous visitors and used a fresh random id instead.

On custom-domain sites the `__session` cookie is third-party and often unreadable, so the client resolves through the worker on every load — anonymous visitors got a new id each page view and were counted as new every time.

## What this changes

In `getGlobalVisitor` (`packages/gitbook/src/components/Insights/visitorId.tsx`):

- **Honor the worker's `deviceId` for every visitor** — removed the anonymous-only branch that returned a per-load random id, so the server's stable id is always used.
- **Persist the resolved visitor in `localStorage`** so it survives third-party-cookie blocking, Safari ITP, and custom domains where the cookie can't be read.
- **Seed the worker's `proposed` param with a `stableId`** (first-party cookie → persisted localStorage → fresh id). When the worker can't read the cookie, it echoes this id back instead of minting a new one.
- **Resolve from the known visitor immediately, revalidate only when stale (>1h)** — no `/__session/2/` request on every page view.

### DIFF
--- a/packages/gitbook/src/components/Insights/visitorId.tsx
+++ b/packages/gitbook/src/components/Insights/visitorId.tsx
@@ -5,7 +5,12 @@ import type { MaybePromise } from 'p-map';
 import React from 'react';
 import { createStore, useStore } from 'zustand';
 
-import { getBrowserCookie, getLocalStorageItem, setLocalStorageItem } from '@/lib/browser';
+import {
+    getBrowserCookie,
+    getLocalStorageItem,
+    removeLocalStorageItem,
+    setLocalStorageItem,
+} from '@/lib/browser';
 
 import { isCookiesTrackingDisabled } from './cookies';
 import { getSession } from './sessions';
@@ -169,6 +174,8 @@ function getGlobalVisitor({
     const withoutCookies = isCookiesTrackingDisabled();
 
     if (withoutCookies || !visitorCookieTrackingEnabled) {
+        // Consent withdrawn or tracking disabled: stop using AND retaining the stored id.
+        removeLocalStorageItem(VISITOR_STORAGE_KEY);
         return { visitor: { deviceId: getProposedVisitorId() }, pendingVisitor: null };
     }
 
--- a/packages/gitbook/src/components/Insights/visitorId.tsx
+++ b/packages/gitbook/src/components/Insights/visitorId.tsx
@@ -44,10 +44,23 @@ interface StoredVisitor {
     updatedAt: number;
 }
 
+interface StoredVisitorRaw {
+    visitor?: unknown;
+    deviceId?: unknown;
+    updatedAt?: unknown;
+}
+
 function getStoredVisitor(): StoredVisitor | null {
-    const value = getLocalStorageItem<StoredVisitor | null>(VISITOR_STORAGE_KEY, null);
-    if (value && isVisitor(value.visitor) && typeof value.updatedAt === 'number') {
-        return value;
+    const value = getLocalStorageItem<StoredVisitorRaw | null>(VISITOR_STORAGE_KEY, null);
+    if (!value || typeof value.updatedAt !== 'number') {
+        return null;
+    }
+    if (isVisitor(value.visitor)) {
+        return { visitor: value.visitor, updatedAt: value.updatedAt };
+    }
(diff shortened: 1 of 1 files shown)


packages/gitbook/src/components/Insights/visitorId.tsx
```diff
@@ -174,10 +172,16 @@ function getGlobalVisitor({
...
+    const stored = getStoredVisitor();
+
+    // Cookie wins, localStorage survives third-party cookie blocking, proposed id is the fallback
+    const stableId = existing?.deviceId ?? stored?.visitor.deviceId ?? proposedId;
+
     const fetchGlobalVisitor = async () => {
         const url = new URL(appURL);
         url.pathname = '/__session/2/';
-        url.searchParams.set('proposed', proposedId);
+        url.searchParams.set('proposed', stableId);
```
gregberge, line 191: Why do we provide the stableId in "proposed" are you sure about this change?

packages/gitbook/src/components/Insights/visitorId.tsx
```diff
@@ -198,33 +202,26 @@ function getGlobalVisitor({
...
-                clearVisitorUpdatedAt();
-                return { deviceId: proposedId };
-            }
-
-            setVisitorUpdatedAt(visitor.deviceId, Date.now());
+            setStoredVisitor(visitor, Date.now());
 
             return visitor;
         } catch (error) {
-            clearVisitorUpdatedAt();
```
gregberge, line 212: So in case of error we don't clear?

gregberge (approved):
It looks good to me. I think it's OK to deduplicate the visitor even if he doesn't accept cookies.

But we may use this for other types of tracking. Let's be sure it's correct from a GDPR point of view before merging this.

packages/gitbook/src/components/Insights/visitorId.tsx
nolannbiron: if the worker can't read the `__session` cookie (blocked cookie on custom domains), it returns `proposed`.

so before `proposed` was a new random id on every load → new visitor every page view, now we propose the id we already know, so the worker returns the same stable id

packages/gitbook/src/components/Insights/visitorId.tsx
nolannbiron: yes, it's on purpose, this would only cleared the timestamp (to retry later), not the id. Now on error we keep the stored id (clearing it would bring the bug back) and we don't update updatedAt, so it stays stale and retries on the next page load

nolannbiron:
@gregberge `localStorage` is only written when cookies are accepted, so it follows the same consent as the `__session` cookie. I also added a cleanup: if the visitor rejects cookies later, we delete the stored id on the next page load

packages/gitbook/src/components/Insights/visitorId.tsx
gregberge: ok

packages/gitbook/src/components/Insights/visitorId.tsx
gregberge: ok

gregberge:
> @gregberge `localStorage` is only written when cookies are accepted, so it follows the same consent as the `__session` cookie. I also added a cleanup: if the visitor rejects cookies later, we delete the stored id on the next page load

OK but then how does it solve the bug?

nolannbiron:
the bug is not about consent, we were generating a new id on every page load when the `__session` cookie was blocked by Safari/custom domains, localStorage fixes the id re-generation whenever cookie tracking is allowed (same condition as the cookie)<br><br>and visitors who rejected cookies keep the old behavior, random id per page load, they're not the cause of the inflation since it's a small fraction of traffic

PR #4299, opened 2026-06-11
add navigateToPage assistant tool

zenoachtig:
## What

Adds a built-in `navigateToPage` assistant tool that lets the AI open a documentation page on the reader's behalf (fixes RND-11321).

## Behaviour

- **No confirmation** — the tool executes directly; the page opens **instantly** the moment the assistant calls it (mid-turn, not after the turn finishes).
- **SPA navigation via the router** — uses `router.push`, so it adds a **browser history entry** (the reader can go back) and does **not** hard-reload or lose the chat conversation.
- **Tells the assistant it navigated** — the tool result returns `navigated: true`.

## How it works

1. The assistant references pages using GitBook's stable content-ref scheme (e.g. `/spaces/<id>/pages/<id>`), which isn't a navigable published-site URL. A small server action (`resolveAINavigationLink`) resolves it to the real site path using the site context/linker.
2. The tool then `router.push`es to that path. A bare push here gets cancelled, because sending the tool result back invokes a server action whose router refresh aborts an in-flight navigation. To avoid that, `execute` **awaits the navigation commit** (the URL actually changing) before returning — so the result-send refresh lands on the new page harmlessly. No artificial delay, no deferral into the chat loop.

## Files

- `useNavigateToPageTool.ts` — the tool (self-contained `execute`: resolve → same-origin guard → push → await commit → report).
- `navigation.ts` (+ `navigation.test.ts`) — pure helpers: `resolveNavigationTarget` (same-origin guard) and `normalizePathname` (decode + trailing-slash normalization for the commit check).
- `server-actions/navigate.ts` — content-ref → site-path resolver.
- `tools.ts` / `useAIChat.tsx` — register the built-in tool; `useAIChat` holds **no** navigation logic.
- `intl/translations/*` — `ai_chat_tools_navigated_to_page` / `ai_chat_tools_navigate_failed` across all 39 locales.

## Reviewer notes

- Design was vetted with an adversarial review across Next.js App Router semantics, the chat streaming flow, edge cases, and dead-code removal. Key hardening: normalized commit comparison (so legit navigations don't false-timeout under proxy/encoding/trailing-slash), `router`/`onNavigationClick` captured in refs (the `execute` closure is memoized), and a fail-safe timeout that resolves + warns rather than hanging the turn.
- Same-origin guard prevents the assistant from navigating the reader off-site.
- Known minor edge (not a regression): cross-page `#section` deep-links rely on the app's existing post-commit hash-scroll handling.

## Testing

- Unit: `bun test packages/gitbook/src/components/AI/navigation.test.ts` (8 passing).
- `bun run typecheck` and `bun run format` clean.
- Verified end-to-end against a live published site: asked the assistant to open a page → navigated instantly to the real site URL, back button works, and the chat conversation survived the navigation.

### DIFF
--- a/packages/gitbook/src/components/AI/navigation.ts
+++ b/packages/gitbook/src/components/AI/navigation.ts
@@ -1,0 +1,42 @@
+import { isExternalLink } from '../utils/link';
+
+/**
+ * Resolve a link into a same-site target to navigate to.
+ *
+ * Returns an `error` when the URL is malformed or points outside of the documentation site,
+ * so the assistant can be told it could not navigate.
+ */
+export function resolveNavigationTarget(
+    url: string,
+    location: { href: string; origin: string }
+): { href: string; pathname: string } | { error: string } {
+    let target: URL;
+    try {
+        target = new URL(url, location.href);
+    } catch {
+        return { error: `Invalid URL: ${url}` };
+    }
+
+    // Only allow navigating within the current documentation site to avoid sending the user to
+    // an external website without their consent.
+    if (isExternalLink(target.href, location.origin)) {
+        return { error: 'Cannot navigate to a page outside of this documentation site.' };
+    }
+
+    return { href: `${target.pathname}${target.search}${target.hash}`, pathname: target.pathname };
+}
+
+/**
+ * Normalize a pathname so two equivalent paths compare equal regardless of percent-encoding or a
+ * trailing slash (e.g. `/h%C3%A9llo/` and `/héllo`). Used to detect when an SPA navigation has
+ * committed by comparing against `window.location.pathname`.
+ */
+export function normalizePathname(pathname: string): string {
+    let decoded = pathname;
+    try {
+        decoded = decodeURIComponent(pathname);
+    } catch {
+        // Keep the raw value if it isn't valid percent-encoding.
+    }
+    return decoded.length > 1 && decoded.endsWith('/') ? decoded.slice(0, -1) : decoded;
+}
--- a/packages/gitbook/src/components/AI/navigation.ts
+++ b/packages/gitbook/src/components/AI/navigation.ts
@@ -1,3 +1,4 @@
+import { withTrailingSlash } from '@/lib/paths';
 import { isExternalLink } from '../utils/link';
 
 /**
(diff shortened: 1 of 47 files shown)


packages/gitbook/src/components/AI/navigation.ts
```diff
@@ -0,0 +1,42 @@
...
+    let target: URL;
+    try {
+        target = new URL(url, location.href);
+    } catch {
+        return { error: `Invalid URL: ${url}` };
+    }
+
+    // Only allow navigating within the current documentation site to avoid sending the user to
+    // an external website without their consent.
+    if (isExternalLink(target.href, location.origin)) {
```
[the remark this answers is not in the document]
zenoachtig: Fixed in cfc2f312e. Non-content-ref URLs now resolve through the site linker (`toInSiteHref`), which enforces both the host and the site base path via `toLinkForContent`/`siteBasePath` — so a same-host URL outside the site (e.g. `gitbook.com/pricing` or another `/url/...` proxied site) is rejected. Added unit tests covering that exact case.

packages/gitbook/src/components/AI/useAIChat.tsx
```diff
@@ -192,6 +193,14 @@ export function AIChatProvider(props: {
     const { siteSpaceId } = useCurrentContent();
     const language = useLanguage();
 
+    // Built-in tools exposed to the assistant (e.g. navigating to a page). They are kept in a
+    // ref so the streaming callback can read the latest definitions without being recreated.
+    const navigateToPageTool = useNavigateToPageTool();
+    const builtInToolsRef = React.useRef([navigateToPageTool]);
+    React.useEffect(() => {
+        builtInToolsRef.current = [navigateToPageTool];
+    }, [navigateToPageTool]);
```
gregberge: Do we have `useLiveRef`?

gregberge (approved):

packages/gitbook/src/components/AI/useAIChat.tsx
zenoachtig: No, there's no `useLiveRef` in the codebase (the closest is the local `useAIMessageContextRef` pattern). But here the ref was actually unnecessary: `useNavigateToPageTool` returns a `useMemo([])`-stable tool, so I dropped the ref+effect entirely (4594e28ec) and reference the tool directly via the callback deps.

PR #4324, opened 2026-06-18
Support agent `goal` query param for `markdownAsk`

zenoachtig:
## What

Adds support for an optional `goal` query parameter on the markdown "ask" interface for published sites. Alongside the existing `?ask=<question>`, an agent can now pass `?goal=<end goal>` describing the broader outcome it is trying to accomplish on behalf of the user. The goal is forwarded to the backend, which uses it to steer the answer toward what is most useful for that goal — without changing the literal question being answered.

This is the **frontend half** of the feature. The backend half (API contract + prompt steering) lives in GitbookIO/gitbook-x#23519.

## Why

`?ask=` answers the immediate question, but an agent usually has a larger task in mind (e.g. *"migrate the user off long-lived API keys to short-lived OAuth tokens"*). Passing that end goal lets the assistant prioritize the information and next steps that actually move the agent forward, instead of answering the question in a vacuum.

## How it works

- **`middleware.ts`** — parses `?goal=` next to `?ask=`. Because the markdown-ask route is statically rendered (`force-static`) and cannot read query params at request time, the goal is encoded as a second path segment (`~gitbook/markdown-ask/<question>/<goal>`), mirroring how the question itself is already path-encoded.
- **`markdown-ask/[question]/[goal]/route.ts`** — new nested route that decodes both segments and calls `serveAskMarkdown(context, question, { goal })`. The existing no-goal route is untouched, so `?ask=` on its own behaves exactly as before.
- **`markdownAsk.ts`** — forwards `goal` in the `streamAskInSite` request `context`, and carries it into the suggested follow-up question links (`&goal=…`) so a multi-step agent keeps its goal across hops.
- **`markdownPage.ts`** — the "Agent Instructions" footer (and the 404 page's ask option) now document `goal` in the canonical example: `?ask=<question>&goal=<end goal>`, with prose marking `goal` optional and noting that `ask` alone is still valid. Putting both params in the single example biases agents toward filling in the goal by default.

## Backend dependency / temporary cast

`goal` is not part of the published `@gitbook/api` types yet, so `markdownAsk.ts` includes a small typed cast (commented as such). Once the spec change in GitbookIO/gitbook-x#23519 is regenerated and the public client is published, the cast can be removed.

## Testing

- `?ask=` alone — unchanged behavior.
- `?ask=&goal=` — the goal flows through to the backend (takes effect against an API carrying the gitbook-x#23519 changes); follow-up links carry the goal forward.
- Works for both `.md` URLs and `Accept: text/markdown` content negotiation.

*— Authored by Claude*


packages/gitbook/src/routes/markdownAsk.ts
```diff
@@ -32,6 +48,9 @@ export async function serveAskMarkdown(context: GitBookSiteContext, rawQuestion:
                 question,
                 context: {
                     siteSpaceId: context.siteSpace.id,
+                    // POC: `goal` is not yet part of the published `@gitbook/api` types.
+                    // The backend reads it from `context.goal`; cast until the client is regenerated + published.
+                    ...(goal ? ({ goal } as { goal: string }) : {}),
```
zenoachtig: Won't merge before https://github.com/GitbookIO/gitbook-x/pull/23519 is in so we can update with proper typing

zenoachtig: Done — gitbook-x#23519 landed and shipped in `@gitbook/api@0.185.0`, which types `goal` on the `streamAskInSite` context. Merged `main` in and dropped the cast in 5fd18c42f, so it's now properly typed.

*— Authored by Claude*

nolannbiron (approved):

PR #4325, opened 2026-06-18
Fix NextLink prefetch side effect on unsigned claims persistance

spastorelli:
This PR aims to fix an issue where NextLink prefetch triggering RSC requests for links including `ask` or `visitor.*` query params has a side effect on unsigned claims cookie persistant and on the AI assistant opening/closing multiple times.

### DIFF
--- a/packages/gitbook/src/components/primitives/Link.tsx
+++ b/packages/gitbook/src/components/primitives/Link.tsx
@@ -180,13 +180,10 @@ export function Link(props: LinkProps) {
     );
 }
 
-const ASK_QUERY_PARAM = 'ask';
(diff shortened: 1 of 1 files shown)


packages/gitbook/src/components/primitives/Link.tsx
```diff
@@ -175,6 +180,34 @@ export function Link(props: LinkProps) {
     );
 }
 
+const ASK_QUERY_PARAM = 'ask';
+const VISITOR_PARAM_PREFIX = 'visitor.';
```
conico974: We should add the `q` param as well

conico974 (approved):
I approve with just a small nit

PR #4326, opened 2026-06-18
Add AI chat reference button on text selection

nolannbiron:
Selecting text in the page content shows a floating "Ask" button that stages the selection as a reference and opens the AI chat. Desktop only (Assistant mode).

### DIFF
--- a/packages/gitbook/src/components/AIChat/AskAITextSelection/AskAITextSelection.tsx
+++ b/packages/gitbook/src/components/AIChat/AskAITextSelection/AskAITextSelection.tsx
@@ -1,0 +1,129 @@
+'use client';
+
+import { CustomizationAIMode } from '@gitbook/api';
+import fnv1a from '@sindresorhus/fnv1a';
+import { AnimatePresence, motion } from 'motion/react';
+import * as React from 'react';
+import { createPortal } from 'react-dom';
+
+import { useAIChatController, useAIConfig } from '@/components/AI';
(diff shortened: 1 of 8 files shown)


gregberge (approved):

packages/gitbook/src/components/AIChat/AskAITextSelection/AskAITextSelection.tsx
```diff
@@ -0,0 +1,129 @@
...
+                        top: coords?.top ?? 0,
+                        left: coords?.left ?? 0,
+                        zIndex: 40,
+                    }}
+                    // Keep the selection alive: prevent the button from stealing focus on click.
+                    onMouseDown={(event) => event.preventDefault()}
+                >
+                    <Button
+                        size="small"
+                        variant="primary"
```
zenoachtig, line 117:
```suggestion
                        variant="secondary"
```
Did you consider making it secondary instead? It looks a bit garish in normal content, but maybe secondary is too subtle

zenoachtig (approved):

packages/gitbook/src/components/AIChat/AskAITextSelection/AskAITextSelection.tsx
nolannbiron: I think `primary` is the best one for visibility, also `secondary` doesn't have a background, `header` does

PR #4330, opened 2026-06-20
Make the AI Assistant panel resizable

nolannbiron:
Add a drag handle on the panel's left edge and an expand/collapse button to resize the AI Assistant on desktop. Width is driven by the `--ai-chat-width `CSS variable, persisted via a Zustand store, and capped to the viewport so the page content stays usable.

Demo:

https://github.com/user-attachments/assets/7a6b112a-43ed-4183-9b31-28b1bb3fdcd6


gregberge (approved):
Brillant! I love it!

PR #4338, opened 2026-06-23
Redesign the 404 page with related-page suggestions and a search input

zenoachtig:
## Overview

- Removes the table of contents sidebar on the 404 page and centers the content — purely CSS (`page-no-toc` + `layout-wide`), no layout/TOC plumbing changes.
- Suggests related pages, ranked server-side with `getSimilarPages` (the same ranking the Markdown 404 already uses). The path that 404'd is passed to a server action, so the full page tree never ships to the client.
- Shows a search **or** assistant input depending on the site's AI configuration: the assistant when one is defined, otherwise a search input that opens the search menu.
- For adaptive content, surfaces the login button with a hint that registration may be available depending on the organization.
- Moves "Go to homepage" to the bottom as a last resort.

## Demo

CleanShot 2026-06-23 at 20 05 47@2x
CleanShot 2026-06-23 at 18 59 05@2x
CleanShot 2026-06-23 at 18 58 55@2x
CleanShot 2026-06-23 at 18 59 14@2x
CleanShot 2026-06-23 at 18 59 23@2x
CleanShot 2026-06-23 at 18 59 48@2x
CleanShot 2026-06-23 at 19 02 57@2x

*— Authored by Claude*


packages/gitbook/src/components/SitePage/SitePageNotFound.tsx
```diff
@@ -1,89 +1,237 @@
...
+        if (ask) {
             router.replace(`${basePath}?${searchParams?.toString()}`);
+            return;
         }
-    }, [basePath, fallback, ask, router, searchParams]);
+
+        // Otherwise, load the pages most similar to the path that 404'd. The ranking happens
+        // server-side (same `getSimilarPages` as the Markdown 404) so the page tree stays off the client.
+        let active = true;
+        getRelatedPages(getRequestedPath(pathname ?? '', basePath)).then(
```
gregberge: Why using a server-action for this and not just using an async component for the 404 page?

gregberge (approved):

packages/gitbook/src/components/SitePage/SitePageNotFound.tsx
zenoachtig: The suggestions need the path that 404’d, and an async not-found.tsx can’t get it:

- Next’s `not-found` boundary receives no params / searchParams / pathname on the server, so it can’t know which path failed (it renders with no props).
- We could stash the path in a middleware header and read it from an async `not-found.tsx` — but the static route (`app/sites/static/.../[pagePath]/page.tsx` is `force-static`, where `headers()` returns empty values), so static sites would silently get no suggestions.
- Rendering the not-found UI straight from the page (where we do have the path + context) would mean dropping `notFound()` and serving a `200` soft-404 instead of a real `404`.

A server action is a separate, always-dynamic request, so it gets the path from the client, runs the same `getSimilarPages` server-side, and works identically on static and dynamic sites while keeping the proper 404 status. The cost is one round-trip — hence the skeleton.

zenoachtig: Went back to the drawing board after a discussion with @conico974. We're now using the `~gitbook/site-index` and doing the filtering clientside, which prevents a request round-trip 🎉

packages/gitbook/src/components/SitePage/SitePageNotFound.tsx
```diff
@@ -1,89 +1,290 @@
...
+
+const normalizePath = (path: string) => path.toLowerCase().replace(/^\/+|\/+$/g, '');
+const lastSegment = (path: string) => path.slice(path.lastIndexOf('/') + 1);
+const similarity = (left: string, right: string) => {
+    const max = Math.max(left.length, right.length);
+    return max === 0 ? 1 : 1 - leven(left, right) / max;
+};
+
+/** Path similarity in [0, 1] — full-path closeness, weighted toward the last segment. */
+function scorePathSimilarity(input: string, candidate: string): number {
```
conico974, line 283: We can probably improve this in the future, but for now I guess it is good enough

conico974 (approved):

PR #4349, opened 2026-06-29
Keep first 3 remote results pinned at the top

conico974:
Keep the first 3 remote results pinned at the top (We assume that remote is always better)
Also don't fully crash if remote fails but local works


utkuufuk (approved):
Looks good to me, though I didn't review the UI aspect.

zenoachtig (approved):
UI looks good. A bit worried that the notice is at the bottom of the results list instead of at the top, but I suspect it's also a bit much to call attention to at the top.

conico974:
>  A bit worried that the notice is at the bottom of the results list instead of at the top, but I suspect it's also a bit much to call attention to at the top.

That was my thought as well, we can always revisit this later if needs be

PR #4350, opened 2026-06-29
Add "Available in MCP" badge for OpenAPI operations

nolannbiron:
Adds an "Available in MCP" badge to OpenAPI operations, driven by the `x-gitbook-mcp` extension.

## What

- Operations with `x-gitbook-mcp: true` show an **"Available in MCP"** badge next to the title, alongside the existing Deprecated / stability badges.
- When `x-gitbook-mcp-url` is set, the badge becomes a button that **copies the MCP server URL** to the clipboard (reusing the existing `OpenAPICopyButton`).
- `x-gitbook-mcp-url` resolves with cascade precedence **operation → path → root** (mirroring `getServers`), so it can be set once at the spec root and overridden per path or operation.
- The label is **localized** across all supported locales; the badge uses the `mcp` icon and an emerald theme consistent with the existing badge styles.

## Extensions

| Extension | Level | Purpose |
|---|---|---|
| `x-gitbook-mcp: boolean` | operation | Gates whether the badge is shown (per-endpoint). |
| `x-gitbook-mcp-url: string` | operation / path / root | Optional MCP server URL; turns the badge into a copy-to-clipboard button. |

## Demo

CleanShot 2026-06-29 at 21 52 04@2x

CleanShot 2026-06-29 at 21 52 08@2x

Resolves RND-11615.


gregberge (approved):

zenoachtig (approved):
I have some doubts about the placement and understandability of the badge, but let's not block it and iterate as we get feedback on it

gregberge:
On my side I don't understand why we don't expose everything. If this just for us or for other customers also?

nolannbiron:
> I have some doubts about the placement and understandability of the badge

@zenoachtig Yeah I agree it's not perfect in terms of readability. I've put it in the same row as the other badges, but happy to move it if you have a better idea for this one (or all badges ?)

> On my side I don't understand why we don't expose everything. If this just for us or for other customers also?

@gregberge wdym by "expose everything" ? This could be used by anyone who did `OpenAPI -> MCP` it's pretty specific but it's what we did, I'll document it in our docs

PR #4362, opened 2026-07-03
Add a navigate action to integration block webframes

gregberge:
## What

Lets an integration block webframe navigate the reader to another page in the site, over the client-only webframe postMessage channel (the same channel as adaptive visitor claims):

- `@webframe.navigate` — a webframe posts this action to navigate the reader to another page in the site, addressed by `path` (with an optional `anchor`). The path is resolved against the site base path, so a webframe can link to a page in any section or space of the site — and nowhere outside it.

Navigation is resolved entirely client-side (never through the integration render request), so the integration render cache is unchanged, and it drives the standard navigation progress bar.

## Usage (from inside a webframe)

The webframe communicates with GitBook over `postMessage`, exactly like the existing resize/ready protocol. Post a `@webframe.navigate` action with a site-root-relative `path`:

```js
// Navigate to another page in the site
window.parent.postMessage(
    { action: { action: '@webframe.navigate', path: 'getting-started/quickstart' } },
    '*'
);

// Optionally scroll to a heading anchor
window.parent.postMessage(
    { action: { action: '@webframe.navigate', path: 'getting-started/quickstart', anchor: 'installation' } },
    '*'
);
```

## Why

**RND-11604** — integrations (e.g. harness) need to navigate the reader to other pages within the site from a webframe. Docs: GitbookIO/integrations#1230.

## How it's wired

- The ContentKit client context exposes `navigate({ path, anchor? })`; `ElementWebframe` maps the `@webframe.navigate` message to it.
- The path is resolved on the client through the (tested) `GitBookLinker` (`toPathInSite`), rebuilt from serializable data passed by `IntegrationBlock` — no duplicated join logic, no server round-trip.

## Notes

- No new integration scope, and no `@gitbook/api` or backend change — the gitbook-x companion PR was closed.
- Scope was reduced during review: the earlier `state.page` current-page context and `pageId`-based navigation were removed; this PR now only adds `@webframe.navigate` by path. Pre-existing visitor claims are unchanged.

## Test plan

- `bun run typecheck` — 27/27
- `bun test` on `adaptive.test.ts` — 3/3
- `biome check` on changed files — clean


packages/gitbook/src/components/DocumentView/Integration/ContentKitWithClientContext.tsx
```diff
@@ -0,0 +1,63 @@
...
+        () => ({
+            getVisitorContext: canAccessVisitorClaims
+                ? () => ({ visitor: visitorClaims?.visitor ?? null })
+                : undefined,
+            getPageContext: page ? () => ({ page }) : undefined,
+            navigate: ({ path, anchor }) => {
+                // Resolve the requested page path within the current space so a webframe can only
+                // navigate within the site, then soft-navigate client-side.
+                const target = joinPath(spaceBasePath, path) + (anchor ? `#${anchor}` : '');
+                router.push(target);
```
gregberge: Actually if only in the space it's not sufficient. It should be able to navigate to any section.

gregberge: Good catch — updated to resolve navigation against the site base path (`toPathInSite`) instead of the space, so a webframe can navigate to a page in any section or space of the site. Done in 31a53a71f.

packages/react-contentkit/src/context.ts
```diff
@@ -18,11 +18,31 @@ export type ContentKitRenderUpdate = Partial<
...
+
+    /**
+     * Client-only current-page context, merged into the webframe state.
+     */
+    getPageContext?: () =>
+        | Record<string, unknown>
+        | null
+        | undefined
+        | Promise<Record<string, unknown> | null | undefined>;
+
```
taranvohra: Why is it typed this way when we know the shape of `pageContext`?

packages/react-contentkit/src/context.ts
```diff
@@ -18,11 +18,31 @@ export type ContentKitRenderUpdate = Partial<
...
+    /**
+     * Navigate the host page to another page, in response to a webframe `@webframe.navigate`
+     * action. The host is responsible for resolving `path` within the current content and for
+     * restricting navigation to trusted destinations.
+     */
+    navigate?: (target: { path: string; anchor?: string }) => void;
```
taranvohra, line 36: Why `path` instead of `pageId`? I feel we would resolve the URL better if we know the pageId whereas path could be arbitrary no?

taranvohra, line 36: Or we support both path | pageId

packages/react-contentkit/src/context.ts
gregberge: Good point — typed it concretely now. Added a `ContentKitWebframePage` (`{ id, path, title }`) type and made `getPageContext` return `{ page: ContentKitWebframePage }` instead of `Record<string, unknown>`. Done in 3090c1c9c.

packages/react-contentkit/src/context.ts
gregberge: Agreed — `pageId` resolves more robustly than an arbitrary path. See the reply below: we now support both.

gregberge: Done in 3090c1c9c — the webframe can now address a destination by either. Split the client context into two functions:

- `navigateToPath` (sync): resolves the path against the site base path (`toPathInSite`), so it works across any section/space.
- `navigateToPageId` (async): resolves the ID against the site page tree via a server action (`resolvePageId` → `linker.toPathInSpace`), guaranteeing a real in-site page rather than trusting an arbitrary path.

`@webframe.navigate` prefers `pageId` when present and falls back to `path`. One limitation: `pageId` resolution is scoped to the current space's revision, so a page in a different section can only be reached by `path` for now.

packages/gitbook/src/components/DocumentView/Integration/server-actions.tsx
```diff
@@ -39,3 +40,23 @@ export async function renderIntegrationUi({
...
+/**
+ * Server action to resolve a page (by its ID) to a path within the current site, in response to a
+ * webframe `@webframe.navigate` action. Resolution is scoped to the current space's revision, so a
+ * page in another section/space can only be reached by `path`. Returns `null` when the page is
+ * unknown.
+ */
+export async function resolveWebframePagePath(pageId: string): Promise<string | null> {
+    return traceErrorOnly('DocumentView.resolveWebframePagePath', async () => {
+        const baseContext = await getServerActionBaseContext();
+        const context = await fetchServerActionSiteContext(baseContext);
```
taranvohra: `IntegrationBlock` has `contentContext`  which has the linker.  You're passing `siteBasePath` to `ContentKitWithClientContext`, pass the contentContext instead which can be passed to this function then. Then you don't need `async`

packages/gitbook/src/components/DocumentView/Integration/ContentKitWithClientContext.tsx
```diff
@@ -0,0 +1,86 @@
...
+ * Join a base path and a relative path, mirroring the server-side linker so navigation stays
+ * within the current site. Kept inline to remain client-safe.
+ */
+function joinPath(prefix: string, path: string): string {
+    const prefixPath = prefix.endsWith('/') ? prefix : `${prefix}/`;
+    const suffixPath = path.startsWith('/') ? path.slice(1) : path;
+    const joined = prefixPath + suffixPath;
+    const withoutTrailingSlash = joined.endsWith('/') ? joined.slice(0, -1) : joined;
+    return withoutTrailingSlash === '' ? '/' : withoutTrailingSlash;
+}
```
taranvohra: Please let's not keep repeating these utils everywhere

packages/gitbook/src/components/DocumentView/Integration/ContentKitWithClientContext.tsx
```diff
@@ -0,0 +1,86 @@
...
+        () => ({
+            getVisitorContext: canAccessVisitorClaims
+                ? () => ({ visitor: visitorClaims?.visitor ?? null })
+                : undefined,
+            getPageContext: page ? () => ({ page }) : undefined,
+            navigateToPath: ({ path, anchor }) => {
+                // Resolve the requested path relative to the site root so a webframe can navigate
+                // to any section or space within the site (and nowhere outside it).
+                const suffix = anchor ? `#${anchor}` : '';
+                navigateTo(joinPath(siteBasePath, path) + suffix);
```
taranvohra: Like I said, if you pass the `contentContext` instead of `siteBasePath` you can use the `linker.toPathInSite(path + suffix)`, it is better because the linker is properly tested

packages/gitbook/src/components/DocumentView/Integration/ContentKitWithClientContext.tsx
gregberge: Done — removed the duplicated `joinPath` helper. Navigation now resolves through the real `GitBookLinker` (`toPathInSite` / `toPathInSpace`), rebuilt on the client from serializable data. Fixed in 7f76ffdb6.

packages/gitbook/src/components/DocumentView/Integration/ContentKitWithClientContext.tsx
gregberge: Agreed, the linker is the right thing to use. Small caveat: `contentContext` itself can't be passed to this client component — its `linker`/`dataFetcher` are functions and don't survive the RSC serialization boundary. So instead I pass the serializable pieces (`linkerData` = base paths + host/protocol) and rebuild the tested linker on the client with `createLinker`, then use `linker.toPathInSite(...)`. Done in 7f76ffdb6.

packages/gitbook/src/components/DocumentView/Integration/server-actions.tsx
gregberge: Dropped the async server action. For `pageId` I now ship a compact `{ pageId: path }` map built server-side from the current revision (`getWebframePagePaths`), so `navigateToPageId` resolves synchronously on the client through `linker.toPathInSpace` — no round-trip and no site-context refetch. Same current-space scoping as before (cross-section still via `path`). Done in 7f76ffdb6.

packages/gitbook/src/components/DocumentView/Integration/adaptive.ts
```diff
@@ -1,26 +1,79 @@
...
+                pagePaths[page.id] = getPagePath(rootPages, page);
+            }
+            if ('pages' in page && page.pages.length > 0) {
+                walk(page.pages);
+            }
+        }
+    };
+    walk(rootPages);
+
+    return pagePaths;
```
taranvohra: Why we need the whole map upfront? It can slow down the page if the revision is massive.

taranvohra (approved):

packages/gitbook/src/components/DocumentView/Integration/adaptive.ts
gregberge: How do you want to navigate if we don't? It's a client-component, we can't pass the whole context to it if I am not wrong.

taranvohra: We don't pass the revision at all to the client component somewhere higher in the tree?

gregberge: I think we should just keep path, people don't even know what is a pageId, it's useless.

gregberge: No passing the revision would be even more costful.

taranvohra: Yeah I'm fine with it. We can add `pageId` later if they ask

PR #4377, opened 2026-07-08
Let visitors type a follow-up while the Assistant is answering (RND-11789)

zenoachtig:
## Proposed changes

RND-11789: while the Assistant streamed an answer, the whole chat input was disabled, so you couldn't begin composing a follow-up until it finished. Worse, a submit attempted mid-stream was silently dropped by the controller (`onPostMessage` returns early when `responding`) — but only *after* the `Input` primitive had already cleared the typed text, so the follow-up you started was lost.

This makes the field stay editable during streaming while still blocking a premature send:

- **`Input` primitive** — new `submitDisabled` prop. When set, the field remains editable but Enter and the submit button are inert and the typed value is **preserved** (`handleSubmit` returns early before clearing). This is deliberately distinct from `disabled`, which also makes the field read-only.
- **`AIChatInput`** — now only fully disables on `disabled` (error) or an active control; it passes `submitDisabled={responding}` and keeps `aria-busy` for assistive tech. So you can type during a stream and send the moment it completes.
- **`AIChat`** — the call site no longer folds `responding` into `disabled` (submit-blocking now flows through the `responding`/`submitDisabled` path).

No change to concurrent-turn safety: `onPostMessage` already ignores a submit while `responding`; this just stops the UI from disabling the field and discarding text in that window.

## Test gap

Behavioural UI tweak validated by reading the flow; I did not add a Playwright test in this draft (no emulator run in this session). Happy to add one covering "type + send during/after a stream" if you'd like before it leaves draft.

## Changelog

- [Fix] Assistant: you can now start typing a follow-up question while an answer is still being written, instead of waiting for it to finish.

*— Drafted by the night shift for @zenoachtig to review.*

---
_Generated by [Claude Code](https://claude.ai/code/session_01CDRwrYgp4VpL3NVth3RzDj)_

### DIFF
--- a/packages/gitbook/src/components/AI/useAIChat.tsx
+++ b/packages/gitbook/src/components/AI/useAIChat.tsx
@@ -125,11 +125,12 @@ export type AIChatState = {
     draft: string;
 
     /**
-     * A follow-up the visitor submitted while a previous turn was still streaming. It is held
-     * here and sent automatically once the current answer finishes (see the flush in
-     * `streamResponse`), so submitting mid-stream isn't lost. `null` when nothing is queued.
+     * Follow-ups the visitor submitted while a previous turn was still streaming. They are held
+     * here and sent automatically, one at a time in submission order, as each answer finishes
+     * (see the flush in `streamResponse`), so submitting mid-stream isn't lost. Empty when
+     * nothing is queued.
      */
-    queuedMessage: string | null;
+    queuedMessages: string[];
 };
 
 export type AIChatEvent =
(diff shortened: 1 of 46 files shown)


nolannbiron:
@zenoachtig isn't there a way to queue messages ? Codex has a pending-message queue, and Claude Code lets you send while it's working and picks it up after, so submit wouldn't need to be disabled

zenoachtig:
We could build it of course

zenoachtig:
@nolannbiron Implemented a cancelable message queue now, check it out 😊

packages/gitbook/src/components/AI/useAIChat.tsx
```diff
@@ -512,6 +528,19 @@ export function AIChatProvider(props: {
                         loading: false,
                         error: false,
                     }));
+
+                    // The turn is fully settled: send the next follow-up the visitor queued while
+                    // it was streaming (oldest first). Skipped when a control is awaiting input
+                    // (posting would throw and the input is disabled anyway) — they stay queued and
+                    // flush after the control continuation ends. Sending one message starts a new
+                    // turn whose own completion flushes the following one, draining the queue in
+                    // order.
```
nolannbiron: we should add something in CLAUDE.md to prevent those big comments, it adds nothing and it's unreadable

nolannbiron (approved):

PR #4379, opened 2026-07-08
Separate prompt block actions

nolannbiron:
Splits the prompt block header actions into two distinct buttons instead of a single combined button group:

- **Primary "Open in" dropdown** — opens the prompt in Claude, ChatGPT, or Cursor.
- **Secondary "Copy prompt" button**.

Also adds a standalone `open_in_ai` ("Open in") translation key across all locales (the existing `open_in` key requires a provider argument, so it can't be reused for the generic dropdown label).

Resolves RND-11847.

Demo:
CleanShot 2026-07-08 at 12 17 24@2x

------------------------------------------------------------------------------------

CleanShot 2026-07-08 at 12 17 30@2x

### DIFF
--- a/packages/gitbook/src/components/DocumentView/Prompt/PromptClient.tsx
+++ b/packages/gitbook/src/components/DocumentView/Prompt/PromptClient.tsx
@@ -1,12 +1,6 @@
 'use client';
 
-import {
-    Button,
-    ButtonGroup,
-    DropdownMenu,
-    DropdownMenuItem,
-    ToggleChevron,
-} from '@/components/primitives';
+import { Button, DropdownMenu, DropdownMenuItem, ToggleChevron } from '@/components/primitives';
 import { getURLForLLM } from '@/components/utils';
 import { tString, useLanguage } from '@/intl/client';
 import { tcls } from '@/lib/tailwind';
@@ -29,7 +23,7 @@ export function PromptClient(props: {
     const [headerHasFocus, setHeaderHasFocus] = React.useState(false);
     return (
         <>
-            <div className="group/prompt-header relative flex min-h-9 flex-row items-center justify-between gap-4 px-3 py-2">
+            <div className="group/prompt-header relative flex min-h-9 flex-row items-center justify-between gap-4 p-3">
                 <button
                     type="button"
                     aria-controls={promptId}
@@ -109,10 +103,10 @@ function PromptActions(props: { prompt: string; openInAIProviders: boolean }) {
     const { prompt, openInAIProviders } = props;
 
     return (
-        <ButtonGroup className="relative z-20 shrink-0 overflow-visible">
+        <div className="relative z-20 flex shrink-0 items-center gap-2">
             <CopyPromptButton prompt={prompt} />
             {openInAIProviders ? <OpenPromptDropdown prompt={prompt} /> : null}
-        </ButtonGroup>
+        </div>
     );
 }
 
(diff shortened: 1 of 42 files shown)


zenoachtig (approved):

packages/gitbook/src/components/DocumentView/Prompt/PromptClient.tsx
```diff
@@ -164,12 +128,10 @@ function OpenPromptDropdown(props: { prompt: string }) {
             className="!min-w-48 max-w-max"
             button={
                 <Button
-                    icon={<ToggleChevron className="size-text-sm" />}
-                    label={tString(language, 'open')}
-                    iconOnly
+                    label={tString(language, 'open_in_ai')}
```
gregberge, line 131: Put the "open_in" action as secondary and the "Copy prompt" as primary. "Open in" is mostly useless because people are using agent code.

gregberge (approved):

packages/gitbook/src/components/DocumentView/Prompt/PromptClient.tsx
nolannbiron: we'll ship it like that and add options to open in claude code/codex later, `Open in` should be the main since it's a shortcut and doesn't require extra steps

gregberge: It's not the main action. I don't want to click on this, neither do you. Trust your own instinct and usage!

nolannbiron: idk honestly, I get your point but if the main action was "Open in Claude code" I would click on that, when you're using this block you're almost always starting a new conversation, and we'll add this option + save the last selected action so it appears first instead of "Open in"

gregberge: I am using Claude Code in VSCode, others in VSCode Agents, some others in other software, others in Terminal. I mean the "Open in" is quite useless for a prompt like this because it's not a prompt you will paste in a ChatGPT or Claude, it's a prompt for coding agents most of the time.

nolannbiron: yeah yeah I know, but I'll add VSCode, claude code and codex in a follow-up PR (today), we can't assume that everyone is using them inside VSCode extensions like we do, more and more people are using Claude Code in the desktop app, and for me it's always better to push for a 1-click option instead of "here is the content do what you want with it", this option is still there anyway

nolannbiron: I can add an event for a week and we decide after also 🤷🏻‍♂️😅

gregberge: Lol if you put the button as primary your event is tricked. A click does not mean the user is satisfied with the result.

nolannbiron: I know I'm not going to put an event on something so trivial, anyway, the `Open in` should be complete like that, it's even possible to open Claude code in VSCode with `vscode://anthropic.claude-code/open?prompt={prompt}`

nolannbiron: I won't fight about it cause I don't really mind, I've swapped the variants for now (copy is the primary), but we should explore the complete menu + saved last-action approach later

gregberge: Yeah do as you want. It was just my feeling.

PR #4380, opened 2026-07-08
Apply and remember the `?theme=` light/dark override in the docs embed

zenoachtig:
## Overview

- Fix the docs embed ignoring `?theme=light`/`?theme=dark` (RND-11571).
- The embed root redirect now forwards the query string to the default tab, so the theme (and other params) aren't lost on the hop to `/assistant` or `/page/`.
- Removed `export const dynamic = 'force-static'` from the dynamic-tree `assistant` and `search` pages, so the layout can actually read the forced theme (and customization overrides) from request headers instead of rendering statically with empty `headers()`.
- Persist the resolved forced theme to the embed's own storage (`gitbook-theme-embed:<siteId>`), so it's remembered across tab navigation and reloads instead of only applying while the query is present. An explicit `?theme=` on a later load still overrides the remembered one; single-theme sites (which intentionally ignore the override) don't persist it.
- Note: following the embedding iframe's CSS `color-scheme` is intentionally **not** included — the embed's dark styling is class-driven (Tailwind `dark:`, `.dark`-gated rules) and CSS has no selector for an element's *used* color-scheme, so it can't be done CSS-natively without reworking the design-system dark variant. `?theme=` is the supported control.

## Demo

_Verifone-style embed on a dark-mode device — before vs after `?theme=light`:_

_Theme remembered across tab navigation (no `?theme=` on the second tab):_

*— Authored by Claude*


packages/gitbook/src/components/Embeddable/EmbeddableRootLayout.tsx
```diff
@@ -50,6 +51,9 @@ export async function EmbeddableRootLayout({
                 contextId={context.contextId}
                 proxyOrigin={context.site.proxy?.origin}
             >
+                {/* Persist the resolved theme (not the raw ?theme=) so it survives tab navigation
+                    without single-theme sites persisting an override they ignore. RND-11571 */}
+                <EmbeddableThemeSync forcedTheme={theme.forcedTheme} />
```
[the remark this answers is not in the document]
zenoachtig: Fixed in 87a2b48 — persistence is now gated on an explicit URL `?theme=` **and** a toggleable site (`toggeable ? forcedTheme : null`), so single-theme sites no longer persist their default (or any ignored override) on plain visits.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/gitbook/src/app/sites/dynamic/[mode]/[siteURL]/[siteData]/~gitbook/embed/search/page.tsx
```diff
@@ -6,8 +6,6 @@ type PageProps = {
     params: Promise<RouteLayoutParams>;
 };
 
-export const dynamic = 'force-static';
-
```
Comment (nolannbiron) [inline packages/gitbook/src/app/sites/dynamic/[mode]/[siteURL]/[siteData]/~gitbook/embed/search/page.tsx:10]: are we sure about that ? seems dangerous to change this

Comment (zenoachtig) [reply packages/gitbook/src/app/sites/dynamic/[mode]/[siteURL]/[siteData]/~gitbook/embed/search/page.tsx:10]: Good instinct to double-check — it's safe, and it's load-bearing for this fix:

- This is only the **dynamic-tree** copy. The middleware routes here only for genuinely dynamic requests (`?theme=`, a customization override, adaptive/VA/`contextId`, or the `gitbook-dynamic-route` cookie). The normal cacheable case is still served by the **static tree**, which keeps `force-static` (untouched) — so nothing that was cached before stops being cached.
- Under `force-static`, `headers()` is empty, so the layout's `getThemeFromMiddleware()` (and the `x-gitbook-customization` override) always resolved to null — the page silently dropped `?theme=` and customization overrides. That's the RND-11571 bug; removing it lets those header-driven requests render correctly.
- Those requests are effectively uncacheable anyway (`contextId` responses already carry `max-age=0, must-revalidate`), so `force-static` gave no caching benefit here — only wrong output.
- Precedent: the docs tab (`page/[pagePath]`) in this same dynamic tree already has no `force-static`; this just makes `assistant`/`search` consistent with it.

If you'd still rather keep `force-static`, the only alternative is applying the theme purely client-side, which reintroduces a first-paint flash and still wouldn't fix the dropped customization override — so I'd recommend against it. Happy to talk it through.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Comment (zenoachtig) [reply packages/gitbook/src/app/sites/dynamic/[mode]/[siteURL]/[siteData]/~gitbook/embed/search/page.tsx:10]: Zeno the human here, I'm also a bit concerned so I'll leave it out of this PR and seeing if the first-paint is a big problem or not. If we get reports on it we can reconsider, but I'd rather not touch it.

Comment (nolannbiron) [reply packages/gitbook/src/app/sites/dynamic/[mode]/[siteURL]/[siteData]/~gitbook/embed/search/page.tsx:10]: hello Zeno the human 👋🏼, I'm checking on my side the path claude is mentioning

Comment (nolannbiron) [reply packages/gitbook/src/app/sites/dynamic/[mode]/[siteURL]/[siteData]/~gitbook/embed/search/page.tsx:10]: @zenoachtig looks like it's safe, maybe @nicolas you can confirm ?

conico974 (changes requested):
This is not how we should do it. Those route should never be dynamic

packages/gitbook/src/app/sites/dynamic/[mode]/[siteURL]/[siteData]/~gitbook/embed/search/page.tsx
Comment (zenoachtig) [reply packages/gitbook/src/app/sites/dynamic/[mode]/[siteURL]/[siteData]/~gitbook/embed/search/page.tsx:10]: Reworked per @conico974's suggestion — the routes stay `force-static`. Instead of making them dynamic, the middleware now threads the forced theme into the **embed route context** (a new `embedTheme` on the encoded site data), scoped to the embed route only (the main site keeps its existing dynamic header path). The embed layouts read it from the route params — which works under static rendering — and the redirect forwards it to the default tab. The theme becomes part of the route's static cache key, so each theme is its own cached entry. `force-static` is restored on `assistant`/`search`. Pushed in 6cc98add1.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/gitbook/src/lib/context.ts
```diff
@@ -82,6 +83,13 @@ export type SiteURLData = Pick<
      * Defaults to true when undefined.
      */
     displayAgentInstructions?: boolean;
+
+    /**
+     * Theme forced via `?theme=` for the docs embed only. Passed through the route context (rather
+     * than a request header) so the embed routes can honor it while staying statically rendered.
+     * Not set for the main site. RND-11571.
```
conico974:
```suggestion
     * Should never be set for the main site. RND-11571.
```

conico974 (approved):

PR #4391, opened 2026-07-13
Fix table-of-contents page group's first item cut off after navigation (RND-11314)

zenoachtig:
Linear issue: https://linear.app/gitbook-x/issue/RND-11314

## Overview

- The first item of a table-of-contents **page group** in the sidebar could render **cut off** — its top faded out under the group header — most reliably after a client-side navigation (e.g. hitting a 404 route, then clicking the header logo to go home).
- The layout is correct; it's a **stale compositing paint**, which is why any DOM change makes the item snap back to normal.

## Root cause

The collapsible-groups work (#4177) drives each sticky group header's fade gradient (an `::after` band at `top-full h-4`, overlapping the group's first item) from an `IntersectionObserver` on a sentinel. It marked the header **stuck** whenever the sentinel stopped intersecting — but that also happens when:

- the group is simply **below the fold**, and
- **transiently during a client-side navigation**, before the sidebar's scroll container has laid out.

So the fade gradient switched **on** for groups that aren't actually pinned, and the compositor could leave it painted over the first item as a stale layer. It's always the *first* item because that's exactly what the gradient overlaps.

## Fix

Mark the header stuck only once the sentinel has actually scrolled **above** the scroll area's top, instead of on any non-intersection:

```js
const rootTop = entry.rootBounds?.top ?? 0;
setIsSticking(!entry.isIntersecting && entry.boundingClientRect.top < rootTop);
```

## Verification

Verified against the live affected site by attaching a real `IntersectionObserver` with the new condition:

- **At scroll 0** (nothing pinned): the old logic flags every below-fold group as stuck (fade on — the bug); the new logic reports **all groups not-stuck** (fade off).
- **Scrolled so headers pin**: genuinely-pinned headers still report stuck under the new logic (fade still works — no regression); below-fold groups correctly report not-stuck.

Typecheck and Biome format are clean on the change.

### Manual check

1. Visit a 404 route on a docs site with page groups (e.g. `…/asdf`).
2. Click the header logo to return home.
3. The first item of each group (e.g. AI Features → Overview) should render fully — no faded/clipped top.
4. Scroll down and confirm items still fade softly as a group header pins to the top.

## Changelog

- [Fix] The first item of a sidebar page group no longer appears cut off (faded under the group header) after navigating between pages.

*— Authored by Claude*


packages/gitbook/src/components/TableOfContents/PageGroupItem.tsx
```diff
@@ -125,9 +125,15 @@ function useIsSticking() {
...
-                if (entry) {
-                    setIsSticking(!entry.isIntersecting);
-                }
+                if (!entry) return;
+                // The sentinel also stops intersecting when the group is simply below the
+                // fold, or transiently during a client-side navigation before layout
+                // settles — not only when the header pins to the top. Treating those as
+                // "stuck" turns on the header's fade gradient, which overlaps the group's
+                // first item and can be left painted over it as a stale layer. Only mark it
+                // stuck once the sentinel has actually scrolled above the scroll area's top.
```
BrettJephson: This comment is noise and reads like a word soup.

BrettJephson (approved):

packages/gitbook/src/components/TableOfContents/PageGroupItem.tsx
zenoachtig: @copilot remove this comment, it's not needed here

PR #4395, opened 2026-07-13
Split the default-scope site search into two parallel API requests

utkuufuk:
Split the default-scope site search into two parallel API requests — one restricted to the current site space and one for the other site spaces — rendering each result set as soon as its response arrives. All results are ranked together by score, with the current site space scores boosted.


packages/gitbook/src/components/Search/reciprocalRankFusion.ts
```diff
@@ -151,19 +151,26 @@ function mergePinnedRemoteResult(
  * Additionally, a Jaro-Winkler title-match bonus is applied per query word:
  * for each query word that appears as a substring in the title, the best
  * Jaro-Winkler similarity score against any individual title word is added.
+ *
+ * Results from the other site spaces (`otherRemoteResults`) are appended at the
```
conico974: I'm actually kind of against doing it like that.
This is a fundamental change from what we had before, and there is ton of cases where you'd want result from other space to be higher in the list. With that you may have to scroll a lot before getting the result you want.

packages/gitbook/src/components/Search/useSearchResults.ts
```diff
@@ -170,68 +197,153 @@ export function useSearchResults(props: {
...
                         default:
                             assertNever(scope);
                     }
                 })();
 
+                let resultsArrived = false;
+                otherSpacesResultsPromise?.then(
+                    () => {
+                        if (!resultsArrived && !cancelled) {
+                            console.warn(
```
conico974: Why the log ?
Also this would make for a very poor user experience, especially if this happens with very few (or none) local result. The entire list would shift

packages/gitbook/src/components/Search/reciprocalRankFusion.ts
utkuufuk: @conico974 I wanted to prevent the results from jumping for one extra time, but I do agree with your point. I updated the PR such that we merge the current & other spaces results and feed them together to RRF.

packages/gitbook/src/components/Search/useSearchResults.ts
```diff
@@ -28,6 +28,9 @@ export type ResultType =
 
 export type { LocalPageResult, MergedPageResult };
 
+// Score multiplier for current site space results when combined with those from other site spaces
+const CURRENT_SITE_SPACE_SCORE_MULTIPLIER = 2;
```
conico974, line 32: We'll probably need to tweak this one at some point depending on feedback

conico974 (approved):

PR #4405, opened 2026-07-16
Add Sites OAuth consent screen

spastorelli:
This PR adds support for the sites OAuth consent flow in GBO by introducing a route that renders the consent screen, showing information about the MCP client requesting access to the site's MCP server.

The corresponding changes to the sites OAuth server are implemented in https://github.com/GitbookIO/gitbook-x/pull/24034.

As with that PR, the consent form is gated behind a feature flag so we can iterate on the screen before rolling it out to all customers.

### DIFF
--- a/packages/gitbook/src/components/SiteOAuthConsent/ConsentForm.tsx
+++ b/packages/gitbook/src/components/SiteOAuthConsent/ConsentForm.tsx
@@ -1,0 +1,93 @@
+'use client';
+
+import { useState, useTransition } from 'react';
+
+import { Button } from '@/components/primitives/Button';
+import { Checkbox } from '@/components/primitives/Checkbox';
+import { tcls } from '@/lib/tailwind';
+
+import { type SubmitConsentInput, submitSiteOAuthConsent } from './actions';
+
+/**
+ * Site's OAuth consent form to present to the user the client's information requesting access to the site's MCP.
+ */
+export function ConsentForm(props: {
+    siteId: string;
+    consentSessionId: string;
+    /** Whether the OAuth server recognizes the client as verified. */
+    verified: boolean;
+}) {
+    const { siteId, consentSessionId, verified } = props;
+    const [isPending, startTransition] = useTransition();
+    const [error, setError] = useState<string>();
+    const [trusted, setTrusted] = useState(false);
+
+    // Unverified clients can only be approved once the visitor explicitly acknowledges they trust
+    // the app. The OAuth server re-checks this, so it can't be bypassed by tampering with the client.
+    const canApprove = verified || trusted;
+
+    const decide = (decision: SubmitConsentInput['decision']) => {
+        setError(undefined);
+        startTransition(async () => {
+            const result = await submitSiteOAuthConsent({
+                siteId,
+                consentSessionId,
+                decision,
+                trusted,
+            });
+            if ('redirectURL' in result) {
+                // Full-page navigation to the client's (external) redirect URI.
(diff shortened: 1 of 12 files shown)


packages/gitbook/src/lib/env/globals.ts
```diff
@@ -132,6 +132,14 @@ export const GITBOOK_ICONS_TOKEN = process.env.GITBOOK_ICONS_TOKEN;
  */
 export const GITBOOK_SECRET = process.env.GITBOOK_SECRET ?? null;
 
+/**
+ * Shared secret used to sign server-to-server requests to the sites OAuth server consent endpoints.
+ * This must match the sites OAuth provider signing secret (`functionsConfig.sitesOAuth.signingSecret`
+ * in gitbook-x); it is a dedicated secret and must not be confused with `GITBOOK_SECRET`.
+ */
+export const GITBOOK_SITE_OAUTH_SIGNING_SECRET =
+    process.env.GITBOOK_SITE_OAUTH_SIGNING_SECRET ?? null;
```
conico974, line 141: Did you forgot to add them here https://github.com/GitbookIO/gitbook/blob/main/.github/composite/deploy-cloudflare/action.yaml
Same for the vercel one

packages/gitbook/src/lib/site-oauth/flag.ts
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * Kept in its own module (free of `node:crypto`/`server-only`) so it can be imported from the edge
+ * middleware. It is never imported into a client bundle.
+ */
+export function isSitesOAuthConsentEnabled(): boolean {
+    const override = process.env.GITBOOK_SITE_OAUTH_CONSENT_ENABLED;
+    if (override !== undefined) {
+        return override === 'true';
+    }
+    return process.env.NODE_ENV !== 'production';
```
conico974: `NODE_ENV` is always equal to production in all env actually

packages/gitbook/src/lib/env/globals.ts
spastorelli: Good catch, thanks. Adding them.

packages/gitbook/src/lib/site-oauth/flag.ts
spastorelli: ha sorry missed that. Changed the approach slightly. Should still be `!== 'production'` when running locally right?

conico974: It depends on everyone really, I know that in some cases I set it to production. But I don't think that's an issue really, everyone should be responsible for that

spastorelli: Ok yeah plus it's fine the ones that have `NODE_ENV === 'production'` will get the old/current flow not tthe new one. Also this flag will be removed once we finished iterating on the consent flow.

packages/gitbook/src/components/SiteOAuthConsent/ConsentForm.tsx
```diff
@@ -0,0 +1,93 @@
...
+        startTransition(async () => {
+            const result = await submitSiteOAuthConsent({
+                siteId,
+                consentSessionId,
+                decision,
+                trusted,
+            });
+            if ('redirectURL' in result) {
+                // Full-page navigation to the client's (external) redirect URI.
+                window.location.href = result.redirectURL;
```
conico974, line 40: Why not redirect in the server action directly ? Not a big deal though

conico974 (approved):

PR #4407, opened 2026-07-16
Defer search index loading until search is opened

nolannbiron:


packages/gitbook/src/components/Search/useLocalSearchResults.tsx
```diff
@@ -110,7 +113,9 @@ async function getOrBuildIndexes(indexURL: string): Promise<Map<string, Document
     }
 
     pendingFetch = (async () => {
-        const response = await fetch(indexURL);
+        // Load FlexSearch lazily so its code lands in an on-demand chunk instead of the
+        // main client bundle — it's only needed once the user actually searches.
+        const [{ Document }, response] = await Promise.all([import('flexsearch'), fetch(indexURL)]);
```
conico974: I'm not sure we should do it like that. It introduce a significant delay before loading the index.
I think we should load it as early as possible and store it in some kind of global state.
Loading the rest async should be fine, but not the index itself. We want this to be as fast as possible for the user.

conico974 (approved):

PR #4414, opened 2026-07-17
Fix ScrollContainer scroll buttons not reflecting overflow immediately

tgargula:
## Summary

Extracted scroll overflow tracking into a custom hook (`useScrollOverflow`) to improve code reuse and ensure overflow state is detected immediately when content changes dynamically. This fixes the "Scroll further" button in search results not appearing right away for overflowing content.

## Changes

* Created new `useScrollOverflow` hook to centralize scroll position and overflow tracking
* Updated `ScrollContainer` to use the new hook
* Added test ID to `ScrollContainer` for improved e2e test selectors
* Enhanced e2e tests to verify the scroll button reflects overflow immediately without manual scrolling

## Demo

Before

https://github.com/user-attachments/assets/7b5cdc74-f01a-4c1f-8c4b-c26eff666fd2

After

https://github.com/user-attachments/assets/04cc527e-1ea5-44d4-9590-0421bb442c10

E2E tests

image

🤖 Generated with Claude Code

resolves RND-11974


packages/gitbook/e2e/internal.spec.ts
```diff
@@ -274,6 +274,85 @@ const searchTestCases: Test[] = [
         // Re-applied per viewport so the replacement survives resize-driven re-renders.
         normalizeBeforeScreenshot: (page) => page.evaluate(overrideAIResponse),
     },
+    {
+        name: 'Search - Results scroll button reflects overflow immediately',
```
tgargula: I think the E2E are on the edge between necessary and "too much", so I'm also good if you prefer to remove them. Your call!

packages/gitbook/src/components/hooks/useScrollOverflow.ts
```diff
@@ -0,0 +1,96 @@
...
+        );
+    }, [orientation, containerRef]);
+
+    useScrollListener(measure, containerRef);
+
+    // Measure synchronously after mount and whenever the rendered children change,
+    // so overflow is detected immediately instead of waiting for a resize/scroll event.
+    React.useLayoutEffect(() => {
+        measure();
+    }, [measure, children]);
```
gregberge: Hmm passing children like this is a bit dangerous IMO!

packages/gitbook/src/components/hooks/useScrollOverflow.ts
```diff
@@ -0,0 +1,96 @@
...
+            if (frame !== null) {
+                return;
+            }
+            frame = requestAnimationFrame(() => {
+                frame = null;
+                measure();
+            });
+        };
+
+        const ro = new ResizeObserver(scheduleMeasure);
```
gregberge, line 64: I guess we should wrap the logic of resize observer in a dedicated hook.

gregberge (approved):

packages/gitbook/src/components/hooks/useScrollOverflow.ts
tgargula: that's why I created useScrollOverflow. Do you want me to go more granularly?

gregberge: Oh I didn't see it, it's fine very good.

packages/gitbook/e2e/internal.spec.ts
gregberge: Yeah it does not worth a E2E, if you can just add a screenshot that has it but honestly it does not even worth it. Also your test is a bit complex. So I think it's better to remove it.

packages/gitbook/src/components/hooks/useScrollOverflow.ts
tgargula: yeah, at this point it could go on every render, so I can drop the deps array. That actually sounds scary but in this case that's what we want anyway

gregberge: IDK but here it's not controlled as you said it can go on every render.

gregberge: So if higher up in the tree we have something that renders a lot, we are bloated. We can't do this.

packages/gitbook/src/components/hooks/useScrollOverflow.ts
```diff
@@ -0,0 +1,95 @@
...
+            orientation === 'horizontal' ? container.scrollLeft : container.scrollTop
+        );
+    }, [orientation, containerRef]);
+
+    useScrollListener(measure, containerRef);
+
+    // Measure synchronously after every render, so overflow is detected
+    // immediately instead of waiting for a resize/scroll event.
+    React.useLayoutEffect(() => {
+        measure();
```
gregberge, line 41: Not good, in this case put this in a requestAnimationFrame. You can't do it like this. You have to put a guard on it.

tgargula: there are two times in which we need to measure:
- React DOM changes
- Resize/Mutations change

The first part `measure()` needs to run syncronously before the paint (that's why the useLayoutEffect vs useEffect). If we wrap it into reqestAnimationFrame (rAF) that would be deferred causing a flush if a cursor is hovering the search container. This needs to run whenever any child changes, namely on every render. That's why no deps, which is unusual. The side effect inside `measure` does not affect the React state (setScrollPosition, setScrollSize hold primitives, so they don't cause rerenders), so no infinite loop. What I can do is to add some additional comments saying that scrollPosition and scrollSize cannot be used to affect the DOM value.

The second part needs to run on every resize/mutation change and here the rAF is used.

In this case, the reference is stable. (added a `console.log('repainting');` in `React.useLayoutEffect`). Runs on every render, does not cause infinite loops.

https://github.com/user-attachments/assets/6025743c-6d7d-42a6-98a5-8060e109e198

packages/gitbook/src/components/hooks/useScrollOverflow.ts
tgargula: added `measure` as the only dependency back. If there's anything not working correctly, I could revisit later but the current code is an improvement anyway.

packages/gitbook/e2e/internal.spec.ts
tgargula: removed

PR #4418, opened 2026-07-20
Make the table search empty state more prominent

zenoachtig:
## Overview

- The table search "no results" empty state was easy to miss on text-heavy pages — with no spacing around it, the message read as part of the table/content below it.
- Adds **2rem of vertical space** above and below the empty-state message (`py-8`) so it stands out clearly as its own element.
- Applies to both grid and card table views, since the empty state is shared.

## Demo

The change is a single additive spacing class on the shared empty-state element. It only appears client-side after a user searches a table (≥7 rows) or filters cards with no matches, so it couldn't be exercised through the local `/url/` dev proxy in this environment. No before/after screenshot is attached for that reason.

Resolves RND-11987.

*— Authored by Claude*

### DIFF
--- a/.changeset/table-search-empty-padding.md
+++ b/.changeset/table-search-empty-padding.md
@@ -1,0 +1,5 @@
+---
+"gitbook": patch
+---
+
+Make the table search "no results" empty state more prominent with vertical spacing so it no longer blends into the content below.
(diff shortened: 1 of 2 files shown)


BrettJephson (commented):
lgtm

BrettJephson (approved):

PR #4421, opened 2026-07-20
Implement API token cookie handling with chunking support

conico974:
Introduce functionality for handling API tokens in cookies, allowing for chunking of oversized tokens. 
This includes methods for reading, writing, and reconstructing tokens from cookies, ensuring compatibility with single-cookie tokens and managing expiration of unnecessary chunks.

### DIFF
--- a/packages/gitbook/src/lib/api-token-cookie.test.ts
+++ b/packages/gitbook/src/lib/api-token-cookie.test.ts
@@ -1,0 +1,90 @@
+import { describe, expect, it } from 'bun:test';
+
+import { getAPITokenFromCookies, getAPITokenResponseCookies } from './api-token-cookie';
+
+const cookieName = 'gitbook-api-token~test';
+const options = {
+    httpOnly: true,
+    sameSite: 'none' as const,
+    secure: true,
+    maxAge: 60 * 60,
+};
+
+describe('API token cookies', () => {
+    it('keeps tokens at the cookie limit in one cookie', () => {
+        const apiToken = 'a'.repeat(4_000);
+
+        expect(getAPITokenResponseCookies({ cookies: [], cookieName, apiToken, options })).toEqual([
+            { name: cookieName, value: apiToken, options },
+        ]);
+    });
+
+    it('splits and reconstructs oversized tokens', () => {
+        const apiToken = `${'a'.repeat(4_000)}b`;
+        const cookies = getAPITokenResponseCookies({ cookies: [], cookieName, apiToken, options });
+
+        expect(cookies).toEqual([
+            { name: cookieName, value: '2', options },
+            { name: `${cookieName}-0`, value: 'a'.repeat(4_000), options },
+            { name: `${cookieName}-1`, value: 'b', options },
+        ]);
+        expect(getAPITokenFromCookies(cookies, cookieName)).toBe(apiToken);
+    });
+
+    it('reads legacy single-cookie tokens', () => {
+        expect(
+            getAPITokenFromCookies([{ name: cookieName, value: 'legacy-token' }], cookieName)
+        ).toBe('legacy-token');
+    });
+
+    it('does not return a partial token when a chunk is missing', () => {
+        expect(
+            getAPITokenFromCookies(
+                [
+                    { name: cookieName, value: '2' },
+                    { name: `${cookieName}-0`, value: 'first' },
+                ],
+                cookieName
+            )
+        ).toBeUndefined();
+    });
+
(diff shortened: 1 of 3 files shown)


packages/gitbook/src/lib/api-token-cookie.ts
```diff
@@ -0,0 +1,116 @@
+import type { ResponseCookie, ResponseCookies } from './visitors';
+
+const COOKIE_CHUNK_SIZE = 4_000;
+
+type RequestCookie = Pick<ResponseCookie, 'name' | 'value'>;
+
+export function getAPITokenFromCookies(
```
taranvohra, line 15: Let's add a comment why we need to do this

packages/gitbook/src/lib/api-token-cookie.ts
```diff
@@ -0,0 +1,116 @@
...
+    return chunks;
+}
+
+function getPreviousChunkCount(cookies: readonly RequestCookie[], cookieName: string): number {
+    const cookie = cookies.find(({ name }) => name === cookieName);
+    return cookie ? (parseChunkCount(cookie.value) ?? 0) : 0;
+}
+
+function parseChunkCount(value: string): number | undefined {
+    if (!/^(?:[2-9]|[1-9]\d+)$/.test(value)) {
```
taranvohra: What is this regex? can we simplify using like `!isNaN(Number(value))`?

packages/gitbook/src/lib/api-token-cookie.ts
```diff
@@ -0,0 +1,116 @@
...
+        const index = Number(indexValue);
+        if (/^\d+$/.test(indexValue) && Number.isSafeInteger(index) && index >= 0) {
+            chunks.set(index, value);
+        }
+    }
+
+    // A numeric legacy token remains a single cookie until at least one chunk identifies it as chunked.
+    if (chunks.size === 0) {
+        return cookie.value;
+    }
```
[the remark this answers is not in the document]
conico974: That's expected, it will fail if it is not correct anyway. There is no "risk" in doing that and it's backward compatible

packages/gitbook/src/lib/api-token-cookie.test.ts
```diff
@@ -0,0 +1,90 @@
...
+                cookies: oldCookies,
+                cookieName,
+                apiToken: 'replacement',
+                options,
+            })
+        ).toEqual([
+            { name: cookieName, value: 'replacement', options },
+            { name: `${cookieName}-0`, value: '', options: { ...options, maxAge: 0 } },
+            { name: `${cookieName}-1`, value: '', options: { ...options, maxAge: 0 } },
+            { name: `${cookieName}-2`, value: '', options: { ...options, maxAge: 0 } },
```
taranvohra, line 114: Does this mean we will treat `''` as a sign to remove them ?

packages/gitbook/src/lib/api-token-cookie.test.ts
```diff
@@ -0,0 +1,90 @@
...
+            { name: cookieName, value: apiToken, options },
+        ]);
+    });
+
+    it('splits and reconstructs oversized tokens', () => {
+        const apiToken = `${'a'.repeat(4_000)}b`;
+        const cookies = getAPITokenResponseCookies({ cookies: [], cookieName, apiToken, options });
+
+        expect(cookies).toEqual([
+            { name: cookieName, value: '2', options },
```
taranvohra: Just a thought but do you think we should store it as ` { chunks: '2' }` as an object is more extensible for any future needs

taranvohra (approved):
Few comments but looks good otherwise 👍

packages/gitbook/src/lib/api-token-cookie.ts
conico974: It's to get the chunk count, I'll change the format so this will change to something else.
I'll add a comment to explain what it is

packages/gitbook/src/lib/api-token-cookie.test.ts
conico974: Yeah an empty chunks is useless anyway

packages/gitbook/src/lib/api-token-cookie.test.ts
conico974: I'd like to avoid a JSON.parse here, I think I'll go with the suggestion from copilot https://github.com/GitbookIO/gitbook/pull/4421#discussion_r3614362451 that we can also extend later if needed

emmerich (approved):

PR #4433, opened 2026-07-22
Serve noindex on internal search/assistant URLs (?q= / ?ask=)

nolannbiron:
## Problem

On published sites, the `?ask=` (AI Assistant) and `?q=` (search) URLs were showing up as
**"Indexed, though blocked by robots.txt"** in Google Search Console. That status occurs when
Google discovers a URL via links, is blocked from crawling it by robots.txt, and indexes the
URL anyway (URL-only). A robots.txt block prevents crawling but not indexing — and it also
prevents Google from ever seeing an on-page `noindex`.

## Fix

- Serve `X-Robots-Tag: noindex` on any request carrying `?ask=` or `?q=`, set in middleware
  per-request so the cached page body is unaffected.
- Remove the `Disallow: /*?*q=*` / `Disallow: /*?*ask=*` rules from robots.txt so Google can
  crawl these URLs, read the `noindex`, and drop them cleanly.

The Assistant and search features are unchanged — the URLs still work client-side for readers.
Existing GSC entries can be cleared with the Search Console Removals tool; new ones stop
appearing once the `noindex` is live.

### DIFF
--- a/.changeset/noindex-search-ask-urls.md
+++ b/.changeset/noindex-search-ask-urls.md
@@ -1,0 +1,5 @@
+---
+"gitbook": patch
+---
+
+Serve `X-Robots-Tag: noindex` on internal search/assistant URLs (`?q=` / `?ask=`) and stop disallowing them in robots.txt, so Google can crawl the directive and drop them from the index instead of reporting "Indexed, though blocked by robots.txt".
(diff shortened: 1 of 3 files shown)


gregberge (approved):

gregberge:
@nolannbiron double check with someone else that is good in SEO

nolannbiron:
I asked a friend who is a SEO expert, this makes the link crawlable, but google will always no-index them, and for those who have already been indexed, it will un-index them

PR #4439, opened 2026-07-24
RND-11830: select store URL param

BrettJephson:
Use url parameter select=javascript,go,etc to store client side selectors that adjust shown content.

Working with tabs (replaces old tab selection hash).

Resolves RND-11831, RND-11830


packages/gitbook/src/lib/select/slug.test.ts
```diff
@@ -0,0 +1,39 @@
...
+        ['Python', 'python'],
+        ['JavaScript', 'javascript'],
+        ['JS', 'js'], // note: does NOT equal "javascript" — the near-duplicate lint case
+        ['  npm  ', 'npm'], // leading/trailing whitespace trimmed
+        ['Two   Words', 'two-words'], // internal whitespace runs collapse to one dash
+        ['on-prem', 'on-prem'], // existing dashes preserved
+        ['On-Prem', 'on-prem'], // case folded
+        ['Node.js', 'node-js'], // punctuation becomes a separator
+        ['C++', 'c'], // trailing separators trimmed
+        ['C#', 'c'],
```
zenoachtig: Would we consider this normalisation an issue/bug?

packages/gitbook/src/lib/select/slug.test.ts
```diff
@@ -0,0 +1,39 @@
...
+        ['.NET', 'net'], // leading separators trimmed
+        ['a_b', 'a-b'], // underscore is not [a-z0-9]
+        ['🚀 Launch', 'launch'], // emoji stripped
+        ['café', 'caf'], // non-ascii letters are dropped (frozen v1 behaviour)
```
zenoachtig: Could/should we normalise to the unaccented form? E.g. `café` → `cafe`?

packages/gitbook/src/lib/select/slug.test.ts
```diff
@@ -0,0 +1,39 @@
...
+        ['安装', ''], // purely non-latin reduces to empty → "no slug"
```
zenoachtig: This might be a problem, but I don't know what the best way to deal with it is

packages/gitbook/src/components/DocumentView/Tabs/Tabs.tsx
```diff
@@ -43,12 +43,70 @@ export function Tabs(props: BlockProps<DocumentBlockTabs>) {
...
+
+    // When printing, we display the tabs one after the other, each as its own single-tab group so
+    // every variant is visible (no selection to hide them).
     if (context.mode === 'print') {
-        return tabs.map((tab) => {
-            return <DynamicTabs key={tab.id} id={id} tabs={[tab]} className={tcls(style)} />;
-        });
+        return tabs.map((tab) => (
+            <Fragment key={tab.id}>
+                <SelectGroupStyle slugs={[tab.slug]} />
```
zenoachtig: This generates a `<style>` for each tab. Probably not terrible, but couldn't this be a single `<style>` for the whole tab group? You can still have precedence within a style block that way.

packages/gitbook/src/components/DocumentView/Tabs/Tabs.tsx
```diff
@@ -43,12 +43,70 @@ export function Tabs(props: BlockProps<DocumentBlockTabs>) {
...
+ * stay selectable, and slugs that collide within a group are suffixed so panes remain distinct
+ * (cross-group syncing still keys off the base slug).
+ */
+function withSelectSlugs<T extends { id: string; title: string }>(
+    items: T[]
+): Array<T & { slug: string }> {
+    const counts = new Map<string, number>();
+    return items.map((item) => {
+        const base = slugifySelectValue(item.title) || slugifySelectValue(item.id) || item.id;
+        const seen = counts.get(base) ?? 0;
```
zenoachtig: Is this dedupe necessary? If we have two tabs of the same name do we need `tab` and `tab-1`? The tab should be selectable throughout, but this approach feels like it might make things more complicated if you share a URL with `?select=tab-1` and its selector later becomes `tab` because you rename one of the two. Feels to me like we should just select the first match.

packages/gitbook/src/components/Select/script.ts
```diff
@@ -0,0 +1,60 @@
+/**
+ * Read the `select` state (URL `?select=` + localStorage) and apply it to `<html>` as `data-sel-N`
+ * attributes as early as possible, so the correct content variant is visible before hydration — no
+ * flash, and it works on cached/static HTML.
+ *
+ * NOTE: this runs in `<head>` before `<body>` exists, and is stringified and injected — so it must be
+ * self-contained (no imports/closures) and touch only `document.documentElement`. The attribute name
+ * and merge rules mirror `lib/select` (`selectRankAttribute`, the store's `normalize`); keep them in
```
zenoachtig, line 8: Is there a better way to keep the merge rules in sync?

packages/gitbook/src/components/Select/SelectStateScript.tsx
```diff
@@ -0,0 +1,23 @@
...
+    const scriptArgs = JSON.stringify([
+        SELECT_STORAGE_KEY,
+        SELECT_URL_PARAM,
+        SELECT_LIST_CAP,
+    ]).slice(1, -1);
+
+    return (
+        <script
+            suppressHydrationWarning
+            dangerouslySetInnerHTML={{
```
zenoachtig, line 18: Is this the right way to load this script? I'd expect there to be a more elegant way to load additional resources at the appropriate time, but maybe not.

packages/gitbook/src/lib/select/constants.ts
```diff
@@ -0,0 +1,39 @@
...
+// DOM contract applied by consumer blocks (tabs, cards, …) and read by the generated CSS.
+
+/** Marks a group of mutually-exclusive options (e.g. a tab group). */
+export const SELECT_GROUP_ATTR = 'data-select-group';
+/** Carries a pane's slug, e.g. `data-select-option="python"`. */
+export const SELECT_OPTION_ATTR = 'data-select-option';
+/** Marks the pane shown when none of the group's slugs are active. */
+export const SELECT_DEFAULT_ATTR = 'data-select-default';
+/** Class prefix identifying a distinct candidate-set so identical sets share one stylesheet. */
+export const SELECT_SET_CLASS_PREFIX = 'sgset-';
```
zenoachtig: Nit: I find this class a bit difficult to read/understand

packages/gitbook/src/lib/select/generateSelectCSS.ts
```diff
@@ -0,0 +1,91 @@
...
+    SELECT_OPTION_ATTR,
+    SELECT_SET_CLASS_PREFIX,
+    selectRankAttribute,
+} from './constants';
+
+/**
+ * Stable, order-independent hash of a candidate set, so two groups offering the same options
+ * (e.g. `npm`/`yarn`/`pnpm` repeated across a docs site) share one generated stylesheet.
+ */
+function hashSlugSet(slugs: string[]): string {
```
zenoachtig, line 19: What is this hash needed for? Why do we hardcode it? I don't understand

packages/gitbook/src/lib/select/slug.ts
```diff
@@ -0,0 +1,31 @@
...
+ * empty slug as "no slug" and fall back to their default. The near-duplicate case (`JS` vs
+ * `JavaScript` slugifying differently and therefore failing to sync) is an editor-lint concern, not
+ * something this function tries to reconcile.
+ */
+export function slugifySelectValue(name: string): string {
+    return name
+        .normalize('NFKC')
+        .toLowerCase()
+        .replace(/[^a-z0-9]+/g, '-')
+        .replace(/^-+|-+$/g, '');
```
zenoachtig: On a Chinese, Japanese, Korean, Greek, or Cyrillic docs site, *every* tab label reduces to `''`, `uniqueSlugs` filters them all out, and the entire select/tab-sync feature is silently inert. "Установка" → `''`. And mixed labels degrade weirdly — `安装 CLI` → `cli`, which is a collision magnet across unrelated groups.

There are two problems here:
1. What characters may a slug *contain*?
2. How do those bytes survive the trip through a query param, a CSS string literal, an HTML attribute, and localStorage? The `[a-z0-9]` allowlist is really answering (2) — it's a blunt instrument for "safe to interpolate anywhere unescaped" — and paying for it with total data loss on (1).

**The normal answer to (1)** is: normalize, then allowlist by *Unicode property* rather than by ASCII range, then percent-encode at the URL boundary. The `slugify(allow_unicode=True)` is literally NFKC + a Unicode-aware `\w` allowlist + collapse — i.e. your function with one regex changed. MediaWiki does it (`zh.wikipedia.org/wiki/安装`), WordPress' `sanitize_title` does it, and `github-slugger` — the thing generating GitHub's own heading anchors — keeps 安装 intact. So you're one character class away from the industry default:

```suggestion
        .normalize('NFKC')
        .replace(/\p{C}+/gu, '')  // controls, bidi/format overrides, lone surrogates — never in a slug
        .toLowerCase()
        .normalize('NFKC')        // case mapping can leave the string un-normalized; re-settle it
        .replace(/[^\p{L}\p{N}\p{M}]+/gu, '-')
        .replace(/^-+|-+$/g, '')
        .slice(0, 64);
```

`\p{M}` matters — Devanagari and Thai leave combining marks standing after NFC, so omitting it shreds those scripts. The second `normalize` is cheap insurance for your byte-identical server/client requirement. Slice by code point (`[...s].slice(0,64).join('')`) if you want to be strict about not cleaving a surrogate pair. A cap is worth having: 30 CJK characters is 270 bytes once percent-encoded.

**Then (2), separately.** Three boundaries, and one of them is a real trap in this PR:

- `?select=` — `URLSearchParams` handles it. You get `%E5%AE%89%E8%A3%85` on copy-paste and 安装 in the address bar. Ugly-when-shared, but correct.
- HTML attributes — React escapes.
- **`generateSelectCSS`** — you interpolate the slug raw into a CSS string literal: `[${attr}="${slug}"]`. Today that's safe *by accident*, because `[a-z0-9-]` can't contain `"` or `\`. The allowlist above also can't, so it stays safe — but I'd stop relying on the accident and add an explicit escape (`\` and `"`) plus a test, because the next person to widen the charset won't reread this thread. Your `selectSetClassName` is a hash, so class-name identifier rules are already a non-issue. `charCodeAt` in `hashSlugSet` is fine — deterministic over surrogates.

packages/gitbook/src/lib/select/slug.test.ts
zenoachtig: Related: https://github.com/GitbookIO/gitbook/pull/4439/changes#r3665532485

zenoachtig (commented):
I want a second opinion from the platform team (maybe @conico974?) whether this approach makes sense from a rendering perspective. It creates the functionality we desire but I'm a bit worried about how magic some of it feels (but maybe it's just clever use of CSS)

packages/gitbook/src/components/DocumentView/Tabs/Tabs.tsx
BrettJephson: The `precedence="high"` means it gets lifted to the head of the page as a single tag. I think.

packages/gitbook/src/components/DocumentView/Tabs/Tabs.tsx
BrettJephson: Will think about how to do this because if we just set javascript you will always get the first javascript tab opening not necessarily the one you clicked. But we want the syncing to work. 

It’s not going to happen much, if an author adds two tabs with the same name in the same tab list they probably went wrong somewhere.

packages/gitbook/src/components/DocumentView/Tabs/Tabs.tsx
BrettJephson: Oh no that’s the other one, this is print. Maybe don’t need it. Will check.

packages/gitbook/src/lib/select/generateSelectCSS.ts
BrettJephson: It’s a performance thing. The CSS for this can get a bit crazy so hash lets us use the same CSS for multiple tab lists that have same slugs.

packages/gitbook/src/lib/select/slug.test.ts
BrettJephson: Fixed

packages/gitbook/src/lib/select/slug.test.ts
BrettJephson: Fixed

packages/gitbook/src/lib/select/slug.test.ts
BrettJephson: Fixed

packages/gitbook/src/components/DocumentView/Tabs/Tabs.tsx
BrettJephson: Removed as not needed for the print version.

packages/gitbook/src/lib/select/generateSelectCSS.ts
BrettJephson: I extracted the hard coded numbers into consts and added a link to the wikipedia. It’s a fast hashing algorithm we don’t really need to understand the maths behind it.

packages/gitbook/src/components/Select/SelectStateScript.tsx
BrettJephson: We do this same sort of thing elsewhere in GBO, where we don’t want to reload the app or involve SSR or the server. I believe this is the right way to do it.

packages/gitbook/src/components/DocumentView/Tabs/Tabs.tsx
BrettJephson: If there are two tabs with the same name, we will show the clicked tab and sync. Then on reload we will open the first tab with the same slug. So removes the complex dedupe stuff but maybe not what the user expects but is an edge case.

packages/gitbook/src/components/Select/script.ts
BrettJephson: Depends what you mean by better.

packages/gitbook/src/components/DocumentView/Tabs/DynamicTabs.tsx
```diff
@@ -277,7 +139,7 @@ const TabItemList = memo(function TabItemList(props: {
                     <TabItem
                         key={tab.id}
                         ref={itemRef}
-                        isActive={tab.id === activeTabId}
+                        isActive={tab.slug === activeSlug}
                         tab={tab}
                         onSelect={onSelect}
                     />
```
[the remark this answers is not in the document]
BrettJephson: Don’t need to in React 19

packages/gitbook/src/components/Select/script.ts
zenoachtig: I meant something that keeps it in sync better, but also fine as-is

zenoachtig (approved):

PR #4443, opened 2026-07-27
Remove search value from URL on unfocus

tgargula:
## Description
Remove search value from URL on unfocus and ensure the number of results is correct on unfocus.

Restores the last search query when reopening search without breaking navigation, and keeps search state clean when closing the search container.

## Demo
https://github.com/user-attachments/assets/96011c0b-c540-47fd-ab5f-efb043e00f74

https://github.com/user-attachments/assets/274af7a1-1b0f-43f0-9908-dc9bf3a05cad

resolves RND-11972

### DIFF
--- a/.changeset/search-last-query.md
+++ b/.changeset/search-last-query.md
@@ -1,0 +1,5 @@
+---
+"gitbook": patch
+---
+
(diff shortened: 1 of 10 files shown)


tgargula:
I added the tags for the demo. They're not committed here

conico974 (approved):

PR #4464, opened 2026-08-04
Reduce bundle size on the server

conico974:
Refactor the code to move mermaid and scalar client out of the bundle. 
Introduce Scalar API client integration with runtime support, including new components for API interaction and enhancements for lazy loading. 
Update scripts for asset generation and clean-up.

Bundle size goes from roughly 39Mb to 25Mb


packages/react-openapi/src/ScalarApiButton.tsx
```diff
@@ -26,121 +37,113 @@ export function ScalarApiButton(props: {
...
+                            strokeOpacity="0.3"
+                            strokeWidth="3"
+                        />
+                        <path
+                            d="M12 2a10 10 0 0 1 10 10"
+                            stroke="currentColor"
+                            strokeWidth="3"
+                            strokeLinecap="round"
+                        />
+                    </svg>
```
tgargula, line 88: do we have separate components for spinners?

tgargula (approved):
L👌TM

packages/react-openapi/src/ScalarApiButton.tsx
conico974: Not in the react-openapi package, and it's not worth adding dependencies just for that

PR #4475, opened 2026-08-06
Render site-space custom home page at root

tgargula:
Render a site-space custom home page at its placement root while preserving the full space and normal page URLs.

This allows a site space's root URL to display a custom home page (if configured via `pageId`) while keeping all other pages accessible at their normal paths. The linker uses direct page paths when a custom home page is set, ensuring consistent routing across all page resolution paths.

## Changes
- Add `resolveSiteSpaceCustomHomePage`, `getSiteSpacePagePaths`, and `getLinkerForSiteSpace` helpers
- Update all page resolution paths (fetch, route handlers, markdown pages, RSS, MCP) to check for custom home pages
- Add `linkerWithDirectPagePaths` to preserve explicit page paths when the root has a custom landing page
- Update table of contents encoding to use site-space aware paths
- Add comprehensive tests for custom home page resolution

## Demo

https://github.com/user-attachments/assets/c4762f86-e42e-4e1c-b9ca-ada25cad5815


packages/gitbook/src/app/sites/dynamic/[mode]/[siteURL]/[siteData]/~gitbook/mcp/handler.ts
```diff
@@ -198,7 +198,11 @@ export async function handleMcpRequest(
                             })
                         );
 
-                        const resolved = resolvePagePath(revision.pages, match.pagePath ?? '');
+                        const resolved =
+                            (match.pagePath === ''
+                                ? resolveSiteSpaceCustomHomePage(match.siteSpace, revision.pages)
+                                : undefined) ??
+                            resolvePagePath(revision.pages, match.pagePath ?? '');
```
Comment (conico974) [inline packages/gitbook/src/app/sites/dynamic/[mode]/[siteURL]/[siteData]/~gitbook/mcp/handler.ts]: I don't think we need the `?? ''` here

packages/gitbook/src/app/sites/dynamic/[mode]/[siteURL]/[siteData]/~gitbook/mcp/handler.ts
```diff
@@ -409,7 +413,11 @@ export async function handleMcpRequest(
...
```
Comment (conico974) [inline packages/gitbook/src/app/sites/dynamic/[mode]/[siteURL]/[siteData]/~gitbook/mcp/handler.ts]: Same

packages/gitbook/src/app/sites/dynamic/[mode]/[siteURL]/[siteData]/~gitbook/mcp/handler.ts
```diff
@@ -198,7 +198,11 @@ export async function handleMcpRequest(
                             })
                         );
 
-                        const resolved = resolvePagePath(revision.pages, match.pagePath ?? '');
+                        const resolved =
+                            (match.pagePath === ''
```
Comment (conico974) [inline packages/gitbook/src/app/sites/dynamic/[mode]/[siteURL]/[siteData]/~gitbook/mcp/handler.ts]: Is `''` the only possible value, or do we need to take into account other stuff here like shareLink or basePath or even `"/"`?

packages/gitbook/src/components/SitePage/fetch.ts
```diff
@@ -51,7 +52,10 @@ async function resolvePage(context: GitBookSiteContext, params: PagePathParams |
     const pathname = rawPathname.toLowerCase();
 
     // When resolving a page, we use the lowercased pathname
-    const page = resolvePagePath(revision.pages, pathname);
+    const page =
+        (pathname === ''
```
conico974: Same

packages/gitbook/src/components/SitePage/SitePage.tsx
```diff
@@ -265,13 +269,27 @@ export async function getSitePageData(props: SitePageProps) {
...
-            context.linker.toPathForPage({
-                pages: context.revision.pages,
-                page: pageTarget.page,
-            })
+    } else {
+        const customHomePage = resolveSiteSpaceCustomHomePage(
+            context.siteSpace,
+            context.revision.pages
         );
+        const isExpectedPagePath = customHomePage
```
conico974: I find this hard to read personally (i.e. the ternary plus the equality at the end)

packages/gitbook/src/lib/sites.ts
```diff
@@ -73,6 +76,45 @@ export function filterSiteSpacesByLocale(
     );
 }
 
+/**
+ * Resolve the custom home page for a site space in the current revision.
+ */
+export function resolveSiteSpaceCustomHomePage({ pageId }: SiteSpace, pages: Revision['pages']) {
+    const resolved = pageId ? resolvePageId(pages, pageId) : undefined;
+
+    // A group ID resolves to its first document, but only documents can be custom home pages.
```
conico974, line 98: I guess it is enforced at the API level, should we throw then here ?

conico974 (commented):
A few question, but if it's not an issue it should be good

packages/gitbook/src/lib/sites.ts
tgargula: It's not possible to assign an invalid page to the site. However, we're not covering in the API when the page goes stale. In that case, we'll redirect to the first page instead of throwing. I added some more comments to explain that

tgargula: anyway, I think it's better to default to the first page in case any error/misconfiguration occurs rather than throwing in the published docs

packages/gitbook/src/app/sites/dynamic/[mode]/[siteURL]/[siteData]/~gitbook/mcp/handler.ts
Comment (tgargula) [reply packages/gitbook/src/app/sites/dynamic/[mode]/[siteURL]/[siteData]/~gitbook/mcp/handler.ts]: basePath and "/" are already handled in extractPagePath (covered by tests in pages.test.ts) -- getURLPathname trims trailing slashes so "/" normalizes the same as the bare base URL, and there’s an explicit branch for root-of-domain sites. 
`shareLink` isn’t a thing here. `SiteSpace.urls` only exposes published, so there’s nothing else to account for.

conico974 (approved):

PR #4477, opened 2026-08-07
Fix text visibility issues on pages with background covers in Firefox and iOS Safari

conico974:
Address text visibility problems that occur on pages with background covers specifically in Firefox and iOS Safari. Implemented a solution that ensures text remains readable by marking elements that overlap the cover and adjusting styles accordingly.


BrettJephson (approved):
lgtm

PR #4478, opened 2026-08-07
feat: add CSS browser compatibility check for pull requests

conico974:
- Introduced a new script to check for newly added CSS declarations and their browser compatibility using stylelint.
- Added a GitHub Actions workflow to run the compatibility check on pull request events.
- Updated package.json to include necessary dependencies for the compatibility check.
- Implemented functions to analyze CSS changes and report unsupported features in a comment on the pull request.
- Created tests for the compatibility checking logic and comment formatting.


tgargula (approved):
L👌TM

should we also add it to linting so we can catch it in the editor/agent level?
the browserslist was last updated one year ago, should we also take a look if it needs any changes?

conico974:
> should we also add it to linting so we can catch it in the editor/agent level?

I need to figure out how to do it properly, we would have 130 violations today, but not all of them are bad
Even this one, I'm sure I'll have to tweak it a bit

> the browserslist was last updated one year ago, should we also take a look if it needs any changes?

Yeah it could be worth checking again

PR #4479, opened 2026-08-07
Fix icons rendering invisible on pages with a background cover

peterwhite:
Icons in headings and paragraphs are invisible on pages with a background cover. 

Before: 
image

After:
image


peterwhite:
Superseded by #4479

PR #4485, opened 2026-08-11
Fix sticky search results when switching the search filter RND-12265

peterwhite:
Switching the search filter leaves results from the previous scope stuck on top of the new ones, permanently.

On https://docs.sonarsource.com/sonarqube-cloud: search `administer projects`, then switch the filter to **SonarQube Cloud — Only the current section**. Two "SonarQube Server / Viewing projects" rows stay pinned over the new results.

It's timing-sensitive - the switch has to land while results are still streaming in, which is why changing the search term sometimes appears to clear it and sometimes doesn't.

### Cause

The stuck rows carry `data-motion-pop-id` - framer-motion's marker for a `popLayout` child that started exiting but never unmounted.

`popLayout` pins a leaving row to `position: absolute !important` until its exit completes. None of the children define an `exit` prop, only `layout="position"` - so when the removal collides with an in-flight layout animation, exit-complete never fires and the row is orphaned.

### Fix

Remove the `AnimatePresence` wrapper. It wasn't doing anything: enter animations are CSS and reordering comes from `motion.div layout="position"`, both of which work without it

### DIFF
--- a/.changeset/tidy-donkeys-shave.md
+++ b/.changeset/tidy-donkeys-shave.md
@@ -1,0 +1,5 @@
+---
+"gitbook": patch
+---
+
+Fix search results from a previous scope staying stuck on top of the new results when switching the search filter.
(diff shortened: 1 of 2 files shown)


tgargula (approved):
L👌TM

PR #4488, opened 2026-08-12
Fix search result links for page matches (RND-11916)

jpreynat:
## Summary

Fixes [RND-11916](https://linear.app/gitbook-x/issue/RND-11916/search-results-link-to-anchors-instead-of-page-top-even-on-title).

Search results now use the API's optional `resultType` discriminator to choose their navigation target:

- page/title matches open at the top of the page
- section matches retain the best-section anchor
- older responses without `resultType` preserve the existing anchored behavior

This supersedes #4455 with the confirmed GitBook-X response contract and focused unit coverage.

## Root cause

Published-docs search always linked a result to its best section whenever a section preview had a body. The UI did not carry through whether the backend matched the page's own fields or the section content, so title matches could incorrectly jump partway down the page.

## Changes

- carry `resultType` from `SearchPageResult` into the computed page result
- select the page or section URL through a small pure helper
- cover page, section, missing-discriminator, and missing-preview cases
- add a patch changeset for `gitbook`

## Impact

Readers clicking broad page/title matches land at the beginning of the article. Specific section matches continue to jump directly to the relevant content, with backward-compatible behavior during mixed API deployments.

## Validation

- `bun test packages/gitbook/src/components/Search` — 15 tests passed
- `bun run format`
- scoped Biome check — no new errors
- `bun run typecheck` — reaches the GitBook package but is currently blocked by unrelated existing generated API/type mismatches; no reported error references this change

## Changelog

- [Fix] Link page-level search matches to the top of the page while preserving anchors for section matches.


packages/gitbook/src/components/Search/SearchPageResultItem.tsx
```diff
@@ -26,9 +27,7 @@ export const SearchPageResultItem = React.forwardRef(function SearchPageResultIt
     const language = useLanguage();
 
     const bestSection = item.type === 'page' ? item.bestSection : undefined;
```
conico974, line 29: This one won't cause other issues right ?

conico974 (approved):

packages/gitbook/src/components/Search/SearchPageResultItem.tsx
jpreynat: It shouldn't, no

PR #4496, opened 2026-08-13
Fix Docs Embed being stuck on the Search tab (RND-12435)

peterwhite:
In the Docs Embed, clicking the Docs or Assistant tab from the Search tab did nothing.

Fix: await the state update so the URL write lands before the push.

Tested against `gitbook.com/docs/~gitbook/embed/search`

### DIFF
--- a/packages/gitbook/src/components/Search/useSearchController.tsx
+++ b/packages/gitbook/src/components/Search/useSearchController.tsx
@@ -176,7 +176,7 @@ export function useSearchController(
 
     const onClose = React.useCallback(
         async (to?: string) => {
-            setSearchState((prev) => {
+            await setSearchState((prev) => {
                 if (!prev) return null;
 
                 if (prev.query !== null) {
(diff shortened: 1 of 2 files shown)


packages/gitbook/src/components/Search/useSearchController.tsx
```diff
@@ -176,7 +176,7 @@ export function useSearchController(
 
     const onClose = React.useCallback(
         async (to?: string) => {
-            setSearchState((prev) => {
+            await setSearchState((prev) => {
```
taranvohra, line 179: Hm I'm not sure this is right. state setters do not usually return a promise

taranvohra: Ah it's encapsulating `nuqs` 👍

taranvohra (approved):
Can you check if there are other places where we are not awaiting it?

PR #4511, opened 2026-08-18
Improve image signature handling and resizing functionality

conico974:
Enhance image processing by adding support for image signature version 3, and refining error handling in image resizing logic. Improve options and accept header handling for resizing functionality.


packages/gitbook/src/lib/images/signatures.ts
```diff
@@ -102,22 +129,41 @@ const generateSignatureV0: SignFn = async (input) => {
...
+    '3': generateSignatureV3,
 };
 
 export function isSignatureVersion(input: string): input is SignatureVersion {
     return Object.keys(IMAGE_SIGNATURE_FUNCTIONS).includes(input);
 }
+
+/**
+ * Compare two signatures in constant time, to avoid leaking a valid signature byte by byte.
+ * We can't use `node:crypto`'s `timingSafeEqual` as this also runs on the edge runtime.
```
jpreynat, line 157: Can't you use an equivalent of https://developers.cloudflare.com/workers/runtime-apis/web-crypto/#timingsafeequal?

packages/gitbook/src/routes/image.ts
```diff
@@ -155,8 +123,37 @@ async function resizeImageWithFallback(
...
+                        // It sometimes happen when upstream block fetch from  our server
+                        throw new Error('Upstream error, falling back to a redirect');
+                    case ImageRejectReason.UnsupportedContentType:
+                        throw new Error('Unsupported content type, falling back to a redirect');
+                    case ImageRejectReason.InternalError:
+                        throw new Error('Internal error, falling back to a redirect');
+                    default:
+                        throw new Error(
+                            `Unknown reject reason "${rejectReason}", falling back to a redirect`
+                        );
```
jpreynat, line 146: The "original images" here are on the `files.gitbook.com` domain or another one?

The main risk of serving just anything on the current domain is that it causes issues with access to domain's private data (especially cookies, etc...). If redirecting to files.gitbook.com is the fallback, it could remain bad IMO because that could leak data from on main app's domain too.
We should just make sure where the images are served, and maybe in this case we could/should serve images on yet another, isolated domain.

packages/gitbook/src/lib/images/signatures.ts
conico974: I don't think it's available in vercel...

packages/gitbook/src/routes/image.ts
```diff
@@ -164,12 +164,31 @@ async function resizeImageWithFallback(
...
+/**
+ * Check if a URL is hosted on a GitBook domain.
+ */
+function isGitBookHostedURL(url: string): boolean {
+    if (!URL.canParse(url)) {
+        return false;
+    }
+
+    const { hostname } = new URL(url);
+    return hostname === 'gitbook.com' || hostname.endsWith('.gitbook.com');
```
jpreynat: Just in case, can you also add the checks for `gitbook.io`?

jpreynat (approved):

PR #4518, opened 2026-08-19
Keep block, toolbar and print CSS out of the render-blocking path (RND-12524)

nolannbiron:
Published pages load 8 render-blocking stylesheets (620 KB raw / 71.6 KB compressed). Four of them are for features the page usually doesn't have: the OpenAPI block CSS (185 KB), ContentKit (33 KB), the admin toolbar CSS modules, and the PDF print styles.

- Build the OpenAPI and ContentKit stylesheets as standalone assets (`scripts/generate-block-css.ts`) and link them from the blocks themselves through React's `precedence` hoisting, so only pages that render those blocks pay for them. The blocks still server-render, and the sheet still lands in `<head>` before the block streams in, so nothing flashes unstyled. Output is rule-for-rule identical to what webpack emitted.
- Move `AdminToolbarClient` behind a lazy chunk. Its barrel had to be trimmed too, otherwise `SiteLayout`'s import pulled the CSS modules straight back into the route.
- Inline the PDF print styles in `PDFRootLayout` — as a CSS import they were leaking into content pages.
- Fold the `ZoomImage` and `headerLinks` CSS modules into `globals.css`. Once the toolbar CSS left, webpack split them into two extra ~1 KB sheets of their own.

Every route goes from 8 stylesheets to 3 (2 font chunks + the main Tailwind chunk). The fonts are handled in the PR stacked on top of this one.


nolannbiron:

|---|---|---|---|---|

PR #4519, opened 2026-08-19
Self-host default fonts and inline only the ones a site uses (RND-12524)

nolannbiron:
`src/fonts/default.ts` instantiated 23 `next/font` families in a single module, so the 507 `@font-face` rules for all of them shipped to every page across two or three render-blocking stylesheets (195 KB) — for the two families a site actually uses.

Resolve the faces from the Google Fonts CSS API at generate time, vendor the woff2 files under `~gitbook/static/fonts` (6.1 MB, 220 files, gitignored and built by `generate:assets`), and inline the picked families' rules in the head — the path custom fonts already took.

The generated data reproduces next/font's 491 faces and all 16 metric-adjusted fallback faces exactly, so rendering and CLS are unchanged. Regenerate the metadata with `bun run generate:fonts`.

Measured on the preview deploy vs production, per cold page load:

| | production | preview | |
|---|---|---|---|
| render-blocking stylesheets | 8 | **1** | −7 |
| CSS | 70.0 KB | **32.6 KB** | −37.4 KB |
| HTML | 89.1 KB | 94.4 KB | +5.3 KB (inline font rules) |
| **total** | 159.1 KB | **127.0 KB** | **−32.1 KB** |

The build still downloads the woff2 files from `fonts.gstatic.com`, but as a dedicated step with retries and resume rather than inside webpack, which is where it was flaky before.


nolannbiron:

|---|---|---|---|---|

PR #4521, opened 2026-08-19
Stop shipping a JavaScript parser and a redundant emoji table to every reader (RND-12524)

nolannbiron:
Lighthouse on `mariadb.com/docs` scores **50**, and the render-blocking CSS audit it was opened against now estimates **0 ms** of savings. The score is lost almost entirely to JavaScript: **LCP 11.9 s** (94% of it *render delay* on a text element — the main thread is too busy to paint) costs 25 points, and **TBT 650 ms** costs 16 more. A docs page ships **26 scripts, 568 KB transferred, 1,814 KB decoded, 5,699 ms of main-thread time**.

I attributed every byte with source maps. Two findings were worth acting on:

**A full JavaScript parser was shipped to every reader (−252 KB).** Every code block rendered client-side called `useEvaluateInlineExpression`, so `@gitbook/expr` — and with it acorn, acorn-loose, acorn-walk, escodegen, esutils, estraverse, source-map and eval-estree-expression — sat in the eager entry of every page. That is more JavaScript than React, loaded so that inline-expression blocks could be evaluated. The server already computes `hasInlineExpression`, so it now picks the boundary: blocks without expressions get a `ClientCodeBlock` that no longer imports the runtime, and the rest go through a `next/dynamic` wrapper. `ssr: true` throughout, so evaluated values stay in the HTML.

**The emoji table was half redundant (−35 KB).** Every value in the generated base → fully-qualified map is its own key plus `fe0f`/`200d` joiners, so the keys carried no information. It now ships the qualified sequences only and builds the index on first use, which also keeps a 1,700-property object off the parse path for pages that render no emoji. A test asserts all 1,698 mappings resolve exactly as before.

**Result: 1,814 KB → 1,529 KB decoded (−285 KB, −15.7%).**

### What this will not do

It will not fix the score on its own. LCP needs to fall below 4 s to score anything at all, and −15.7% of parse/execute will not take 11.9 s there. Expect a few points, mostly via TBT. The remaining weight is `@base-ui/react` (187 KB), `zod` via the always-mounted `AIChatProvider` (132 KB) and `framer-motion` (128 KB) — each needs a design change rather than an import fix, and they are not in this PR.

Also included, as hygiene rather than a measured win: `Emoji` left the primitives barrel, `getAIChatName` moved out of `AIChat.tsx`, and icon-only call sites now import the specific module. Those did not move the bundle on their own — the real importers are legitimate — but they stop the next heavy dependency riding a barrel into every chunk.


nolannbiron:

|---|---|---|---|---|

PR #4522, opened 2026-08-19
Self-host default fonts and inline only the ones a site uses (RND-12524)

nolannbiron:
Supersedes #4519, rebased standalone on `main` (no longer stacked on #4518, which measured no paint improvement on its own and stays closed).

Every page currently ships 3 render-blocking stylesheets containing 507 `@font-face` rules for all 23 default font families, while a site uses at most two. On production (cross-origin `static-2v.gitbook.com`), the font CSS → woff2 chain is the largest render-blocking item Lighthouse reports (794 ms of ~1,300 ms on `mariadb.com/docs`), gating the text LCP element.

Changes:
- `scripts/generate-font-faces.ts` generates a per-family manifest from the Google Fonts CSS API (same UA as next/font — byte-identical faces, verified 491/491), committed as `src/fonts/generated/`
- `scripts/download-fonts.ts` (wired into `generate:assets`) downloads the woff2 files to `public/~gitbook/static/fonts/`, served first-party via `getAssetURL`
- `src/fonts/default.ts` replaces the 23 `next/font/google` calls with the generated manifest; `CustomizationRootLayout` inlines the `@font-face` rules for the site's content + mono fonts and emoji only
- Fallback font metrics (`size-adjust`/`ascent-override`) preserved via `next/dist/server/font-utils`, so no CLS regression

Measured on preview (`/url/mariadb.com/docs`): Lighthouse's modeled render-blocking savings drop from 400–500 ms (baseline, n=13) to 160–330 ms (n=7). Same-origin previews cannot reproduce the production topology, where the removed font-CSS → woff2 chain is the largest render-blocking item (794 ms of ~1,300 ms); production impact to be measured after merge.

Note: font assets 404 on aliased `*.workers.dev` preview hosts because `getAssetURL` resolves against the deployment's configured `GITBOOK_URL` — same as existing generated assets (icons, math, scalar); validate on the Vercel preview or the canonical 2c preview URL.


packages/gitbook/scripts/download-fonts.ts
```diff
@@ -0,0 +1,73 @@
+import { copyFile, mkdir, readdir, rm, writeFile } from 'node:fs/promises';
+import { dirname, join, relative } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+import sourcesData from '../src/fonts/generated/sources.json' with { type: 'json' };
+import type { FontSourcesData } from '../src/fonts/types';
+
+const CONCURRENCY = 8;
+const ATTEMPTS = 3;
```
conico974: We could bump this one to 5

packages/gitbook/package.json
```diff
@@ -118,8 +118,9 @@
     },
     "scripts": {
         "generate": "./scripts/generate.sh",
-        "generate:assets": "bun ./scripts/generate-mermaid-runtime.ts && bun ./scripts/generate-scalar-runtime.ts",
-        "clean": "rm -rf ./.next && rm -rf ./public/~gitbook/static/icons && rm -rf ./public/~gitbook/static/math && rm -rf ./public/~gitbook/static/mermaid && rm -rf ./public/~gitbook/static/scalar",
+        "generate:assets": "bun ./scripts/generate-mermaid-runtime.ts && bun ./scripts/generate-scalar-runtime.ts && bun ./scripts/download-fonts.ts",
+        "generate:fonts": "bun ./scripts/generate-font-faces.ts",
```
conico974, line 122: Don't we need to do that on the generate:assets phase ? I don't see it used anywhere

packages/gitbook/scripts/download-fonts.ts
conico974: Forget about that, github didn't let me access this comment without posting :smile:

conico974 (approved):

PR #4526, opened 2026-08-20
Load the admin toolbar and its CSS lazily (RND-12524)

nolannbiron:
The admin toolbar is rendered on every published page but only ever becomes visible for admins, client-side, after hydration. Its CSS modules still reach the layout's static import graph, so every visitor pays a render-blocking stylesheet in `<head>` for UI they will never see.

- `AdminToolbarLazy` wraps the client component in `dynamic(..., { ssr: false })`, so the toolbar's CSS is only injected with its lazy chunk when it actually mounts — correct since the toolbar only ever appears post-hydration.
- `headerLinks.module.css` and `ZoomImage.module.css` shared a CSS chunk with the toolbar; extracting the toolbar alone would have left them behind as two new tiny render-blocking sheets, so their rules move into `globals.css` with global class names.

Verified on a production build: the `(content)` layout's `entryCSSFiles` drop from 2 stylesheets to 1 (the main Tailwind sheet), with no new chunks created. On a published page that's one fewer render-blocking request; no change expected on Lighthouse scores.

Extracted from the closed #4518, whose remaining scope (per-block OpenAPI/ContentKit stylesheets) stays closed.


zenoachtig (approved):
Nit: this PR is more generally about removing render-blocking CSS modules, combined with the lazy admin toolbar change. Took me a minute to understand why these CSS changes were in here, but makes sense. Looks good 👌

PR #4529, opened 2026-08-20
Defer tracking scripts until after load and idle

nolannbiron:
Analytics integration scripts were preloaded (`ReactDOM.preload`) and injected as `<script async>` in the SSR HTML — more aggressive than Google's own install snippet, and beyond what `@next/third-parties` does. On `mariadb.com/docs`, the two GA tags plus ZoomInfo account for ~353 KB and ~250 ms of long tasks during the load window — where early clicks produce the INP failures (CrUX p75: 204 ms, sole failing Core Web Vital).

Scripts of integrations on an explicit allowlist of pure trackers (no visitor-facing UI: `googleanalytics`, `zoominfo`, `amplitude`, `heap`, `mixpanel`, `fathom`, `plausible`, `piwik`, `fullstory`, `koala`, `marketo`, `ahrefs`, `reo`, `salesviewer`, `unify`) are now loaded with `next/script strategy="lazyOnload"` (after `load` + idle), per web.dev guidance. **Everything else — consent managers (OneTrust), chat widgets, assistants, and any integration not explicitly listed — keeps the exact current behavior** (preload + eager async script), so unknown or UI-bearing integrations cannot regress.

Analytics impact for allowlisted trackers: visits bouncing before `load` (~2-3 s) are no longer counted.


packages/gitbook/src/components/SiteLayout/SiteLayout.tsx
```diff
@@ -18,6 +19,31 @@ import { GITBOOK_API_PUBLIC_URL, GITBOOK_ASSETS_URL, GITBOOK_ICONS_URL } from '@
 import { getResizedImageURL } from '@/lib/images';
 import { isSiteIndexable } from '@/lib/seo';
 
+// Pure trackers with no visitor-facing UI, safe to load after `load` + idle. Anything not
```
zenoachtig, line 22: Feels a bit random and fragile to have in here but I have no real better way besides doing it on the integration category or something else that would require more tooling setup.

zenoachtig (approved):

packages/gitbook/src/components/SiteLayout/SiteLayout.tsx
nolannbiron: that's what I'm gonna do next, but I'd like to deploy this first to see if it improves load time or not

packages/gitbook/src/components/SiteLayout/SiteLayout.tsx
```diff
@@ -88,9 +116,13 @@ export async function SiteLayout(props: {
...
 
             <LoadIntegrations />
-            {scripts.length > 0
-                ? scripts.map(({ script }) => <script key={script} async src={script} />)
-                : null}
+            {scripts.map(({ script }) =>
+                isDeferrableScript(script) ? (
+                    <Script key={script} src={script} strategy="lazyOnload" />
+                ) : (
+                    <script key={script} async src={script} />
```
gregberge, line 123: Why are we not using `<Script` for this one? I know it was not the case but just a curiosity.

packages/gitbook/src/components/SiteLayout/SiteLayout.tsx
gregberge: Yeah I think it has to be done at integration level, it's not scalable here. But you can do it in two steps so we fix it right now.

gregberge (approved):

packages/gitbook/src/components/SiteLayout/SiteLayout.tsx
nolannbiron: `<Script>` renders after hydration, so cookie banners and similar would load later than they currently do, I'm keeping `<script>` for now, until I can measure the delay

PR #4532, opened 2026-08-21
Add API token expiration check

conico974:
Introduce a mechanism to verify if an API token is expired or about to expire, along with corresponding tests to ensure functionality. This enhancement improves the handling of API tokens during requests.

### DIFF
--- a/packages/gitbook/src/lib/data/lookup.ts
+++ b/packages/gitbook/src/lib/data/lookup.ts
@@ -1,12 +1,16 @@
-import type { PublishedSiteContentLookup, SiteVisitorPayload } from '@gitbook/api';
+import type { GitBookAPI, PublishedSiteContentLookup, SiteVisitorPayload } from '@gitbook/api';
 
 import { apiClient } from './api';
 import { getExposableError } from './errors';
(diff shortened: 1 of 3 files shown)


packages/gitbook/src/lib/data/lookup.ts
```diff
@@ -28,24 +32,50 @@ export async function lookupPublishedContentByUrl(
...
+            () => resolveURL()
         );
 
+        const resolved = callResult.error ? null : callResult.data.data;
+        if (
+            resolved &&
+            !('redirect' in resolved) &&
+            // Only alternatives we'd actually accept are worth a second round-trip.
+            (alternative.primary || resolved.complete) &&
+            isAPITokenExpired(resolved.apiToken)
```
taranvohra, line 64: Why would we need this check? It means our API is returning expired tokens or if it is returning from the cache then it cached it longer than it's expiry, no?

conico974: Yeah that's exactly why, there is an entire thread in slack in on-call about that

taranvohra (approved):

PR #4537, opened 2026-08-24
Preserve backend ranking in published search

jpreynat:
## Overview

Published search currently overrides the canonical backend page order twice: the route flattens space groups and sorts by raw BM25 score, then the default multi-section scope multiplies and sorts the two independently fetched lists by those non-comparable scores. `reciprocalRankFusion` consequently pins the wrong first three remote results.

This change preserves the backend decision through the final search UI:

- Reconstruct the global page order from ascending `SearchPageResult.rank` after flattening space groups.
- Keep context records after all canonical page results.
- Carry `rank` in the computed remote page model and retain it when local and remote pages merge.
- Replace the current-space raw-score multiplier with weighted RRF (`k = 60`, current-space weight `1.01`). At displayed result depths, this gives the current space a modest equal-rank preference without changing either backend list internally.
- Feed the correctly ordered remote list into the existing three-result pinning behavior.
- Derive the click target and preview together:
  - Page matches link to the page root and display the page description only.
  - Section matches link to the best section and display its heading/excerpt.
  - An empty page description remains empty instead of falling back to unrelated section content.

Raw search scores remain available in the response model, but they are no longer mutated or used to reorder ranked pages. The page/section presentation also fixes [RND-12578](https://linear.app/gitbook-x/issue/RND-12578/remove-the-section-title-from-the-result-when-its-a-link-to-the-top-of).

## Backend dependencies

This consumes the released `@gitbook/api@0.196.0` contract from:

- [GitbookIO/gitbook-x#25054](https://github.com/GitbookIO/gitbook-x/pull/25054), which adds the canonical one-based `SearchPageResult.rank`.
- [GitbookIO/gitbook-x#25061](https://github.com/GitbookIO/gitbook-x/pull/25061), which adds the authoritative page-level `SearchPageResult.description`.

The generated API types are used directly; no generated types are hand-edited and no rollout compatibility cast remains.

## Benchmark expectations

The production-derived replay loses five expectations after the backend engine. The regression fixtures cover those failures with the canonical ranks and destinations preserved:

| Expectation | Before | After |
| --- | ---: | ---: |
| Snyk `Snyk CLI` exact-title target | 4 | 1 |
| Runway `types of automations` | 2 | 1 |
| Runway `types of notifications` | 3 | 1 |
| Cardstream `Method URL` | 2 | 1 |
| Vectra page/section destination | inconsistent | matches `resultType` |

| Replay layer | Before | After / expected |
| --- | ---: | ---: |
| gitbook-x engine benchmark | 14/15 | 14/15 (unchanged) |
| final published replay | 9/14 | 14/14 |

The production replay harness is not in this repository, so the final 14/14 is the expected replay result encoded by the captured minimal fixtures rather than a production rerun from this checkout.

## Performance

The frontend fetch plan is unchanged: default multi-section search still makes the same two parallel backend requests, and every other scope still makes one. The page description is carried by the existing search response and is already retrieved by the backend search query, so this introduces no extra backend request or serial network work.

The new ranking work is an in-memory rank sort/fusion. A Bun microbenchmark of 100,000 weighted-RRF merges with 40 results each averaged 0.0008 ms per merge on this machine, so no material latency overhead is expected.

## Testing

- Focused Bun tests for global rank reconstruction, current/other-space fusion, pinned remote order, merging, and page/section presentation: 13 pass.
- `bun run typecheck` in `packages/gitbook`: passes.
- `bun run lint`: passes.
- `bun run format`: passes.

The ranking regression was added red first: a rank-1 Snyk result with score 1 had to remain ahead of a rank-4 result with score 42 across interleaved space groups.

### DIFF
--- a/packages/gitbook/src/components/Search/getPageResultHref.ts
+++ b/packages/gitbook/src/components/Search/getPageResultHref.ts
@@ -2,6 +2,7 @@ import type { ComputedPageResult } from './search-types';
 
 type PageResultPresentation = {
     href: string;
+    description?: string;
     preview?: {
         title?: string;
         body?: string;
@@ -10,19 +11,20 @@ type PageResultPresentation = {
 
 /** Keep the displayed preview consistent with the result's single click destination. */
 export function getPageResultPresentation(
-    result: Pick<ComputedPageResult, 'bestSection' | 'href' | 'resultType'>
+    result: Pick<ComputedPageResult, 'bestSection' | 'description' | 'href' | 'resultType'>
 ): PageResultPresentation {
     const { bestSection } = result;
 
     if (result.resultType === 'page') {
         return {
             href: result.href,
-            preview: bestSection?.body ? { body: bestSection.body } : undefined,
+            description: result.description,
         };
     }
 
     return {
         href: bestSection?.href ?? result.href,
+        description: bestSection ? undefined : result.description,
         preview: bestSection
             ? {
                   title: bestSection.title,
@@ -33,7 +35,7 @@ export function getPageResultPresentation(
 }
 
 export function getPageResultHref(
-    result: Pick<ComputedPageResult, 'bestSection' | 'href' | 'resultType'>
+    result: Pick<ComputedPageResult, 'bestSection' | 'description' | 'href' | 'resultType'>
 ): string {
     return getPageResultPresentation(result).href;
 }
--- a/packages/gitbook/src/components/Search/getPageResultHref.ts
+++ b/packages/gitbook/src/components/Search/getPageResultHref.ts
@@ -1,12 +1,39 @@
 import type { ComputedPageResult } from './search-types';
 
-/** Return the page top for page matches and the preview anchor for section matches. */
+type PageResultPresentation = {
+    href: string;
+    preview?: {
+        title?: string;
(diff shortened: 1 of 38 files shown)


packages/gitbook/src/components/Search/getPageResultHref.ts
```diff
@@ -1,12 +1,39 @@
...
     result: Pick<ComputedPageResult, 'bestSection' | 'href' | 'resultType'>
-): string {
+): PageResultPresentation {
+    const { bestSection } = result;
+
     if (result.resultType === 'page') {
-        return result.href;
+        return {
+            href: result.href,
+            preview: bestSection?.body ? { body: bestSection.body } : undefined,
```
conico974: Won't this make this RND-12578 more of an issue ?

jpreynat: Good catch. Addressed in 91cb47b8e after gitbook-x #25061 and @gitbook/api 0.196.0: page results now show the page description and discard the section preview entirely; section results keep their anchor, heading, and excerpt. Added Vectra coverage plus an empty-description case to ensure section text is never used as a fallback.

conico974 (approved):

PR #4554, opened 2026-08-28
Chunk oversized visitor authentication cookies

peterwhite:
Sites issuing large visitor JWTs (an Entra site with approx. 60 group claims triggered this) end up with a visitor auth cookie over the 4KB browser limit. 

The browser drops the `Set-Cookie` silently, so after a successful sign-in every request arrives without a token and gets redirected back to the IdP. 

#4421 fixed the same failure for API token cookies by splitting them into 4,000-char chunks. This PR moves that logic into `lib/chunked-cookies.ts` unchanged and applies it to the visitor auth cookie. `api-token-cookie.ts` now uses the shared helper.

Existing single cookies still parse, so no need to sign-in again after deploy. A missing chunk reads as no token and the visitor re-authenticates, and writing a smaller token expires leftover chunk cookies from a previous bigger one.

### DIFF
--- a/packages/gitbook/e2e/internal.spec.ts
+++ b/packages/gitbook/e2e/internal.spec.ts
@@ -1656,6 +1656,71 @@ const testCases: TestsCase[] = [
         ],
     },
     {
+        name: 'Visitor Auth - Space (oversized token)',
+        contentBaseURL: 'https://gitbook.gitbook.io/gbo-va-space/',
+        // Our Cloudflare stack still folds multiple Set-Cookie headers into one,
+        // breaking chunked cookies (variant of opennextjs-cloudflare#501).
+        skip: process.env.ARGOS_BUILD_NAME === 'v2-cloudflare',
+        tests: [
+            {
+                name: 'Oversized token is chunked into cookies and survives navigation',
+                url: () => {
+                    const privateKey = '70b844d0-c519-4532-8586-5970ce48c537';
+                    const token = jwt.sign(
+                        {
+                            name: 'gitbook-open-tests',
+                            // Inflate the token above the ~4KB browser cookie limit,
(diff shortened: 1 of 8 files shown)


taranvohra (commented):
Can we add an E2E test case to ensure we don't regress. 

Create a massive JWT token for a VA site using it's private key. Add a bunch of random claims to make it larger than 4 KB.

Then in the e2e test you pass it as `?jwt_token=...` and asser that
- The test works
- Navigating the pages work because the cookie is properly persisted in a chunked manner

conico974 (approved):
We knew this day would come 😄

peterwhite:
> Can we add an E2E test case to ensure we don't regress.
> 
> Create a massive JWT token for a VA site using it's private key. Add a bunch of random claims to make it larger than 4 KB.
> 
> Then in the e2e test you pass it as `?jwt_token=...` and asser that
> 
> * The test works
> * Navigating the pages work because the cookie is properly persisted in a chunked manner

Good shout, on it!

packages/gitbook/e2e/internal.spec.ts
```diff
@@ -1655,6 +1655,71 @@ const testCases: TestsCase[] = [
             },
         ],
     },
+    {
+        name: 'Visitor Auth - Space (oversized token)',
+        contentBaseURL: 'https://gitbook.gitbook.io/gbo-va-space/',
+        // Our Cloudflare stack still folds multiple Set-Cookie headers into one,
+        // breaking chunked cookies (variant of opennextjs-cloudflare#501).
```
peterwhite, line 1662: Ticketed here: https://linear.app/gitbook-x/issue/RND-12667/cloudflare-folds-multiple-set-cookie-headers-into-one

PR #4555, opened 2026-08-28
Disable zoom for line sized inline images

tgargula:
## Demo

https://github.com/user-attachments/assets/0f0f3943-59f7-48e0-81bf-975db091d444

### DIFF
--- a/packages/gitbook/src/components/DocumentView/InlineImage.tsx
+++ b/packages/gitbook/src/components/DocumentView/InlineImage.tsx
@@ -9,7 +9,7 @@ import { type ResolvedContentRef, resolveContentRefInDocument } from '@/lib/refe
 import { tcls } from '@/lib/tailwind';
 
 export async function InlineImage(props: InlineProps<DocumentInlineImage>) {
-    const { document, inline, context, ancestorInlines } = props;
+    const { document, inline, context } = props;
     const { size = 'original' } = inline.data;
 
     const [src, darkSrc] = await Promise.all([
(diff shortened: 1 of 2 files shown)


peterwhite (approved):
Good catch!

viktorrenkema (approved):

packages/gitbook/src/components/DocumentView/InlineImage.tsx
```diff
@@ -54,7 +53,6 @@ export async function InlineImage(props: InlineProps<DocumentInlineImage>) {
                 loading={context.mode === 'print' ? 'eager' : 'lazy'}
                 style={[size === 'line' ? ['max-h-lh', 'h-lh', 'w-auto'] : null]}
                 inline
-                zoom={!isInLink}
```
taranvohra, line 57: Thinking out loud if we should disable it right away or show a zoom option on hover. Because I don't know how many people could already be relying on this.

iKettles: @taranvohra What if we limit it to only preventing zoom if the size isn't set to line? I can't think of a practical use case for a line-sized inline image that requires zoom, most of the time it's an icon

taranvohra: I agree. I am just thinking of old use cases where people unwillingly added inline images from the editor because back then it was very easy to add an image as inline. So wondering if people are relying on the zoom in that case

tgargula: I could do: `size !== line && !isInLink` just to be safe

PR #4556, opened 2026-08-28
Run the cached-404 guard on regional cache hits

peterwhite:
A section linked to an empty space throws notFound() mid-stream, which gets cached as a 200 with an empty body. Cache interception replays it, and Chrome downloads the empty response as a 0-byte file.

### DIFF
--- a/packages/gitbook/openNext/incrementalCache/incrementalCache.ts
+++ b/packages/gitbook/openNext/incrementalCache/incrementalCache.ts
@@ -15,6 +15,36 @@ export type KeyOptions = {
     cacheType?: CacheEntryType;
 };
 
+//TODO: This is a workaround to handle 404 responses in the cache.
+// It should be handled by OpenNext cache interception directly. This should be removed once OpenNext cache interception is fixed.
+// Cache interception serves entries with the rewrite status code (always 200), so a cached 404
+// would otherwise be replayed as a 200 (RND-12656).
+export function returnNullOn404<CacheType extends CacheEntryType = 'cache'>(
+    cacheEntry: WithLastModified<CacheValue<CacheType>> | null
+): WithLastModified<CacheValue<CacheType>> | null {
+    if (!cacheEntry?.value) return null;
+    if ('meta' in cacheEntry.value && cacheEntry.value.meta?.status === 404) {
+        return null;
+    }
+    return cacheEntry;
+}
+
+/**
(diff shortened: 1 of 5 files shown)


packages/gitbook/openNext/incrementalCache/incrementalCache.ts
```diff
@@ -55,13 +55,21 @@ export class GitbookIncrementalCache implements IncrementalCache {
...
         cacheEntry: WithLastModified<CacheValue<CacheType>> | null
     ): WithLastModified<CacheValue<CacheType>> | null {
         if (!cacheEntry?.value) return null;
-        if ('meta' in cacheEntry.value && cacheEntry.value.meta?.status === 404) {
+        const value = cacheEntry.value;
+        if ('meta' in value && value.meta?.status === 404) {
+            return null;
+        }
+        // A notFound() thrown mid-stream can be cached as a 200 with an empty body (RND-12656).
+        if ('type' in value && (value.type === 'app' || value.type === 'page') && !value.html) {
```
conico974: Doesn't seems right, the cached data should be correct. Do you have a reproduction of this ? Does the issue not occurs on 2v

peterwhite: You were right, cached data is fine, it doesn't occur on Vercel

Repro, it's unreliable though:

1. Repro: https://allthethings-inc.gitbook-staging.io/rnd-12656-repro/empty-section (staging site, section linked to a space with no pages)
2. curl -i it a few times: first hit is a correct 404, then within the regional cache TTL you get 200 + x-opennext-cache: HIT with the exact same not-found body.

I changed the approach in the PR to add the existing 404 workaround onto the regional cache, which I think should fix it?

conico974 (changes requested):
This is definitely not the right fix, even my initial workaround was wrong and should not be needed anymore.
Doing that means that we are not caching 404 as we should

peterwhite:
> This is definitely not the right fix, even my initial workaround was wrong and should not be needed anymore.
> Doing that means that we are not caching 404 as we should

Fair enough, thought I'd give it a go! I'll close it now 😄

packages/gitbook/openNext/incrementalCache/incrementalCache.ts
conico974: Pretty sure this is not the issue, if it was, during the entire ttl of the cache it would be broken, this is not the case here

PR #4561, opened 2026-08-28
Support external links

tgargula:
Refactor site structure handling to use switch statements and `assertNever` for improved type safety. This ensures exhaustive handling of all site structure node types across components and utilities, providing better compile-time guarantees for site structure traversal.

Changes include refactoring `getSiteStructureSections`, `listAllSiteSpaces`, and related utilities to use exhaustive switch statements with proper type narrowing and filtering.

## Demo

https://github.com/user-attachments/assets/7c09771a-97a0-440b-a3e9-60da8f899d41


conico974 (commented):
It's probably worth adding an e2e test for that

conico974 (approved):
Approving to unblock, but we should add e2e tests in a separate PR

tgargula:
We currently do not test sections & site structure in E2E tests. The external links are added to GBX behind a feature flag and the groups/sections are already live, so there should be no breaking change here.
I suggest to work on the E2E tests separately: https://linear.app/gitbook-x/issue/RND-12681/add-site-structure-e2e-tests

PR #4563, opened 2026-08-29
Prewarm Turbopuffer when published search opens

jpreynat:
## Changes

- observe the centralized published-search open state and trigger one prewarm on each closed-to-open transition
- call the generated prewarmSiteSearch operation through the existing middleware-resolved organization, site, API base URL, and content or visitor authentication
- keep prewarming best-effort so it never blocks the dialog, input, or search requests
- update @gitbook/api to 0.198.0, the first release containing prewarmSiteSearch

Opening search now gives Turbopuffer an early cache-warming hint, reducing the likelihood that the first visitor query pays cold-namespace latency. This adds one best-effort API call per search opening. It does not block the dialog and does not execute an extra search.

## Testing

- bun test src/components/Search/useSearchPrewarm.test.ts — 5 passed
- bun run format — passed
- bun run lint — passed
- bun run typecheck from packages/gitbook — passed

## Links

- https://linear.app/gitbook-x/issue/RND-11915/prewarm-tpuf-cache-when-opening-the-gbo-search-dialog
- https://github.com/GitbookIO/gitbook-x/pull/25222

### DIFF
--- a/packages/gitbook/src/components/Search/prewarmSearch.ts
+++ b/packages/gitbook/src/components/Search/prewarmSearch.ts
@@ -1,0 +1,14 @@
+'use server';
+
+import { getSiteURLDataFromMiddleware } from '@/lib/middleware';
+import { getServerActionBaseContext } from '@/lib/server-actions';
+
+export async function prewarmPublishedSearch() {
+    const [context, siteURLData] = await Promise.all([
+        getServerActionBaseContext(),
+        getSiteURLDataFromMiddleware(),
+    ]);
+    const apiClient = await context.dataFetcher.api();
+
+    await apiClient.orgs.prewarmSiteSearch(siteURLData.organization, siteURLData.site);
+}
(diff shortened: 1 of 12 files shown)


packages/gitbook/src/components/Search/prewarmSearch.ts
```diff
@@ -0,0 +1,14 @@
+'use server';
+
+import { getSiteURLDataFromMiddleware } from '@/lib/middleware';
+import { getServerActionBaseContext } from '@/lib/server-actions';
+
+export async function prewarmPublishedSearch() {
```
conico974: We should not use a server action for that, server actions runs sequentially, not in parallel. Using this mean blocking any other server actions from running while it resolves

jpreynat: Makes sense, thanks for flagging 👍

jpreynat: Addressed in e21b3851e. Prewarming now uses a dedicated same-origin POST route instead of a server action, so it can run concurrently with other server actions. The route still resolves the existing middleware context and authenticated API client before calling prewarmSiteSearch.

conico974 (approved):

PR #4566, opened 2026-08-31
Preserve linked page destinations in published search

jpreynat:
## Changes

- preserve absolute HTTP(S) paths already returned for linked sidebar pages as the exact search result `href`, including query strings and fragments in regular and embeddable search
- keep existing relative GitBook page paths, embeddable GitBook links, section anchors, unresolved published-space handling, metadata, ranks, scores, and ordering unchanged
- extract the pure page-result transform into a sibling module so the public Runway example can be covered with synthetic regression data
- add no search request, redirect lookup, backend call, or meaningful latency; this only changes how an already-returned path is converted into an `href`

The backend change in [GitbookIO/gitbook-x#25237](https://github.com/GitbookIO/gitbook-x/pull/25237) makes linked sidebar pages searchable. This GBO change preserves their configured external destination when rendering the returned result.

## Testing

- `bun run format`
- `bun run lint`
- `cd packages/gitbook && bun run typecheck`
- `cd packages/gitbook && bun test src/app/sites/dynamic/[mode]/[siteURL]/[siteData]/~gitbook/search/route.test.ts` (8 passed)

## Follow-up

Linear: [RND-10265](https://linear.app/gitbook-x/issue/RND-10265/search-should-index-follow-link-pages)

Do not resolve the Linear issue on merge. After both backend and GBO changes are deployed, the affected Runway space still needs to be reindexed and verified in published search.


packages/gitbook/src/lib/search.ts
conico974, line 1: I think it would make sense to move this into lib rather than in the app path

conico974 (approved):

packages/gitbook/src/lib/search.ts
jpreynat: Moved `transformSitePageResult` to `packages/gitbook/src/lib/search.ts` and updated the route and regression test imports in 34acd0390.

PR #4569, opened 2026-09-01
Allow adding query params to the @webframe.navigate action

peterwhite:
Fixes [RND-12678](https://linear.app/gitbook-x/issue/RND-12678/allow-adding-a-query-param-webframenavigate-action): the `@webframe.navigate` action (#4362) accepts a `path` and `anchor` but no query params, which we need to open the ask widget from a webframe.

Adds an optional `query` object to the action, serialized with `URLSearchParams` and appended between the resolved path and the anchor:

```js
window.parent.postMessage(
    { action: { action: '@webframe.navigate', path: 'some/page', query: { ask: 'true' } } },
    '*'
);
```

### DIFF
--- a/.changeset/webframe-navigate-query.md
+++ b/.changeset/webframe-navigate-query.md
@@ -1,0 +1,6 @@
+---
+"gitbook": patch
+"@gitbook/react-contentkit": patch
+---
+
+Support query params in the `@webframe.navigate` action.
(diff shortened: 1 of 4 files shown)


spastorelli (changes requested):
Can we have the query params validated against a whitelist of params? We don't want for example integration passing `jwt_token` (at least for now, maybe later behind a scope)

conico974 (approved):

peterwhite:
> Can we have the query params validated against a whitelist of params? We don't want for example integration passing `jwt_token` (at least for now, maybe later behind a scope)

Good call, thank you!

gregberge (approved):

spastorelli (approved):

PR #4573, opened 2026-09-02
Drop the body-anchored cover gate from paragraph contrast text

peterwhite:
Every paragraph carried `page-cover-background:[&:not(:has(.button,input))]:text-contrast-cover`. The variant compiles to `body:has(...)`, so any DOM insertion on the page re-styled all 18k paragraphs, each re-running its own `:has()`. On the Harness release notes page that was ~200 ms per insertion and most of the 8.8 s of style work during load.

The gate is redundant: `text-contrast-cover` only applies to elements the client marks with `data-over-cover`, which only happens on background-cover pages. Paragraph styles now live behind a single `paragraph` class in `globals.css` instead of a utility string on every element. Style time during load drops from 8.8 s to 4.7 s, total main-thread time from 11.0 s to 7.0 s, and the HTML from 42 MB to 35.5 MB, with no visual change.

Fixes RND-12713
https://linear.app/gitbook-x/issue/RND-12713/harness-smp-release-notes-page-is-43-mb-and-triggers-page-unresponsive

### DIFF
--- a/packages/gitbook/src/components/DocumentView/Paragraph.tsx
+++ b/packages/gitbook/src/components/DocumentView/Paragraph.tsx
@@ -12,21 +12,13 @@ export function Paragraph(props: BlockProps<DocumentBlockParagraph>) {
     const { block, style, ...contextProps } = props;
     const { context } = contextProps;
 
-    // InlineActionButtons use flex-grow to take the available width. This requires the parent to be a flex container.
-    const inlineButtonStyle =
-        'has-[.button,input]:flex has-[.button,input]:flex-wrap has-[.button,input]:gap-2 has-[.button,input]:items-center';
-
     const paragraph = (
         <p
+            // Cover-aware contrast text applies only to the page body, not to documents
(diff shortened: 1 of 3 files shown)


packages/gitbook/src/components/DocumentView/Paragraph.tsx
```diff
@@ -22,8 +22,7 @@ export function Paragraph(props: BlockProps<DocumentBlockParagraph>) {
             className={tcls(
                 // Cover-aware contrast text applies only to the page body, not to documents
                 // rendered in overlays (search answers, AI chat) on a background-cover page.
-                context.isPageBody &&
-                    'page-cover-background:[&:not(:has(.button,input))]:text-contrast-cover',
+                context.isPageBody && '[&:not(:has(.button,input))]:text-contrast-cover',
```
gregberge: Go deeper, add a class "paragraph" and refer it in CSS so we keep it as light as possible. A docs can have tons of paragraph we have to optimize it here.

gregberge (approved):

PR #4591, opened 2026-09-07
Serve markdown with text/plain content-type for ChatGPT

tgargula:
## Summary

Changed all markdown endpoints (llms.txt, llms-full.txt, and markdown pages) to serve content with `text/plain` Content-Type instead of `text/markdown` for better compatibility with ChatGPT and other LLM clients. Updated tests to reflect the new content type.

## Changes

- Updated Content-Type header from `text/markdown` to `text/plain` in three route handlers

resolves RND-12762


packages/gitbook/src/routes/markdownPage.ts
```diff
@@ -171,7 +171,7 @@ export async function serveMarkdown(
             typeof result === 'string' ? { markdown: result, robots: 'noindex' } : result;
         return new Response(markdown, {
             headers: {
-                'Content-Type': 'text/markdown; charset=utf-8',
+                'Content-Type': 'text/plain; charset=utf-8',
```
conico974: I don't think we should do that for everyone, if it affects only chatGPT, we should do it only for them. We should be able to change the value in the middleware where we should have the necessary data

packages/gitbook/src/lib/context.ts
```diff
@@ -101,6 +101,9 @@ export type SiteURLData = Pick<
      * the static cache of the other routes.
      */
     isAiAgent?: boolean;
+
+    /** Whether the request comes from ChatGPT. Only set for Markdown and LLM routes. */
```
conico974, line 105: That's not what I had in mind, this will effectively double the number of cache entry for markdown.
The idea is to use `response.headers.set` in the middleware directly to override only for chatGPT and on these specific route

conico974 (changes requested):

packages/gitbook/src/lib/context.ts
tgargula: checked, and it doesn't work with Vercel deployments because it seems that they ignore these

conico974 (approved):
Annoying that it doesn't work in vercel as it should....

PR #4597, opened 2026-09-08
Restore Edit on Git page actions

tgargula:
## Proposed changes

Restores the "Edit on Git" page action for Git-synced pages. When the `Git` page action is enabled in customization and Git Sync is active, the page data fetcher now lazy-loads Git metadata to provide the Edit on GitHub link in the page actions dropdown.

The implementation checks if the Git action is enabled and Git Sync is configured, then fetches metadata only when needed to avoid unnecessary API calls for pages that don't require it.

## Changelog

- [Feature] Restore Edit on Git page actions for Git-synced pages

### DIFF
--- a/.changeset/bright-pages-describe.md
+++ b/.changeset/bright-pages-describe.md
@@ -1,0 +1,5 @@
+---
+"gitbook": patch
+---
+
+Include published page descriptions in the page's Markdown output.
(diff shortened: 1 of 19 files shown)


arnauio (approved):

conico974 (changes requested):
No we should not do that for sure. We should always have `metadata: true` if we need it, then on GBX we fix caching for that (maybe another additional option like `cache: true`)

conico974 (approved):

PR #4599, opened 2026-09-09
Fix fallback URL cleanup after successful page navigation

peterwhite:
Related to https://linear.app/gitbook-x/issue/RND-11894

Cleaning up `fallback=true` currently calls `router.push`, adding a history entry. Pressing Back then revisits the same URL and triggers cleanup again.

Use `history.replaceState` to clean up the current entry without another navigation. Keep using `pathname` and `searchParams` to construct the URL, preserving other query parameters and the anchor.

### DIFF
--- a/packages/gitbook/src/components/SitePage/PageClientLayout.tsx
+++ b/packages/gitbook/src/components/SitePage/PageClientLayout.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import { usePathname, useRouter, useSearchParams } from 'next/navigation';
+import { usePathname, useSearchParams } from 'next/navigation';
 import React from 'react';
 
 import type { PageMetaLinks } from './SitePage';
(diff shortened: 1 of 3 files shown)


packages/gitbook/src/components/SitePage/PageClientLayout.tsx
```diff
@@ -33,17 +33,17 @@ export function PageClientLayout({ pageMetaLinks }: { pageMetaLinks: PageMetaLin
...
-            params.delete('fallback');
-            router.push(`${pathname}?${params.toString()}${window.location.hash ?? ''}`);
+        // Middleware strips fallback from the rewritten URL, so read the browser URL directly.
+        const url = new URL(window.location.href);
+        if (url.searchParams.has('fallback')) {
+            url.searchParams.delete('fallback');
+            window.history.replaceState(null, '', url);
         }
-    }, [router, pathname, searchParams]);
+    }, [pathname, searchParams]);
```
gregberge, line 50: I am not sure to understand what was wrong. It's a bit weird to watch some stuff and not use it (pathname and searchParams).

gregberge (approved):

conico974 (approved):

packages/gitbook/src/components/SitePage/PageClientLayout.tsx
peterwhite: Good point, I’ve updated it to use pathname and searchParams directly

PR #4614, opened 2026-09-16
Hide Intercom launcher when Assistant opens

tgargula:
## Summary
Hide the Intercom launcher bubble while the AI Assistant panel is open to prevent it from overlapping the message input box.

## Demo

I couldn't find an example with the Intercom widget, so here's a mocked one. `window.Intercom` is a standard way to handle the widget: https://developers.intercom.com/installing-intercom/web/attributes-objects#messenger-attributes

https://github.com/user-attachments/assets/476cff20-5ad5-434a-ba03-8c813cca9c3c

Intercom allows 20 updates in 30 mins per instance. So the launcher may still bug if you open / close many times in a row. If that becomes a problem, we can tackle it in a different way

## Changes
- Added `setIntercomLauncherHidden()` function to manage Intercom launcher visibility
- Integrated with AI chat open/close handlers
- Includes tests and temporary mock for local development


peterwhite (approved):
Looks good, IMHO test is redundant, could be removed

peterwhite (approved):
Looks good, IMHO test is redundant, could be removed

PR #4616, opened 2026-09-16
Filter table with client side selection

BrettJephson:
Choosing an option in a tab or a select button now narrows any table or card grid whose Select column offers it. For example, a Platform column follows a macOS tab; a Status column is left alone.

Matching is on the option's **label**, slugified the way tab titles are, so a table and the tabs beside it agree. Most recent wins when a column offers several active options.

Screenshot 2026-09-18 at 09 05 50

https://claude.ai/code/session_017cDq5dtdh4eGcpekaqtCxE


BrettJephson:
**Blocker found in review — must fix before this leaves draft.**

The slug filter can narrow a table with no visible indication and no way to clear it. `shouldShowTableSearch` is `searchOverride ?? (recordCount >= 7 && viewType === 'grid')`, so the filter controls are absent in three common cases:

- **cards view** — always, by default
- **grid tables under 7 rows**
- **any table where the author set `search: false`**

In those cases a reader sees fewer rows than the author wrote, with nothing saying why. I had argued earlier that "the tab that set the slug is the control" — that's wrong: the tab is often on a *different page*, so the reader arrives at an already-narrowed table with no nearby control at all.

Second half, and the sharper one: the reader's own filter change is component state while the slug is in `localStorage`, so **a reload always reasserts the slug filter**. The reader can't override it durably.

Proposed fix: show a "Filtered to macOS · Clear" affordance whenever the slug filter is active, independent of the search bar, and have Clear call the store's existing `deactivate(slug)` so the override persists without new storage.

BrettJephson:
**Blocker addressed** in 4604ff439.

- A `TableSelectionFilter` notice renders **independently of the search bar** — "Filtered to macOS" with a Clear — so cards, short grids and `search: false` tables get it too. That was the whole invisible-filter class, not just the cards case.
- **Clear calls the store's `deactivate`**, not local state. The selection is what persists, so dropping it there is what stops the filter returning on the next load. It is site-wide by design: a tab elsewhere on the page reverts to its default too.
- The applying effect now **reconciles against the columns it narrowed last** instead of only ever adding, so a new selection — or a clear — undoes the previous one. Without that, clearing left the old filter applied.

New translation key `table_filtered_by_selection` added across all 39 language files, per the existing convention.

This also answers the earlier note about mirroring the store into state: the effect now synchronises with an external store and reconciles its own previous writes, which is the sanctioned use rather than mirroring a derived value.

Still open from the original list: the brief unfiltered paint before the filter applies. Tabs avoid it with pre-paint CSS; rows have no equivalent yet.

peterwhite (approved):

